PR #214, opened 2018-03-16
Updating golib to get gzip compression

keitwb:
FYI @mdubbyap


mpetazzoni:
What's up with the merge? Shouldn't this have been rebased beforehand?

PR #217, opened 2018-03-19
Ansible deploy

prasadadireddi:
Ansible playbook to deploy Signalfx Agent


deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
```diff
@@ -0,0 +1,36 @@
...
+    gpgkey: "{{ rhel_gpg_key_url }}"
+    gpgcheck: yes
+    enabled: yes
+
+- name: Install Signalfx Agent via yum package manager
+  yum:
+    name: signalfx-agent
+    state: present
+
+- name: add signalfx group
```
keitwb: You don't really need to add the user/group after the package is installed, since the install scripts in the package do that for both Debian and RHEL.  It only needs to be done before the package is installed if you want to render the config file with the right owner/group, but you can just do it after the package is installed so that it is never necessary to make a user/group.

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
```diff
@@ -0,0 +1,36 @@
...
+  group:
+    name: signalfx-agent
+    state: present
+
+- name: add signalfx user
+  user:
+    name: signalfx-agent
+    state: present
+
+- name: write signalfx config
```
keitwb: This file should have user/group of `signalfx-agent`.

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: Add Signalfx Agent repo into source list
+  yum_repository:
+    name: signalfx-agent
+    description: SignalFx Agent Repository
+    baseurl: "{{ rhel_repo_url }}"
+    gpgkey: "{{ rhel_gpg_key_url }}"
+    gpgcheck: yes
+    enabled: yes
+
+- name: Install Signalfx Agent via yum package manager
```
keitwb: All of the logic here and below is repeated from the debian task file.  Pull the common steps out to the main.yml file so that there is no repetition.

deployments/ansible/roles/signalfx-agent/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+- name: Check SignalFx Access Token is defined
+  fail: msg="Bailing out. This play requires 'signalFxAccessToken'"
+  when: conf['signalFxAccessToken'] is undefined or conf['signalFxAccessToken'] is none or (conf['signalFxAccessToken'] | trim == '')
+
+# Import tasks according to the distribution
+- name: Import signalfx-agent deploy for CentOS or RHEL
+  import_tasks: yum_repo.yml
+  when: ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux' or ansible_distribution == 'Amazon'
+
+- name: Import signalfx-agent deploy for Debian or Ubuntu
```
keitwb: Have it throw an error if the distribution is neither centos or debian based (e.g. Alpine, ArchLinux, etc.).

deployments/ansible/roles/signalfx-agent/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+# Check Signalfx Access Toekn is defined
+- name: Check SignalFx Access Token is defined
+  fail: msg="Bailing out. This play requires 'signalFxAccessToken'"
+  when: conf['signalFxAccessToken'] is undefined or conf['signalFxAccessToken'] is none or (conf['signalFxAccessToken'] | trim == '')
+
+# Import tasks according to the distribution
+- name: Import signalfx-agent deploy for CentOS or RHEL
+  import_tasks: yum_repo.yml
+  when: ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux' or ansible_distribution == 'Amazon'
```
keitwb: Try using the Jinja "in" operator as described at http://jinja.pocoo.org/docs/dev/templates/#other-operators so that this will be a bit easier to read.

deployments/ansible/group_vars/all
```diff
@@ -0,0 +1,32 @@
...
+debian_gpg_key_url: "{{ repo_base_url }}/debian.gpg"
+
+# Variables specific to rhel
+rhel_repo_url: "{{ repo_base_url }}/rpms/signalfx-agent/{{ package_stage }}"
+rhel_gpg_key_url: "{{ repo_base_url }}/yum-rpm.key"
+
+conf_file_path: /etc/signalfx/agent.yaml
+
+conf:
+  signalFxAccessToken: MY_TOKEN
```
keitwb: If you set it here, won't it default to this and not give an error if the user doesn't override it?

deployments/ansible/roles/signalfx-agent/tasks/debian_repo.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: Add an Apt signing key for Signalfx Agent
+  apt_key:
+    url: "{{ debian_gpg_key_url }}"
+    keyring: /etc/apt/trusted.gpg.d/signalfx.gpg
+    state: present
+
+- name: Add Signalfx Agent repository into sources list
+  apt_repository:
+    repo: "{{ debian_repo_url }}"
```
keitwb: Construct this string directly here instead of using a separate variable.  There is really no need to since the value is only used once.

deployments/ansible/group_vars/all
```diff
@@ -0,0 +1,32 @@
+---
```
keitwb: Is this some kind of defaults file or just for testing?

deployments/ansible/group_vars/all
```diff
@@ -0,0 +1,32 @@
...
+# Variables for Signalfx Agent package
+
+repo_base_url: https://dl.signalfx.com
+package_stage: beta
```
keitwb: Default should be `final`, not `beta`.

deployments/ansible/hosts
```diff
@@ -0,0 +1,3 @@
+# Inventory file for production servers
```
keitwb: This file probably shouldn't be here.

deployments/ansible/roles/signalfx-agent/handlers/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# signalfx-agent handler
+- name: restart signalfx-agent
```
keitwb: How does this get used?  Does it just automatically restart the agent every time ansible is run or only on certain conditions?  Ideally the agent should only restart when its config file changes and not otherwise.

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Add Signalfx Agent repo into source list
+  yum_repository:
+    name: signalfx-agent
+    description: SignalFx Agent Repository
+    baseurl: "{{ rhel_repo_url }}"
```
keitwb: Construct this variable here instead of in the vars file.

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
```diff
@@ -0,0 +1,36 @@
...
+    gpgkey: "{{ rhel_gpg_key_url }}"
```
keitwb: Same thing as above.

deployments/ansible/site.yml
```diff
@@ -0,0 +1,10 @@
+---
+# This playbook deploys a Signalfx auto-discovery agent
```
keitwb: Is this for testing or is it used in the actual usage of the module?

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
prasadadireddi: got it, removed the tasks to add user/group explicitly

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
prasadadireddi: done

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
prasadadireddi: sure, moved common tasks to main

deployments/ansible/roles/signalfx-agent/tasks/main.yml
prasadadireddi: No, it will not throw any error. It will just skip the tasks if the condition not matched. Going to add the 'conditional' with appropriate message when the play is performed on other distributions!

deployments/ansible/group_vars/all
prasadadireddi: It is going to be empty config as we have in chef & puppet in final commit, and will mention the format in the documentation. It was set with dummy for review purpose, will be deleted in next commit

deployments/ansible/roles/signalfx-agent/tasks/debian_repo.yml
prasadadireddi: yeah, updated

deployments/ansible/group_vars/all
prasadadireddi: 'all' is default file in Ansible which defines common variables across all types of hosts. And, there will be group specific var files as siblings if any, which contains variables specific to that group of hosts, like 'webservers', 'dbservers', etc.

deployments/ansible/group_vars/all
prasadadireddi: ok, updated

deployments/ansible/hosts
prasadadireddi: yes, removed. This was added for test. Ideally the inventory can be passed to playbook from any where by the customer depends on their requirement.

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
prasadadireddi: done

deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml
prasadadireddi: done

deployments/ansible/site.yml
prasadadireddi: This is actual usage of the module. Renamed to playbook.yml for better readability and to avoid confusion.

prasadadireddi:
Implemented review comments, please check the in-line response

deployments/ansible/README.md
```diff
@@ -0,0 +1,57 @@
+# SignalFx Agent Ansible Playbook
```
keitwb: I think this is a "role".  Playbooks are what specify your infrastructure as a whole right?  Confirm this terminology.

deployments/ansible/README.md
```diff
@@ -0,0 +1,57 @@
...
+        - type: collectd/protocols
+        - type: collectd/signalfx-metadata
+        - type: collectd/uptime
+        - type: collectd/vmem
+    ```
+
+	It is probably going to be simpler to keep this config in `target-group` var file at the 
+	`group_vars` and update the same in playbook, which will make it visible to that group of hosts under inventory.
+
+ - `{{ package_stage }}`: The package repo stage to use: `main`, `beta`, or `test`
```
keitwb: "main" was switched to "final" before the initial release.

prasadadireddi: Updated

deployments/ansible/README.md
prasadadireddi: Yes, it is the 'signalfx-agent' role defined in the playbook which will play in the defined infrastructure. Ideally customer going to copy our role to their playbook and play in their infra to install signalfx agent. Updated the doc accordingly.

deployments/ansible/roles/signalfx-agent/tasks/debian_repo.yml
```diff
@@ -0,0 +1,24 @@
...
+    state: present
+
+- name: Add Signalfx Agent repository into sources list
+  apt_repository:
+    repo: "deb {{ repo_base_url }}/debs/signalfx-agent/{{ package_stage }} /"
+    filename: 'signalfx-agent'
+    mode: 644
+    state: present
+
+- name: Clean up the existing Signalfx Agent
```
keitwb: This would result in the agent being uninstalled and reinstalled every time ansible is run on a server right?

deployments/ansible/group_vars/signalfx-host-group
```diff
@@ -0,0 +1,24 @@
+---
```
keitwb: I think this file should be in the `defaults` dir.  Look at how the [MongoDB role](https://github.com/William-Yeh/ansible-mongodb) and [Nginx role](https://github.com/geerlingguy/ansible-role-nginx) do it.

deployments/ansible/playbook.yml
```diff
@@ -0,0 +1,10 @@
+---
```
keitwb: Call this something like `example-playbook.yml` so that it is clear this is for development.  People that are actually using this role won't need a playbook from us.

keitwb:
Also add meta information for adding this to the Ansible Galaxy site.  I will upload it there once this gets merged but I want it to be ready.

deployments/ansible/group_vars/signalfx-host-group
prasadadireddi: The existing structure set variables via facts according to the values defined per inventory group or all, followed the official ansible structure [MongoDB] (https://github.com/ansible/ansible-examples/tree/master/mongodb) and [other examples] (https://github.com/ansible/ansible-examples) which overrides the defaults one. Yes, the `defaults` dir from above example make more sense in defining default values for *signalfx-agent* role. Will update the structure accordingly!

deployments/ansible/playbook.yml
prasadadireddi: updated

deployments/ansible/roles/signalfx-agent/tasks/debian_repo.yml
prasadadireddi: Updated the task with `force: yes` option as the `downgrade` option is not yet supported for *apt* module

PR #218, opened 2018-03-19
Make SmartAgent 2 words

barbara-sfx:


keitwb:
Isn't this part of #216 ?

PR #220, opened 2018-03-20
Created salt deployment for signalfx-agent

gautam-swaroop:


keitwb:
@aeuio since you know way more about Salt than I do, do you mind taking a look at this as well?

deployments/salt/pillar.example
```diff
@@ -0,0 +1,19 @@
+isignalfx-agent:
```
keitwb: Why does it start with `i`?

deployments/salt/pillar.example
```diff
@@ -0,0 +1,19 @@
...
+  repo_base_url: https://dl.signalfx.com
+  package_stage: beta
```
keitwb: I would make this `final` instead of `beta` since most people are going to want to use that.

deployments/salt/signalfx-agent/init.sls
```diff
@@ -0,0 +1,89 @@
...
+    - file: /etc/apt/sources.list.d/signalfx-agent.list
+    - key_url: {{ signalfx_repo_base_url }}/debian.gpg
+    - gpgcheck: 1
+    - enabled: 1
+{% endif %}
+
+
+# signalfx-agent user and group creation.
+
+signalfx-agent.user:
```
keitwb: I would omit this because the package installation will do it anyway.

keitwb: The way that the Chef/Puppet scripts make the user beforehand is unnecessary and will be removed from those as well.

deployments/salt/signalfx-agent/init.sls
```diff
@@ -0,0 +1,89 @@
...
+    - gid_from_name: True
+
+
+# Installation of signalfx-agent package and starting of service.
+
+signalfx-agent.packages:
+  pkg.installed:
+    - name: signalfx-agent
+{% if salt['pillar.get']('signalfx-agent:version') is not None %}
+    - version: {{ salt['pillar.get']('signalfx-agent:version') }}
```
keitwb: Are you sure that the string "latest" will work here?  I'm not familiar with how Salt processes that but I don't think it is a valid version name in apt or yum.

deployments/salt/signalfx-agent/init.sls
```diff
@@ -0,0 +1,89 @@
...
+{% endif %}
+  service.running:
+    - name: signalfx-agent
+    - enable: True
+    - full_restart: True
```
keitwb: The behavior should be that it starts up the agent if it isn't running, and only restarts if the configuration has changed.  Chef and PUppet both have the concept of "notifications" upon changes to a resource, so see if Salt can do the same.

keitwb: Just make sure it isn't restarting the agent every time they run salt, regardless of whether the config has changed.

aeuio: you can set a "watch" on certain files that will restart a service if those files change - you can do this either on the service itself or on the config files.  Checkout the "watch" documentation in saltstack

deployments/salt/signalfx-agent/init.sls
aeuio: if the package is doing this best to omit this because salt could override what the package does

deployments/salt/pillar.example
gautam-swaroop: Removed i

deployments/salt/pillar.example
gautam-swaroop: done

deployments/salt/signalfx-agent/init.sls
gautam-swaroop: done

deployments/salt/signalfx-agent/init.sls
gautam-swaroop: Only the version numbers will work, 'latest' word will not work. Changed the pillar accrodingly

deployments/salt/signalfx-agent/init.sls
gautam-swaroop: Removed 'full_restart' and used 'watch' to observe the changes made in agent.yaml

keitwb:
Rebase this branch with the latest from master and it should fix the build error in CircleCI.

deployments/salt/signalfx-agent/init.sls
```diff
@@ -0,0 +1,82 @@
...
+
+
+
+# Installation of signalfx-agent package and starting of service.
+
+signalfx-agent.packages:
+  pkg.installed:
+    - name: signalfx-agent
+{% if salt['pillar.get']('signalfx-agent:version') is not none %}
+    - version: {{ salt['pillar.get']('signalfx-agent:version') }}
```
keitwb: Be sure it will support downgrading the agent version.  Prakash will check this in his testing as well.

deployments/salt/signalfx-agent/init.sls
```diff
@@ -0,0 +1,82 @@
...
+      - pkg: signalfx-agent
+    - watch:
+      - pkg: signalfx-agent
+      - file: /etc/signalfx/agent.yaml
+
+
+# Changing the agent.yaml configuration.
+
+{{ conf_file_path }}:
+  file.managed:
```
keitwb: The config file should be in place before the agent is started.  I don't quite understand the ordering in Salt, but make sure it does try and start the agent before the config is there.

keitwb:
Looks good.  Just two additional comments/questions to consider.

deployments/salt/signalfx-agent/init.sls
gautam-swaroop: The service will watch for changes in conf file and restarts whenever there is a change. So when we replace the existing conf file with our own, the service restarts

deployments/salt/signalfx-agent/init.sls
gautam-swaroop: The agent will downgrade when lower version is specified in pillar file

PR #265, opened 2018-04-30
Mask Apache, ES, and GenericJMX password field in output

keitwb:


jchengsfx:
Passwords for kafka, activemq, cassandra, and elasticsearch should also be masked.

charliesignalfx:
lgtm, are integration tests expected to fail right now?

keitwb:
No, integration tests are never expected to fail, that was a race condition caused by a bunch of tests running simultaneously but it should be fixed in this commit now.

PR #287, opened 2018-05-14
Updated collectd-mysql.md

rcastley:
Added note about how MySQL treats localhost vs. IP/Hostname


keitwb:
Merged into https://github.com/signalfx/signalfx-agent/pull/296

PR #291, opened 2018-05-16
Makefile and test updates

charliesignalfx:


scripts/windows/common.ps1
```diff
@@ -0,0 +1,28 @@
+Set-PSDebug -Trace 1
+$MY_SCRIPT_DIR = Split-Path $script:MyInvocation.MyCommand.Path
+
+function extra_cflags_build_arg() {
```
keitwb: This is useless on windows because it is flags for gcc optimization for building collectd.

scripts/windows/common.ps1
```diff
@@ -0,0 +1,28 @@
...
+
+function do_docker_build([string]$image_name, 
+                         [string]$image_tag,
+                         [string]$target_stage,
+                         [string]$agent_version,
+                         [string]$operating_system) {
+    $agent_version = if ($agent_version -eq "") { $image_tag }
+    $operating_system = if ($operating_system -eq "") { "windows" }
+    echo "$MY_SCRIPT_DIR"
+    $cache_flags = if ($PULL_CACHE -eq "yes") { "$MY_SCRIPT_DIR\..\docker-cache-from $target_stage" } else { "" }
```
keitwb: Can it run shell scripts like this?  If we have a dedicated windows build machine, we probably won't need the docker cache on quay.io since the machine will maintain a persistent cache unlike CircleCI.

Makefile
```diff
@@ -79,6 +94,20 @@ run-dev-image:
 		-v $(PWD)/collectd:/usr/src/collectd:cached \
 		-v /tmp/scratch:/tmp/scratch \
 		signalfx-agent-dev /bin/bash
+else
+	docker exec -it -e COLUMNS=`tput cols` -e LINES=`tput lines` signalfx-agent-dev /bin/bash -l -i 2>/dev/null || \
```
keitwb: I just tried the following and it works with far less duplication:

```
ifneq ($(OS),Windows_NT)
extra_run_flags = -v /:/hostfs:ro -v /var/run/docker.sock:/var/run/docker.sock:ro
endif

.PHONY: run-dev-image
run-dev-image:
	docker exec -it -e COLUMNS=`tput cols` -e LINES=`tput lines` signalfx-agent-dev /bin/bash -l -i 2>/dev/null || \
	  docker run --rm -it \
		$(extra_run_flags) \
		--cap-add DAC_READ_SEARCH \
		--cap-add SYS_PTRACE \
		--net host \
		-p 6060:6060 \
		--name signalfx-agent-dev \
		-v $(PWD)/local-etc:/etc/signalfx \
		-v $(PWD):/go/src/github.com/signalfx/signalfx-agent:cached \
		-v $(PWD)/collectd:/usr/src/collectd:cached \
		-v /tmp/scratch:/tmp/scratch \
		signalfx-agent-dev /bin/bash
```

internal/core/config/loader_test.go
```diff
@@ -448,7 +445,7 @@ var _ = Describe("Config Loader", func() {
 			signalFxAccessToken: abcd
 			monitors:
 			- type: collectd/mysql
-			  password: {"#from": "%s/agent/mysql-password.yaml", default: "s3cr3t"}
```
keitwb: How does this work on Windows?  I thought paths had to use backslashes?

internal/observers/kubelet/kubelet_test.go
```diff
@@ -83,7 +83,9 @@ func TestNoPods(t *testing.T) {
 				Removed: func(se services.Endpoint) { delete(endpoints, se.Core().ID) },
 			},
 		}
-		kub.Configure(config)
+		if err := kub.Configure(config); err != nil {
+			t.Error(err)
```
keitwb: `t.Fatal` since the test should stop if the observer could not be configured.

scripts/windows/common.ps1
charliesignalfx: leaving this in per our discussion

internal/core/config/loader_test.go
charliesignalfx: will obviously need to test to be 100% sure, which I'll get to in the future, but for now the test passes.

scripts/windows/common.ps1
charliesignalfx: yeah i missed that this was invoking other shell scripts.  I'm going to rip out the cache flags for now and if I get a moment, i might port the scripts that this depends on and add it back.

PR #293, opened 2018-05-16
Doc updates for df, disk, load, and mysql metrics

jchengsfx:


keitwb:
Squash before merging please.

PR #295, opened 2018-05-18
K8S integration tests with minikube

jchengsfx:
- Added minikube, final image, and test jobs for K8S v1.9.0, v1.8.0, and v1.7.0 to circleci config.
- Updated docs for df, disk, load, and mysql metrics.


tests/kubernetes/minikube.py
```diff
@@ -0,0 +1,204 @@
...
+    def load_kubeconfig(self, kubeconfig_path="/kubeconfig", timeout=300):
+        assert wait_for(p(container_cmd_exit_0, self.container, "test -f %s" % kubeconfig_path), timeout_seconds=timeout), "timed out waiting for the minikube cluster to be ready!\n\n%s\n\n" % self.container.logs().decode('utf-8').strip()
+        self.kubeconfig = "/tmp/scratch/kubeconfig-%s" % self.container.id[:12]
+        time.sleep(2)
+        rc, output = self.container.exec_run("cp -f %s %s" % (kubeconfig_path, self.kubeconfig))
+        assert rc == 0, "failed to get %s from minikube!\n%s" % (kubeconfig_path, output.decode('utf-8'))
+        time.sleep(2)
+        kube_config.load_kube_config(config_file=self.kubeconfig)
+
+    def connect(self, name, timeout, version=None):
```
keitwb: What exactly is this connecting to?

tests/kubernetes/minikube.py
```diff
@@ -0,0 +1,204 @@
...
+            rm=True,
+            forcerm=True)
+        self.container = self.host_client.containers.run(
+            image.id,
+            detach=True,
+            **options)
+        self.load_kubeconfig(timeout=timeout)
+        self.container.reload()
+        self.get_client()
+        self.get_ip()
```
keitwb: What is the point of calling this?  It doesn't appear to cause any side effects and isn't used in any way.

tests/kubernetes/minikube.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def create_secret(self, key, secret):
+        if self.container:
+            rc, output = self.container.exec_run("kubectl create secret generic %s --from-literal=access-token=%s" % (key, secret))
```
keitwb: Probably better to use the k8s api client instead of sometimes using kubectl.

tests/kubernetes/minikube.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @contextmanager
+    def deploy_yamls(self, yamls=[], services_dir=K8S_SERVICES_DIR):
+        self.yamls= []
+        if services_dir:
+            if len(yamls) == 0:
+                yamls = sorted([os.path.join(services_dir, y) for y in os.listdir(services_dir) if y.endswith(".yaml")])
+            else:
+                yamls = sorted([os.path.join(services_dir, y) for y in yamls if y.endswith(".yaml")])
+        print()
```
keitwb: Remove

tests/kubernetes/minikube.py
```diff
@@ -0,0 +1,204 @@
...
+            rc, output = self.container.exec_run("kubectl create secret generic %s --from-literal=access-token=%s" % (key, secret))
+            assert rc == 0
+            print_lines(output.decode('utf-8'))
+
+    @contextmanager
+    def deploy_yamls(self, yamls=[], services_dir=K8S_SERVICES_DIR):
+        self.yamls= []
+        if services_dir:
+            if len(yamls) == 0:
+                yamls = sorted([os.path.join(services_dir, y) for y in os.listdir(services_dir) if y.endswith(".yaml")])
```
keitwb: It is better to have a separate function called `deploy_service_yamls` or something that does the `os.listdir` and calls `deploy_yamls` with those files, that way the difference in encoded in the function name and this function is simpler.

tests/kubernetes/minikube.py
```diff
@@ -0,0 +1,204 @@
...
+        if image_id:
+            try:
+                self.client.images.get(image_id)
+                return
+            except:
+                pass
+        print("\nPulling %s:%s to the minikube container ..." % (name, tag))
+        self.container.exec_run("cp -f /etc/hosts /etc/hosts.orig")
+        self.container.exec_run("cp -f /etc/hosts /etc/hosts.new")
+        self.container.exec_run("sed -i 's|127.0.0.1|%s|' /etc/hosts.new" % get_host_ip())
```
keitwb: This is pretty hacky, what is this trying to do?  We can probably figure out a better way to do this.

tests/kubernetes/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+            print("\nAgent image '%s:%s' not found in local registry." % (final_agent_image_name, final_agent_image_tag))
+            print("\nAttempting to pull from remote registry ...")
+            final_image = client.images.pull(final_agent_image_name, tag=final_agent_image_tag)
+        except:
+            final_image = None
+    assert final_image, "agent image '%s:%s' not found!" % (final_agent_image_name, final_agent_image_tag)
+    try:
+        client.containers.get("registry")
+        print("\nRegistry container localhost:5000 already running")
+    except:
```
keitwb: Avoid pokemon exception handlers, it makes it frustrating when running tests locally by preventing `Ctrl-C` from exiting the process if pressed when this code is executing.  It also masks other problems besides the container just not existing.

tests/kubernetes/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+                name='registry',
+                detach=True,
+                ports={'5000/tcp': 5000})
+            print("\nStarted registry container localhost:5000")
+        except:
+            pass
+        print("\nWaiting for registry container localhost:5000 to be ready ...")
+        start_time = time.time()
+        while True:
+            assert (time.time() - start_time) < 30, "timed out waiting for registry container to be ready!"
```
keitwb: `wait_for`, it takes an optional arg for the timeout.

tests/kubernetes/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+                break
+            except:
+                time.sleep(2)
+    agent_image_name = getattr(request.module, "AGENT_IMAGE_NAME")
+    agent_image_tag = getattr(request.module, "AGENT_IMAGE_TAG")
+    print("\nTagging %s:%s as %s:%s ..." % (final_agent_image_name, final_agent_image_tag, agent_image_name, agent_image_tag))
+    final_image.tag(agent_image_name, tag=agent_image_tag)
+    print("\nPushing %s:%s ..." % (agent_image_name, agent_image_tag))
+    client.images.push(agent_image_name, tag=agent_image_tag)
+    yield
```
keitwb: Not sure how pytest fixtures work but you probably want to do something like this to make sure the registry gets axed if an exception is thrown:
```
try:
    yield
finally:
   client.containers.get("registry").remove(force=True)
```

tests/kubernetes/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+    try:
+        client.containers.get("registry").remove(force=True)
+    except:
```
keitwb: When would this ever fail?

tests/kubernetes/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+        start_time = time.time()
+        while True:
+            assert (time.time() - start_time) < 30, "timed out waiting for registry container to be ready!"
+            try:
+                client.containers.get("registry")
+                time.sleep(2)
+                break
+            except:
+                time.sleep(2)
+    agent_image_name = getattr(request.module, "AGENT_IMAGE_NAME")
```
keitwb: Split out the logic for tagging/pushing the agent to at the very least a separate function, and perhaps even a separate fixture that is dependent on this fixture.

tests/kubernetes/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+            except:
+                pass
+
+    request.addfinalizer(teardown)
+    if k8s_container:
+        mk.connect(k8s_container, k8s_timeout)
+    else:
+        mk.deploy(k8s_version, k8s_timeout)
+    try:
+        mk.create_secret("signalfx-agent", "testing123")
```
keitwb: Putting secret creation here seems to be mixing concerns too much.

tests/kubernetes/data.py
```diff
@@ -0,0 +1,152 @@
+from tests.kubernetes.utils import get_metrics_from_doc, get_dims_from_doc
+
+# list of monitor definitions
+# the monitor should be a YAML-based dictionary which will be used for the signalfx-agent agent.yaml configuration
+MONITORS_WITHOUT_ENDPOINTS = [
```
keitwb: All of this should go into separate monitor tests in the `tests/monitors` dir.  We might need to start putting subdirs in there for specific monitors if there are multiple files per monitor.

docs/monitors/collectd-df.md
```diff
@@ -34,18 +34,18 @@ cause only a subset of metrics to be emitted.
 
 | Name | Type | Description |
 | ---  | ---  | ---         |
-| `gauge.df_complex.free` | gauge | Free disk space in bytes |
```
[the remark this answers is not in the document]
jchengsfx: https://github.com/signalfx/signalfx-agent/pull/293

tests/helpers/fake_backend.py
```diff
@@ -51,41 +51,46 @@ def do_POST(self):
...
     class FakeAPIServer(BaseHTTPRequestHandler):
         def do_PUT(self):
             if '/dimension/' not in self.path:
                 self.send_response(404)
                 self.end_headers()
                 return
 
-            body = self.rfile.read(int(self.headers.getheader('Content-Length')))
+            try:
+                body = self.rfile.read(int(self.headers.getheader('Content-Length')))
```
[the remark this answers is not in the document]
jchengsfx: I think I added that when I first started and ran into some incompatibility issues with python2.  I'll revert the change since we should only be using python3.

tests/kubernetes/kubernetes_test.py
```diff
@@ -0,0 +1,133 @@
...
+            backend=backend,
+            image_name=AGENT_IMAGE_NAME,
+            image_tag=AGENT_IMAGE_TAG,
+            namespace="default") as agent:
+            agent_status = agent.get_status()
+            container_logs = agent.get_container_logs()
+            # temporary workaround for https://signalfuse.atlassian.net/browse/INT-595
+            #assert "testing123" not in agent_status, "plaintext password(s) found in agent-status output!\n\n%s\n" % agent_status
+            npasswords = agent_status.lower().count("testing123")
+            assert npasswords <= 1, "plaintext password(s) found in agent-status output!\n\n%s\n" % agent_status
```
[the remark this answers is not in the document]
jchengsfx: Yep, I already reverted in https://github.com/signalfx/signalfx-agent/pull/295/commits/14651063f7d68ec3767d4e9d77987a0658c24705.

.circleci/config.yml
```diff
@@ -12,6 +12,50 @@ references:
...
+      command: |
+        mkdir ~/testresults
+        docker run -d --rm \
+            -e K8S_VERSION=$(eval "echo -n $K8S_VERSION") \
+            -p 8443 \
+            -p 2375 \
+            -v /tmp/scratch:/tmp/scratch \
+            --privileged \
+            --name minikube \
+            minikube:latest
```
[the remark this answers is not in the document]
jchengsfx: For some reason, I had some issues running minikube from within the dev-image in circleci. So I just run it separately and connect to the container with the `--k8s-container` pytest option.

tests/conftest.py
```diff
@@ -0,0 +1,159 @@
...
+        "--k8s-test-timeout",
+        action="store",
+        default=K8S_DEFAULT_TEST_TIMEOUT,
+        help="Timeout (in seconds) for each K8S test (default=%d)." % K8S_DEFAULT_TEST_TIMEOUT
+    )
+    parser.addoption(
+        "--k8s-container",
+        action="store",
+        default=None,
+        help="Name of a running minikube container to use for the tests (the container should not have the agent or any services already running). If not specified, a new minikube container will automatically be deployed."
```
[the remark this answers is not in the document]
jchengsfx: See comment above.

.circleci/config.yml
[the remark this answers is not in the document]
keitwb: Hmm...see if you can figure out the error and let me know, I dealt a lot with docker inside the dev container so I might know a solution.  The simpler we can make this the better long term.

tests/kubernetes/utils.py
```diff
@@ -0,0 +1,520 @@
...
+    start_time = time.time()
+    while True:
+        assert (time.time() - start_time) <= timeout, "timed out waiting for daemonset \"%s\" to be deleted!" % name
+        if has_daemonset(name, namespace=namespace):
+            time.sleep(2)
+        else:
+            break
+
+
+def deploy_k8s_service(**kwargs):
```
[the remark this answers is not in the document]
jchengsfx: This actually isn't being used by anything anymore so I can remove it.

tests/kubernetes/utils.py
```diff
@@ -0,0 +1,520 @@
...
+    pod_template = get_pod_template(
+        name=name,
+        image=image,
+        ports=ports,
+        labels=labels,
+        env=env,
+        command=command,
+        args=args,
+        volume_mounts=volume_mounts)
+    if replicas > 1:
```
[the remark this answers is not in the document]
jchengsfx: Same as https://github.com/signalfx/signalfx-agent/pull/295#discussion_r189655091.

tests/kubernetes/minikube.py
jchengsfx: See https://github.com/signalfx/signalfx-agent/pull/295#discussion_r189635230.

tests/helpers/fake_backend.py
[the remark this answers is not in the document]
jchengsfx: Here's the exception:
```Exception happened during processing of request from ('172.17.0.3', 52592)
Traceback (most recent call last):
  File "/usr/lib/python3.5/socketserver.py", line 313, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python3.5/socketserver.py", line 341, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python3.5/socketserver.py", line 354, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.5/socketserver.py", line 681, in __init__
    self.handle()
  File "/usr/lib/python3.5/http/server.py", line 422, in handle
    self.handle_one_request()
  File "/usr/lib/python3.5/http/server.py", line 410, in handle_one_request
    method()
  File "/go/src/github.com/signalfx/signalfx-agent/tests/helpers/fake_backend.py", line 66, in do_PUT
    body = self.rfile.read(int(self.headers.getheader('Content-Length')))
AttributeError: 'HTTPMessage' object has no attribute 'getheader'```

It doesn't seem to affect the tests, but I can update it to just capture AttributeError exceptions.

File: tests/kubernetes/minikube.py [hunk 21]
```diff
@@ -0,0 +1,203 @@
...
+        self.ip = None
+
+    def get_client(self):
+        if self.container:
+            self.client = docker.DockerClient(base_url="tcp://%s:2375" % self.container.attrs["NetworkSettings"]["IPAddress"], version='auto')
+            return self.client
+        else:
+            return None
+
+    def get_ip(self):
```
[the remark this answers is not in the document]
jchengsfx: I'll change it to use `tests.helpers.util.container_ip()` instead.

tests/kubernetes/minikube.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            return None
+
+    def load_kubeconfig(self, kubeconfig_path="/kubeconfig", timeout=300):
+        assert wait_for(p(container_cmd_exit_0, self.container, "test -f %s" % kubeconfig_path), timeout_seconds=timeout), "timed out waiting for the minikube cluster to be ready!\n\n%s\n\n" % self.container.logs().decode('utf-8').strip()
+        self.kubeconfig = "/tmp/scratch/kubeconfig-%s" % self.container.id[:12]
+        time.sleep(2)
+        rc, output = self.container.exec_run("cp -f %s %s" % (kubeconfig_path, self.kubeconfig))
+        assert rc == 0, "failed to get %s from minikube!\n%s" % (kubeconfig_path, output.decode('utf-8'))
+        time.sleep(2)
```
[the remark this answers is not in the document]
jchengsfx: I ran into some possible timing issues between the creation of the kubeconfig file and the copy, so I just added an extra check.

tests/helpers/fake_backend.py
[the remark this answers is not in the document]
keitwb: Can it just `get`?

tests/kubernetes/minikube.py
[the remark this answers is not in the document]
keitwb: ok, yeah forgot that was already there.

.circleci/config.yml
```diff
@@ -12,6 +12,50 @@ references:
...
+        docker run --rm -t --net host \
```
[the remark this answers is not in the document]
jchengsfx: I'll remove it.

.circleci/config.yml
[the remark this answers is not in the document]
jchengsfx: Looks to be a non-issue now. Probably just a circleci issue when I was initially testing.  I'll remove it from the circleci config.

tests/helpers/assertions.py
```diff
@@ -53,3 +53,33 @@ def has_log_message(output, level="info", message=""):
 
 def udp_port_open_locally(port):
     return os.system("cat /proc/net/udp | grep %s" % (hex(port)[2:].upper(),)) == 0
+
+
+# check the fake backend if any metric in `metrics` exist
+def any_metric_found(backend, metrics):
```
keitwb: `backend` -> `fake_services` to match the rest of the assertion functions.

keitwb: Also, this whole function can just be `return any([has_datapoint_with_metric_name(backend, m) for m in metrics])`

tests/kubernetes/minikube.py
jchengsfx: The registry container used to get the agent image to the minikube container will only listen on localhost (unless we use TLS). So this just temporarily routes localhost to the host IP that the registry container is running on so that we can pull agent image.

tests/kubernetes/minikube.py
jchengsfx: I'll remove this for now, and re-implement if necessary based on how the monitor tests are restructured.

tests/kubernetes/utils.py
```diff
@@ -78,12 +113,11 @@ def create_serviceaccount(body=None, namespace="default", timeout=K8S_API_TIMEOU
...
-    configmaps = api.list_namespaced_config_map(namespace=namespace)
-    if configmaps:
-        for cm in configmaps.items:
-            if cm.metadata.name == name:
-                return True
-    return False
+    try:
+        api.read_namespaced_config_map(name, namespace=namespace)
+        return True
+    except:
```
keitwb: This catch-all exception handler (and the others like it) will mask other errors in the K8s API, which could make debugging test failures a lot more difficult.  Try and figure out exactly what exception the K8s client throws when the resource does not exist and catch it only.

keitwb: Let any other types of exceptions bubble up and crash the test runner so it is obvious if there are problems.

tests/kubernetes/minikube.py
```diff
@@ -161,25 +154,20 @@ def deploy_yamls(self, yamls=[], services_dir=K8S_SERVICES_DIR):
...
                 self.client.images.get(image_id)
                 return
             except:
                 pass
         print("\nPulling %s:%s to the minikube container ..." % (name, tag))
-        self.container.exec_run("cp -f /etc/hosts /etc/hosts.orig")
-        self.container.exec_run("cp -f /etc/hosts /etc/hosts.new")
-        self.container.exec_run("sed -i 's|127.0.0.1|%s|' /etc/hosts.new" % get_host_ip())
-        self.container.exec_run("cp -f /etc/hosts.new /etc/hosts")
         time.sleep(5)
```
keitwb: Is this still necessary?

tests/kubernetes/utils.py
```diff
@@ -73,8 +74,11 @@ def has_secret(name, namespace="default"):
     try:
         api.read_namespaced_secret(name, namespace=namespace)
         return True
-    except:
-        return False
+    except ApiException as e:
+        if 'Reason: Not Found' in str(e):
```
keitwb: `if e.status == 404:` will be more robust against client library upgrades.

jchengsfx: Thanks, I missed that.

PR #299, opened 2018-05-21
Add postgresql monitor to smart agent

charliesignalfx:


internal/monitors/collectd/postgresql/postgresql.go
```diff
@@ -0,0 +1,184 @@
...
+	// default set will be used [`custom_deadlocks`, `backends`, `transactions`,
+	// `queries`, `queries_by_table`, `query_plans`, `table_states`,
+	// `query_plans_by_table`, `table_states_by_table`, `disk_io`,
+	// `disk_io_by_table`, `disk_usage`]
+	Queries []string `yaml:"queries"`
+}
+
+// Result maps values from a query to a metric
+type Result struct {
+	// type defines a metric type
```
keitwb: Titlecase the first word of all of these comments to match the style of everything else.

internal/monitors/collectd/postgresql/postgresql.tmpl
```diff
@@ -0,0 +1,67 @@
+LoadPlugin postgresql
+
+<Plugin postgresql>
```
keitwb: You need to set the Interval at the plugin level too to match the `.IntervalSeconds` variable that is part of the monitor core config.

internal/monitors/collectd/postgresql/postgresql.tmpl
```diff
@@ -0,0 +1,67 @@
...
+  <Query custom_deadlocks>
```
keitwb: Why is this query specially given here?

internal/monitors/collectd/postgresql/postgresql.tmpl
```diff
@@ -0,0 +1,67 @@
...
+      ValuesFrom {{range $v := $r.ValuesFrom}}"{{$v}}" {{- end}}
+    </Result>
+    {{- end}}
+  </Query>
+  {{end}}
+  {{range $db := .Databases}}
+  <Database "{{$db.Name}}">
+    Host "{{$.Host}}"
+    Port "{{$.Port}}"
+    ReportHost {{$.ReportHost}}
```
keitwb: Use `{{toBool $.ReportHost}}` instead of relying on the stringified form of golang bools in the templating engine.

internal/monitors/collectd/postgresql/postgresql.tmpl
charliesignalfx: I don't know the full history, but I think it might have been query that was included in older versions of collectd and later removed.  We added it in our default config.

PR #300, opened 2018-05-22
Add couchbase

charliesignalfx:
* add couchbase monitor
* clean up doc examples for openstack and postgres


docs/monitors/collectd-postgresql.md
```diff
@@ -39,15 +40,15 @@ monitors:
     - "hostname"
     statement: "Select * From test Where host = $1;"
     results:
-    - type: gauge
+    - type: "gauge"
```
keitwb: It's fine to leave them in, but you typically don't need quotes in YAML unless the string contains special characters.

internal/monitors/collectd/couchbase/couchbase.tmpl
```diff
@@ -0,0 +1,19 @@
...
+  Import "couchbase"
+
+  <Module "couchbase">
+    Host "{{.Host}}"
+    Port "{{.Port}}"
+    CollectTarget  "{{.CollectTarget}}"
+    {{with .CollectBucket -}}CollectBucket "{{.}}"{{- end}}
+    {{with .ClusterName -}}ClusterName "{{.}}"{{- end}}
+    {{with .CollectMode -}}CollectMode "{{.}}"{{- end}}
+    {{with .IntervalSeconds -}}Interval {{.}}{{- end}}
```
keitwb: `IntervalSeconds` is *always* present and doesn't need a with block.

internal/monitors/collectd/couchbase/metricdoc.go
```diff
@@ -0,0 +1,127 @@
+package couchbase
+
+// GAUGE(gauge.bucket.basic.dataUsed): Size of user data within buckets of the specified state that are resident in RAM (%)
```
keitwb: Confirm that these metrics actually do start with `gauge.` and weren't just documented that way.

PR #301, opened 2018-05-24
Metadataupdates

charliesignalfx:


internal/monitors/collectd/collectd.go
```diff
@@ -149,33 +149,14 @@ func (cm *Manager) MonitorDidShutdown(monitorID types.MonitorID) {
...
-
-	if len(cm.activeMonitors) > 1 {
-		cm.RequestRestart()
+	if len(cm.activeMonitors) == 1 && !utils.IsSignalChanClosed(cm.stop) {
+		close(cm.stop)
+		cm.deleteExistingConfig()
+		<-cm.terminated
+		return
 	}
+	delete(cm.activeMonitors, monitorID)
```
keitwb: Sorry, I told you wrong on this.  The monitorID needs to be deleted before the return on line 156 too, as well as here.  The key thing is to do the delete *after* collectd is stopped.

PR #304, opened 2018-05-25
wait for service containers to come up on their designated tcp socket

charliesignalfx:


tests/monitors/apache_test.py
```diff
@@ -15,7 +15,9 @@
 
 def test_apache():
     with run_service("apache") as apache_container:
-        config = apache_config.substitute(host=apache_container.attrs["NetworkSettings"]["IPAddress"])
+        host_addr = apache_container.attrs["NetworkSettings"]["IPAddress"]
```
keitwb: Use `container_ip(apache_container)` from the utils module.

PR #307, opened 2018-05-29
Adding a default agent.yaml for the docker image

brandoncurrie:
Added the agent.yaml to be bundled into the docker image with a short readme.

### DIFF
--- a/deployments/docker/agent.yaml
+++ b/deployments/docker/agent.yaml
@@ -1,6 +1,6 @@
 # *Required* The access token for the org that you wish to send metrics to.
-signalFxAccessToken: ${SIGNALFX_ACCESS_TOKEN}
-ingestUrl: ${SIGNALFX_INGEST_URL:"https://ingest.signalfx.com"}
+signalFxAccessToken: ${SFX_ACCESS_TOKEN}
+ingestUrl: ${SFX_INGEST_URL}
 
 intervalSeconds: 10
 
@@ -24,6 +24,7 @@ monitors:
   - type: collectd/memory
   - type: collectd/protocols
   - type: collectd/signalfx-metadata
+  - type: host-metadata
     etcPath: /hostfs/etc
   - type: collectd/uptime
   - type: collectd/vmem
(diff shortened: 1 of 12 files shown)


deployments/docker/agent.yaml
```diff
@@ -0,0 +1,32 @@
+# *Required* The access token for the org that you wish to send metrics to.
+signalFxAccessToken: {"#from": "/etc/signalfx/token"}
```
keitwb: Change this to `signalFxAccessToken: $SIGNALFX_ACCESS_TOKEN` and it will pull it from an envvar, which will be much easier to configure in a docker container.

deployments/docker/agent.yaml
```diff
@@ -0,0 +1,32 @@
...
+ingestUrl: {"#from": "/etc/signalfx/ingest_url", default: "https://ingest.signalfx.com"}
```
keitwb: Same thing, but the var `$SIGNALFX_INGEST_URL`.

deployments/docker/agent.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+intervalSeconds: 15
```
keitwb: This is supposed to be 10 sec.

deployments/docker/agent.yaml
keitwb: Change the associated docs above too.

brandoncurrie: Do ENVVARs require brackes? e.g. ${SIGNALFX_ACCESS_TOKEN}

brandoncurrie: Done

deployments/docker/agent.yaml
brandoncurrie: Done

deployments/docker/agent.yaml
brandoncurrie: Done

keitwb:
@brandoncurrie Take a look now, I merged together the docs for running and configuring and actually added the config file to the docker image.  I'll merge it later today if all passes.

PR #309, opened 2018-05-30
Add collectd/jenkins monitor

charliesignalfx:


internal/monitors/collectd/jenkins/jenkins.go
```diff
@@ -0,0 +1,103 @@
...
+	SSLCACerts string `yaml:"sslCACerts"`
+}
+
+// Monitor is the main type that represents the monitor
+type Monitor struct {
+	collectd.MonitorCore
+}
+
+// Validate will check the config for correctness.
+func (c *Config) Validate() error {
```
keitwb: Remove if a noop.  Are there any groups of config options that should always occur together (e.g. SSL*, Username/APIToken) that you should validate?

tests/monitors/jenkins_test.py
```diff
@@ -0,0 +1,35 @@
...
+    with run_service("jenkins", buildargs={"JENKINS_VERSION": version, "JENKINS_PORT": str(port)}) as jenkins_container:
+        host = container_ip(jenkins_container)
+        config = jenkins_config.substitute(host=host, port=port)
+        assert wait_for(p(tcp_socket_open, host, port), 60), "service didn't start"
+
+        with run_agent(config) as [backend, _, _]:
+            assert wait_for(p(has_datapoint_with_dim, backend, "plugin", "jenkins")), "Didn't get jenkins datapoints"
+
+
+def test_oldest_supported_jenkins():
```
keitwb: Pytest parameterization would be less verbose.  If you really want to keep the "oldest"/"latest" tags on the tests you can use ids: https://docs.pytest.org/en/3.6.0/example/parametrize.html#different-options-for-test-ids

tests/monitors/jenkins_test.py
```diff
@@ -0,0 +1,35 @@
...
+    # technically we support 1.580.3, but the scripts needed to programmatically
+    # setup jenkins do not work prior to 1.651.3
+    run_jenkins_test("1.651.3-alpine", 8081)
```
keitwb: Specifying the port is superfluous because jenkins is going to run in separate docker containers with their own network namespace so ports aren't shared.

internal/monitors/collectd/jenkins/jenkins.go
charliesignalfx: No, it was a leftover

tests/monitors/jenkins_test.py
```diff
@@ -0,0 +1,32 @@
...
+    host: $host
+    port: $port
+    metricsKey: 33DD8B2F1FD645B814993275703F_EE1FD4D4E204446D5F3200E0F6-C55AC14E
+""")
+
+
+@pytest.mark.parametrize("version", [
+    # technically we support 1.580.3, but the scripts needed to programmatically
+    # setup jenkins do not work prior to 1.651.3
+    pytest.param("1.651.3-alpine"),
```
keitwb: wrapping it in `pytest.param()` is quite unnecessary.

PR #313, opened 2018-06-01
add basic auth to http_status test assertion

charliesignalfx:
* add basic auth to http_status test assertion
* fix couchbase test to wait for authentication to be set

### DIFF
--- a/tests/monitors/jenkins_test.py
+++ b/tests/monitors/jenkins_test.py
@@ -27,7 +27,7 @@ def test_jenkins(version):
         host = container_ip(jenkins_container)
         config = jenkins_config.substitute(host=host, port=8080)
         assert wait_for(p(tcp_socket_open, host, 8080), 60), "service not listening on port"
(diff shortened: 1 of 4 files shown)


tests/monitors/jenkins_test.py
```diff
@@ -27,7 +27,7 @@ def test_jenkins(version):
         host = container_ip(jenkins_container)
         config = jenkins_config.substitute(host=host, port=8080)
         assert wait_for(p(tcp_socket_open, host, 8080), 60), "service not listening on port"
-        assert wait_for(p(http_status, 200, "http://{0}:8080/metrics/33DD8B2F1FD645B814993275703F_EE1FD4D4E204446D5F3200E0F6-C55AC14E/ping/".format(host)), 120), "service didn't start"
+        assert wait_for(p(http_status, "http://{0}:8080/metrics/33DD8B2F1FD645B814993275703F_EE1FD4D4E204446D5F3200E0F6-C55AC14E/ping/".format(host)), 120), "service didn't start"
```
keitwb: It is not at all intuitive that 200 is a "default" status and if I was reading through this code without seeing the implementation of `http_status` I would be very confused as to what this assertion did.  Just make the 200 explicit, it is worth the extra 4 characters.

PR #314, opened 2018-06-01
INT‌-593: Update collectd build for mysql8

jchengsfx:
- Updated Dockerfile to install libmysqlclient-dev from the mysql apt repo instead of from ubuntu.
- Updated package tests in circleci config to also trigger if Dockerfile is changed.
- Added install_pytest build step in circleci config to install pip with get-pip.py to avoid conflicts with ubuntu packages from apt repo when installing pytest requirements.


tests/kubernetes/services/mysql.yaml
```diff
@@ -26,3 +26,12 @@ spec:
               value: testing123
             - name: MYSQL_DATABASE
               value: testdb
+        - name: mysql-latest
```
keitwb: Run this as two separate pods (which will require two deployments).  I don't even understand how this works because pods share a network namespace so only one of the mysql instances should work since they both bind to port 3306.

jchengsfx: I'll revert this change. This test will be parametrized for different versions of mysql when the monitor tests are restructured in a future PR.

.circleci/config.yml
```diff
@@ -31,6 +31,13 @@ references:
     run: |
       docker load -i /tmp/workspace/final_image.tar
 
+  install_pytest: &install_pytest
+    run: |
+      sudo apt-get update && sudo apt-get install -y python3-dev
+      curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
```
keitwb: Try the package `python3-pip` instead.  This doesn't pin the version of pip which kind of makes me wary of this.

keitwb:
Please squash when merging.

PR #317, opened 2018-06-04
remove vendor step from goexecutor setup

charliesignalfx:


charliesignalfx:
This significantly shortened the time that lint, vet, and gotests take in circle

PR #333, opened 2018-06-11
Add adapter for new plugins

charliesignalfx:


internal/utils/telegraf/emitter/baseemitter/baseemitter.go
```diff
@@ -0,0 +1,224 @@
...
+	// You can add metrics and events to exclude by name using
+	// ExcludeDatum(string) and ExcludeData(string).  You should look up
+	// excluded events and metrics using Excluded(string)bool
+	excluded map[string]bool
+}
+
+// AddTag adds a key/value pair to all measurement tags.  If a key conflicts
+// the key value pair in AddTag will override the original key on the
+// measurement
+func (B *BaseEmitter) AddTag(key string, val string) {
```
keitwb: This is quite unconventional to use a capital letter for the method struct var.

internal/utils/telegraf/emitter/baseemitter/baseemitter.go
```diff
@@ -0,0 +1,224 @@
...
+	if len(t) > 0 {
+		return t[0]
+	}
+	return time.Now()
+}
+
+// BaseEmitter immediately converts a telegraf measurement into datapoints and
+// sends them through Output
+type BaseEmitter struct {
+	lock   *sync.Mutex
```
keitwb: Use of pointer here is unnecessary.  Default value of sync.Mutex is an unlocked mutex.

internal/utils/telegraf/emitter/baseemitter/baseemitter.go
```diff
@@ -0,0 +1,224 @@
...
+	for _, tag := range tags {
+		B.OmitTag(tag)
+	}
+}
+
+// FilterTags - filter function for util.CloneAndFilterStringMapWithFunc()
+// that will exclude any tags that
+func (B *BaseEmitter) FilterTags(key string, value string) (include bool) {
+	B.lock.Lock()
+	include = B.omittedTags[key]
```
keitwb: This is a rather gratuitous use of named return values.  The normal idiom for this is simply:
```
B.lock.Lock()
defer B.lock.Unlock()
return B.omittedTags[key]
```

which is both shorter and clearer.

keitwb: Same thing above with the `Excluded` method.

internal/utils/telegraf/emitter/baseemitter/baseemitter.go
```diff
@@ -0,0 +1,224 @@
...
+	tags map[string]string, metricType datapoint.MetricType,
+	originalMetricType string, t ...time.Time) {
+	for field, val := range fields {
+		// telegraf doc says that tags are owned by the calling plugin and they
+		// shouldn't be mutated.  So we copy the tags map
+		var metricDims = utils.CloneAndFilterStringMapWithFunc(tags, B.FilterTags)
+
+		// add additional tags to the metricDims
+		if len(B.addTags) > 0 {
+			B.lock.Lock()
```
keitwb: If this `Add` method can be called concurrently, you might gain a bit by using a `RWMutex` since this only requires locking for reading.

internal/utils/telegraf/emitter/baseemitter/baseemitter.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// Add parses measurements from telegraf and emits them through Output
+func (B *BaseEmitter) Add(measurement string, fields map[string]interface{},
+	tags map[string]string, metricType datapoint.MetricType,
+	originalMetricType string, t ...time.Time) {
+	for field, val := range fields {
+		// telegraf doc says that tags are owned by the calling plugin and they
+		// shouldn't be mutated.  So we copy the tags map
+		var metricDims = utils.CloneAndFilterStringMapWithFunc(tags, B.FilterTags)
```
keitwb: The normal idiom for this is:

`metricDim := utils.Clon...`

internal/utils/telegraf/emitter/baseemitter/baseemitter.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		parse.SetPluginDimension(measurement, metricDims)
+		parse.RemoveSFXDimensions(metricDims)
+
+		// Get the metric value as a datapoint value
+		if metricValue, err := datapoint.CastMetricValue(val); err == nil {
+			var dp = datapoint.New(
+				metricName,
+				metricDims,
+				metricValue.(datapoint.Value),
```
keitwb: Why is the type assertion here?

internal/utils/telegraf/emitter/batchemitter/batchemitter.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/signalfx/signalfx-agent/internal/utils"
+	"github.com/signalfx/signalfx-agent/internal/utils/telegraf/emitter/baseemitter"
+	"github.com/signalfx/signalfx-agent/internal/utils/telegraf/measurement"
+)
+
+// BatchEmitter gathers a batch of telegraf measurements that can be modified
+// and then emitted
+type BatchEmitter struct {
+	*baseemitter.BaseEmitter
+	lock         *sync.Mutex
```
keitwb: Unnecessary pointer.

internal/utils/telegraf/measurement/measurement.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Measurement is a storage struct for storing measurements from telegraf
+// plugins
+type Measurement struct {
+	Measurement        string
+	Fields             map[string]interface{}
+	Tags               map[string]string
+	MetricType         datapoint.MetricType
+	OriginalMetricType string
+	T                  []time.Time
```
keitwb: `T` is ambiguous.  I might guess it stood for "timestamp" but I wouldn't guess it meant a slice of times.

vendor/github.com/influxdata/telegraf/plugins/inputs/win_perf_counters/pdh.go
```diff
@@ -0,0 +1,432 @@
...
+//
+// This is the official list of 'win' authors for copyright purposes.
+//
+// Alexander Neumann <an2048@googlemail.com>
+// Joseph Watson <jtwatson@linux-consulting.us>
+// Kevin Pors <krpors@gmail.com>
+
+// +build windows
+
+package win_perf_counters
```
keitwb: Why is this pulled in?

internal/utils/telegraf/emitter/baseemitter/baseemitter.go
```diff
@@ -0,0 +1,224 @@
...
+// to exclude
+func (B *BaseEmitter) ExcludeData(names []string) {
+	for _, name := range names {
+		B.ExcludeDatum(name)
+	}
+}
+
+// Excluded - A thread safe function for checking if events or metrics should be
+// excluded from emission
+func (B *BaseEmitter) Excluded(name string) (excluded bool) {
```
keitwb: `IsExcluded` would be clearer.  Also should it be part of the `Emitter` interface?

internal/utils/telegraf/accumulator/accumulator.go
```diff
@@ -0,0 +1,58 @@
...
+	"time"
+
+	"github.com/signalfx/golib/datapoint"
+	"github.com/signalfx/signalfx-agent/internal/utils/telegraf/emitter"
+)
+
+// Accumulator is an interface used to accumulate telegraf measurements from
+// Telegraf plugins.
+type Accumulator struct {
+	Emitter emitter.Emitter
```
keitwb: This may not need to be exported if it isn't meant to be used outside of the methods on this class.

internal/utils/telegraf/emitter/baseemitter/baseemitter.go
charliesignalfx: Will rename.  It could be added to the interface, but I don't think it's something that monitors/clients will directly invoke.  I think really the only code that might invoke it outside of baseemitter is any extension of baseemitter.  In that case it is an exported Function so any packages that extend it should have access to do what they need with it behind the interface.

internal/monitors/telegraf/common/emitter/batchemitter/batchemitter.go
```diff
@@ -5,52 +5,52 @@ import (
...
+	b.lock.Unlock()
 }
 
 // NewEmitter returns a new BatchEmitter that gathers a batch of telegraf
 // measurements so they can be modified and then emitted
 func NewEmitter() *BatchEmitter {
 	return &BatchEmitter{
 		BaseEmitter:  baseemitter.NewEmitter(),
-		lock:         &sync.Mutex{},
+		lock:         sync.Mutex{},
```
keitwb: Unnecessary.

internal/monitors/telegraf/common/emitter/baseemitter/baseemitter.go
```diff
@@ -201,24 +201,23 @@ func (B *BaseEmitter) Add(measurement string, fields map[string]interface{},
...
+	b.Logger.Error(err)
 }
 
 // NewEmitter returns a new BaseEmitter
-func NewEmitter() (b *BaseEmitter) {
-	b = &BaseEmitter{
-		lock:        &sync.Mutex{},
+func NewEmitter() *BaseEmitter {
+	return &BaseEmitter{
+		lock:        sync.RWMutex{},
```
keitwb: Unnecessary.

PR #337, opened 2018-06-19
Win perf

charliesignalfx:
- refactor make file to run powershell scripts
- add perf counter monitor
- enhance emitters to rename metrics using a map[string]string
- add vendor deps


internal/monitors/telegraf/common/emitter/baseemitter/baseemitter.go
```diff
@@ -137,6 +140,35 @@ func (b *BaseEmitter) FilterTags(key string, value string) (include bool) {
 	return include
 }
 
+// RenameMetric adds a mapping to rename a metric by it's name
+func (b *BaseEmitter) RenameMetric(original string, override string) {
```
keitwb: Unused.  Also does it really need to be thread safe since metric name mapping will only happen on `Configure` and given that monitor instances are no longer reconfigured?  If not, take out the locks, esp the reader lock in `GetMetricName` since that is on a somewhat hot path and the read lock still involves mutating shared memory across threads which will kill performance at high DPM.

internal/monitors/telegraf/common/emitter/baseemitter/baseemitter.go
```diff
@@ -137,6 +140,35 @@ func (b *BaseEmitter) FilterTags(key string, value string) (include bool) {
...
+		b.RenameMetric(original, override)
+	}
+}
+
+// GetMetricName parses the metric name and takes name overrides into account
+func (b *BaseEmitter) GetMetricName(measurement string, field string, metricDims map[string]string) (name string, isSFX bool) {
+	name, isSFX = parse.GetMetricName(measurement, field, metricDims)
+
+	b.lock.RLock()
+	newName := b.nameMap[name]
```
keitwb:
```
if newName := b.nameMap[name]; newName != "" {
    name = newName
}
```

internal/monitors/telegraf/monitors/winperfcounters/win_perf_counters_windows.go
```diff
@@ -0,0 +1,74 @@
...
+	// create the accumulator
+	ac := accumulator.NewAccumulator(baseemitter.NewEmitter(m.Output, logger))
+
+	// create contexts for managing the the plugin loop
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+
+	// gather metrics on the specified interval
+	utils.RunOnInterval(m.ctx, func() {
+		if err := plugin.Gather(ac); err != nil {
+			logger.WithError(err)
```
keitwb: I think this just creates a logger instance that has that error but doesn't actually log anything out.  For that I think you need to use `.Error(...)` at the end.

internal/monitors/telegraf/monitors/winperfcounters/win_perf_counters.go
```diff
@@ -0,0 +1,74 @@
...
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `acceptsEndpoints:"false" deepcopier:"skip"`
+	Object               []Perfcounterobj `yaml:"objects" default:"[]"`
+	// number of nanoseconds that wildcards in counter paths should be expanded
+	// and how often to refresh counters from configuration
+	CountersRefreshInterval int `yaml:"counterRefreshInterval" default:"60"`
+	// if `true`, instance indexes will be included in instance names, and wildcards will
+	// be expanded and localized (if applicable).  If `false`, non partial wildcards will be expanded and instance names will not include instance indexs.
+	UseWildcardsExpansion bool
```
keitwb: Always provide yaml tag.

internal/monitors/telegraf/monitors/winperfcounters/win_perf_counters.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `acceptsEndpoints:"false" deepcopier:"skip"`
+	Object               []Perfcounterobj `yaml:"objects" default:"[]"`
+	// number of nanoseconds that wildcards in counter paths should be expanded
+	// and how often to refresh counters from configuration
+	CountersRefreshInterval int `yaml:"counterRefreshInterval" default:"60"`
+	// if `true`, instance indexes will be included in instance names, and wildcards will
+	// be expanded and localized (if applicable).  If `false`, non partial wildcards will be expanded and instance names will not include instance indexs.
```
keitwb: Really looooong line.

internal/monitors/telegraf/monitors/winperfcounters/win_perf_counters.go
```diff
@@ -0,0 +1,74 @@
...
+	UseWildcardsExpansion bool
+	// print out the configurations that match available performance counters
+	PrintValid bool `yaml:"printValid"`
+}
+
+// Monitor for Utilization
+type Monitor struct {
+	Output types.Output
+	cancel func()
+	ctx    context.Context
```
keitwb: Doesn't need to be on struct.

internal/monitors/telegraf/common/emitter/baseemitter/baseemitter.go
charliesignalfx: GetMetricName could be extended and used concurrently.  This is a base struct feature.

That said, you're right that configuration of the map of metrics to exclude will probably never be accessed concurrently.

charliesignalfx: And RenameMetric will be used in a future PR

PR #343, opened 2018-06-27
Add collectd-kong

rmfitzpatrick:


internal/monitors/collectd/kong/kong.go
```diff
@@ -0,0 +1,164 @@
...
+	monitors.Register(monitorType, func() interface{} {
+		return &Monitor{
+			*collectd.NewMonitorCore(CollectdTemplate),
+		}
+	}, &Config{})
+}
+
+// Header defines name/value pair for AuthHeader option
+type Header struct {
+    // Name of header to include with GET
```
keitwb: Run this through `go fmt`.  The tabs/spaces are off.

keitwb: Golang uses tabs for the base indent levels and spaces for alignment.

internal/monitors/collectd/kong/kong.tmpl
```diff
@@ -0,0 +1,74 @@
...
+    {{end}}{{with .ClientCert}}
+    ClientCert "{{.}}"
+    {{end}}{{with .ClientCertKey}}
+    ClientCertKey "{{.}}"
+    {{end}}{{with .IntervalSeconds}}
+    Interval {{.}}
+    {{end}}{{range $l := .Metrics -}}
+    Metric "{{$l.MetricName}}" {{$l.ReportBool}}
+    {{end}}{{with .ReportAPIIDs}}
+    ReportAPIIDs {{.}}
```
keitwb: Do all of these `Report*` options default to false in the plugin?  Because if not, I think they are only going to be configured if the user provided `true`.

tests/monitors/kong_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+        with run_container(kong_image, environment=kong_env, command='sleep inf') as migrations:
+
+            def db_is_reachable():
+                return migrations.exec_run('psql -h {} -U postgres'.format(db_ip)).exit_code == 0
+
+            assert wait_for(db_is_reachable)
+            assert migrations.exec_run('kong migrations up --v').exit_code == 0
+
+        with run_container(kong_image, environment=kong_env, ports={'8001/tcp': None}) as kong:
```
keitwb: Don't use port forwarding to the host as we are trying to decouple the tests as much as possible from the docker environment to give us maximum flexibility on CI platforms and in dev environments.  You should be able to use the kong container ip directly in the tests, just like you are doing below on line 80.

rmfitzpatrick: updated.

internal/monitors/collectd/kong/kong.go
rmfitzpatrick: updated.

internal/monitors/collectd/kong/kong.tmpl
rmfitzpatrick: No, they have various defaults (https://github.com/signalfx/collectd-kong/blob/master/README.md) and I've confirmed desired fields propagate for configuration changes and plugin defaults are used.

defaults are those of collectd, since I'm not specifying agent defaults:
```yaml
    reportRouteIds: false  # true by default
    reportHttpMethods: false  # true by default
    reportStatusCodes: true   # true by default
    reportStatusCodeGroups: false  # true by default
```

```
[0004]  INFO report_api_ids: True collectdInstance=global (default)
[0004]  INFO report_api_names: True collectdInstance=global (default)
[0004]  INFO report_http_methods: False collectdInstance=global (agent.yml)
[0004]  INFO report_route_ids: False collectdInstance=global (agent.yml)
[0004]  INFO report_service_ids: True collectdInstance=global (default)
[0004]  INFO report_service_names: True collectdInstance=global (default)
[0004]  INFO report_status_code_groups: False collectdInstance=global (agent.yml)
[0004]  INFO report_status_codes: True collectdInstance=global (agent.yml)
```

keitwb:
Wait for tests to pass before merging with squash.

PR #346, opened 2018-07-02
Host-metadata (DO NOT MERGE YET)

charliesignalfx:
DO NOT MERGE THIS YET
I'm testing changes to circle.yaml


docs/monitors/host-metadata.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+## Metrics
+
+This monitor emits the following metrics.  Note that configuration options may
+cause only a subset of metrics to be emitted.
+
+| Name | Type | Description |
+| ---  | ---  | ---         |
+| `gauge.sf.host-plugin_uptime` | gauge | The time this monitor has been running in milliseconds.  Dimensions include `signalfx_agent`, `collectd`, `kernel_release`, `kernel_version` |
```
keitwb: Dimensions should be documented using the special Dimension note comments where you can give a more complete description of them.

internal/monitors/metadata/hostmetadata/hostmetadata.go
```diff
@@ -0,0 +1,190 @@
...
+
+var logger = log.WithFields(log.Fields{"monitorType": monitorType})
+
+func init() {
+	monitors.Register(monitorType, func() interface{} { return &Monitor{Monitor: &metadata.Monitor{}} }, &Config{})
+}
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `acceptsEndpoints:"false"`
```
keitwb: Set `singleInstance:"true"` as well since this should never be configured twice.

internal/monitors/metadata/hostmetadata/hostmetadata.go
```diff
@@ -0,0 +1,190 @@
...
+	// environments.
+	ProcFSPath string `yaml:"procFSPath" default:"/proc"`
+	// The path to the main host config dir. Useful to override in
+	// containerized environments.
+	EtcPath string `yaml:"etcPath" default:"/etc"`
+}
+
+// Monitor for host-metadata
+type Monitor struct {
+	*metadata.Monitor
```
keitwb: Make it not a pointer.

internal/monitors/metadata/hostmetadata/hostmetadata.go
```diff
@@ -0,0 +1,190 @@
...
+		time.Duration(rand.Int63n(600)+86400) * time.Second,
+	}
+
+	// set plugin start time
+	m.startTime = time.Now()
+
+	// set HOST_PROC and HOST_ETC for gopsutil
+	if conf.ProcFSPath != "" {
+		if err := os.Setenv("HOST_PROC", conf.ProcFSPath); err != nil {
+			logger.Errorf("Error setting HOST_PROC env var %v", err)
```
keitwb: Shouldn't these errors cause the whole `Configure` method to return an error instead of just continuing on?

internal/monitors/metadata/hostmetadata/hostmetadata.go
```diff
@@ -0,0 +1,190 @@
...
+			logger.Errorf("Error setting HOST_ETC env var %v", err)
+		}
+	}
+
+	// create contexts for managing the the plugin loop
+	var ctx context.Context
+	ctx, m.cancel = context.WithCancel(context.Background())
+
+	// gather metadata on intervals
+	utils.RunOnIntervals(ctx,
```
keitwb: I really don't like the name being different by only one letter at the end.  Can you call it something like `RunOnIrregularIntervals` or something like that to make it more distinct?

internal/monitors/metadata/hostmetadata/hostmetadata.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		properties := meta.ToStringMap()
+		for k, v := range properties {
+			m.EmitProperty(k, v)
+		}
+	}
+}
+
+func (m *Monitor) getUptime(curr time.Time) int64 {
+	return int64((curr.UnixNano() - m.startTime.UnixNano()) / 1000000)
```
keitwb: This can just return a time.Duration and then you can use the `.Milliseconds()` method on it to get the milliseconds instead of doing division here and returning an ambiguous integer value.

internal/utils/time.go
```diff
@@ -59,3 +59,53 @@ func RunOnInterval(ctx context.Context, fn func(), interval time.Duration) {
...
+
+				if len(intv) > 1 {
+					// advance the interval
+					intv = intv[1:]
+				} else {
+					// evaluate repeat policies
+					if repeatPolicy == RepeatNone {
+						return
+					} else if repeatPolicy == RepeatAll {
+						intv = intervals[:] // copy the original interval list
```
keitwb: Don't assume the original intervals slice's elements are immutable.  A slice only points to a backing array and that can change due to logic outside of this function that reuses the `intervals` slice that was passed in, such as reassigning distinct elements in the slice.  You may not be doing that in the monitor code now, but it is a potential source of a very tricky bug.

I would just stick with the copy in `intv` that you have and use an index variable to keep track of where you are in it instead of reassigning it.  That copy is good because it happens before the function returns.  If it is `RepeatLast` you can just leave the index stuck at the end forever.

internal/utils/time.go
```diff
@@ -59,3 +59,53 @@ func RunOnInterval(ctx context.Context, fn func(), interval time.Duration) {
...
+					}
+				}
+				timer.Reset(intervals[0])
```
keitwb: Shouldn't this be `intv[0]`?

internal/utils/time.go
```diff
@@ -59,3 +59,53 @@ func RunOnInterval(ctx context.Context, fn func(), interval time.Duration) {
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-timer.C:
+
+				if len(intv) > 1 {
+					// advance the interval
+					intv = intv[1:]
+				} else {
```
keitwb: Doesn't this skip the last interval for repeat policy of `None` or `All` since you are resetting the time _after_ the slice has been pared down?  So when there is one interval remaining in the slice, the `else` block will be hit and then it will return or reset `intv` with the last element ignored.

Maybe add some unit tests to test all of the various policies.

charliesignalfx: No, it does execute.  The timer is initialized with intv[0].  We do not pare down the interval array until after the timer fires.  Then we evaluate based on the current length of the interval what to do.  If the len(intv) > 1 we advance and reset the timer.  If the len(intv) == 1 it means we've already run the last interval through the timer.  That's when the reset policies are evaluated.  I could change the first condition to `if (len(intv) >= 2){} else{}` if that's clearer?

charliesignalfx: I think I understand what you're saying here, yes you're right because the function is executed after the evaluation of the repeat policy it would effectively skip it.

PR #352, opened 2018-07-06
Updated Get to return nil string if key not found

jchengsfx:


internal/core/services/rules.go
```diff
@@ -35,7 +35,7 @@ func get(args ...interface{}) (interface{}, error) {
 		return val, nil
 	}
 
-	return nil, errNoValueFound
+	return "", nil
```
keitwb: Hmm..ok since the return type is `interface{}` let's make it return `nil, nil` instead.

internal/core/services/rules.go
```diff
@@ -35,18 +35,17 @@ func get(args ...interface{}) (interface{}, error) {
...
 	"Contains": func(args ...interface{}) (interface{}, error) {
-		if _, err := get(args...); err != nil {
-			// drop error if no value found
-			if err == errNoValueFound {
-				return false, nil
-			}
+		val, err := get(args...)
+		if err != nil {
 			return false, err
+		} else if val == nil {
```
keitwb: Do this instead:

```
if err != nil {
    return false, err
}
return val != nil, nil
```

internal/core/services/rules_test.go
```diff
@@ -30,10 +30,10 @@ func TestMapFunctions(t *testing.T) {
 	})
 	t.Run("Get() map does not contain key", func(t *testing.T) {
 		val, err := ruleFunctions["Get"](interfacemap, "nokey")
-		assert.Error(t, err, "should error out if the map does not contain the desired value")
+		assert.NoError(t, err, "should not error out if the map does not contain the desired value")
 		assert.NotEqual(t, "world", val, "should return the expected value")
```
keitwb: Assert that it is `nil` instead of just that it isn't "world".

PR #355, opened 2018-07-11
Account for -e and pipefail for diff check and update ansible test

rmfitzpatrick:
The changes-include-dir check was exiting early on the initial grep when no matches were found.  Since we only care about the return value for grep the pipefail is also unnecessary.

### DIFF
--- a/scripts/changes-include-dir
+++ b/scripts/changes-include-dir
@@ -12,8 +12,7 @@ set -euo pipefail
 DIRS=$@
 
 for DIR in $DIRS; do
-    git diff origin/master...$CIRCLE_BRANCH --name-status | awk '{print $2}' | grep "^$DIR"
-    if [ $? -eq 0 ]; then
+    if [ $(git diff signalfx/master...$CIRCLE_BRANCH --name-status | awk '{print $2}' | grep -c "^$DIR") -gt 0 ]; then
         exit 0
     fi
 done
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -406,7 +406,8 @@ jobs:
           make dev-image
           docker run --rm \
             signalfx-agent-ansible-dev \
-            ansible-playbook -i inventory example-playbook.yml --connection=local > ~/testresults/ansible.out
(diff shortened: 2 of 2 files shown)


scripts/changes-include-dir
```diff
@@ -7,7 +7,7 @@
 # This compares the current branch to the master, so it might have some false
 # positives if using multiple layers of branches off of master.
 
-set -euo pipefail
+set -u
```
keitwb: I would be better to just put the command that tests for the diff in the `if` statement directly, which skips the `-e` check (as well as `pipefail` since that only affects the exit code of the pipeline).

keitwb: See the rules at https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html#The-Set-Builtin

rmfitzpatrick:
@keitwb, good idea since more functionality could be added where these options are desired.  Updated.

scripts/changes-include-dir
```diff
@@ -12,8 +12,7 @@ set -euo pipefail
 DIRS=$@
 
 for DIR in $DIRS; do
-    git diff origin/master...$CIRCLE_BRANCH --name-status | grep $DIR
-    if [ $? -eq 0 ]; then
+    if [ $(git diff signalfx/master...$CIRCLE_BRANCH --name-status | grep -c $DIR) -gt 0 ]; then
```
keitwb: Can't this just be `if git diff signalfx/master...$CIRCLE_BRANCH --name-status | grep -c $DIR; then`?

rmfitzpatrick: That doesn't swallow the stdout and leaves some litter:

```
CIRCLE_BRANCH=update_ansible_test scripts/changes-include-dir deployments asdf asdfasga asdgasdf cgggg adsfasdf changes
0
0
0
0
0
0
1
```

rmfitzpatrick:
Updated to include c3bce24aab396b6b2047ad0505bf1c824ecc6601

scripts/changes-include-dir
rmfitzpatrick: updated

PR #357, opened 2018-07-11
Pass through the InstanceFrom config to GenericJMX

keitwb:
This is a breaking change, but minor so we will release a new minor version.

### DIFF
--- a/selfdescribe.json
+++ b/selfdescribe.json
@@ -556,10 +556,10 @@
                     {
                       "yamlName": "instanceFrom",
                       "doc": "",
-                      "default": "",
+                      "default": null,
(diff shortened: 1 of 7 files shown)


selfdescribe.json
```diff
@@ -556,10 +556,10 @@
                     {
                       "yamlName": "instanceFrom",
                       "doc": "",
-                      "default": "",
+                      "default": null,
```
rmfitzpatrick: Does it matter that we're moving to null instead of the empty list?

keitwb: That is the default value for a slice in Go. It is functionally equivalent to an empty list.

PR #360, opened 2018-07-12
K8s test improvements

jchengsfx:


tests/kubernetes/utils.py
```diff
@@ -498,8 +499,11 @@ def container_is_running(client, name):
             raise
 
 def pod_port_open(container, host, port):
-    rc, _ = container.exec_run("nc -z %s %d" % (host, port))
-    return rc == 0
+    try:
+        rc, _ = container.exec_run("nc -z %s %d" % (host, port))
+        return rc == 0
+    except ConnectionError as e:
```
keitwb: Hmm..looking at this more closely it seems like that could only be caused by the connection to the docker engine in the minikube instance being broken, right?  If so, this doesn't seem like a very appropriate place to catch that exception since it is a sign that something is really wrong.

jchengsfx: I noticed a few intermittent ConnectionError failures in circleci for both the docker engine on the host and the docker engine in minikube (see https://circleci.com/gh/signalfx/signalfx-agent/13266 as an example).  Not sure if this is related to resource contention issues or with python/requests/docker-py within the circleci environment, but I haven't run into any of these failures yet with these changes.  I can do more testing and try to see if it can be reproduced and if catching the exception actually helps.

keitwb:
@jchengsfx what is the status of this?

jchengsfx:
Most of it is already in master with other merged PRs.  I'll close this and re-open when ready.

PR #361, opened 2018-07-16
Add K8s resource quota metrics

keitwb:
To add integration tests for this I had to do some refactoring of the
k8s test harness to accept generic resources and also to do more
detailed assertions against the metrics received.


tests/helpers/assertions.py
```diff
@@ -41,6 +41,23 @@ def has_datapoint_with_dim(fake_services, key, value):
     return has_datapoint_with_all_dims(fake_services, {key: value})
 
 
+def has_datapoint(fake_services, metric_name=None, dimensions=None, value=None):
+    """
+    Returns True if there is a datapoint seem in the fake_services backend that
+    has the given attributes.  If a property if not specified it will not be
```
rmfitzpatrick: nitpicks:
s/seem/seen
if a property <is> not

PR #372, opened 2018-07-23
Add yarn support to spark monitor and add hadoop monitors

charliesignalfx:


internal/monitors/collectd/hadoopjmx/hadoopjmx.go
```diff
@@ -0,0 +1,143 @@
...
+type Monitor struct {
+	*genericjmx.JMXMonitorCore
+}
+
+// Configure configures the hadoopjmx monitor and instantiates the generic jmx
+// monitor
+func (m *Monitor) Configure(conf *Config) error {
+	// create the mbean map with the appropriate mbeans for the given node type
+	var newMBeans genericjmx.MBeanMap
+	switch conf.NodeType {
```
keitwb: The YAML decoder doesn't actually raise any errors if `nodeType` is not in your set of constants, so you might want to add a `Validate` method to the `Config` struct so that if a user makes a typo they don't get very confused on why nothing is working.

PR #379, opened 2018-07-26
add hadoop integration test for 2.x and 3.x

charliesignalfx:


tests/monitors/hadoop/hadoop.py
```diff
@@ -0,0 +1,68 @@
...
+  - type: collectd/hadoop
+    host: $host
+    port: $port
+    verbose: true
+""")
+
+
+def distribute_hostnames(containers):
+    """
+    iterate over each container and pass it's hostname and ip to etc host on
```
keitwb: _it's_

PR #382, opened 2018-07-26
Updates for monitoring kafka

asuresh4:
Updates
- GenericJMX plugin
   - add CustomDimension config option
   - use CustomDimension to populate monitorID
   - add mbeansToOmit config option
   - collect MBean properties as dimensions
 - Kafka monitor
   - add new default metrics
   - add new config option for cluster name
- Add monitors for monitoring java based kafka consumers and producers


docs/monitors/collectd-kafka_consumer.md
```diff
@@ -0,0 +1,97 @@
...
+    host: localhost
+    port: 9099
+    mBeansToOmit:
+      - fetch-size-avg-per-topic
+```
+
+Note that this monitor requires Kafka v0.9.0.0 or above and collects metrics from the new consumer API.
+Also, per topic metrics that are collected by default are not available through the new consumer API in
+v0.9.0.0 which can cause the logs to flood with warnings related to MBean not being found.
+Use `mBeansToOmit` config option in such cases. The above example configuration will not attempt to
```
keitwb: "Use the ..."

docs/monitors/collectd-kafka_consumer.md
```diff
@@ -0,0 +1,97 @@
...
+collect the MBean referenced by `fetch-size-avg-per-topic`. Here is a
+[list] (https://github.com/signalfx/signalfx-agent/tree/master/internal/monitors/collectd/kafka/mbeans.go)
```
keitwb: Shouldn't this link point to the kafka-consumer monitor package instead?

docs/monitors/collectd-kafka_consumer.md
```diff
@@ -0,0 +1,97 @@
...
+monitors:
+  - type: collectd/kafka_consumer
+    host: localhost
+    port: 9099
+    mBeansToOmit:
+      - fetch-size-avg-per-topic
+```
+
+Note that this monitor requires Kafka v0.9.0.0 or above and collects metrics from the new consumer API.
+Also, per topic metrics that are collected by default are not available through the new consumer API in
```
keitwb: "per-topic"

docs/monitors/collectd-kafka_consumer.md
```diff
@@ -0,0 +1,97 @@
...
+v0.9.0.0 which can cause the logs to flood with warnings related to MBean not being found.
```
keitwb: "related to the MBean ..."

tests/monitors/kafka/kafka_test.py
```diff
@@ -28,22 +28,73 @@ def run_kafka(version="1.0.1"):
...
+               clusterName: testCluster
+               mBeansToOmit:
+                 - kafka-active-controllers
+            """.format(kafkahost))) as [backend, _, _]:
+                assert not wait_for(p(has_datapoint_with_metric_name, backend, "gauge.kafka-active-controllers")), \
+                        "Didn't get kafka datapoints"
+
+
+def test_all_kafka_monitors():
+    for version in versions:
```
keitwb: This should be done with pytest parameterization instead so that the tests can run in parallel.  See for example https://github.com/signalfx/signalfx-agent/blob/master/tests/monitors/hadoop/hadoop_test.py#L40

PR #387, opened 2018-08-02
Service tests

jchengsfx:
- Added service and system (container) restart tests.
- Added installer_tests to circleci workflow.
- Marked flaky integration tests to allow re-runs.
- Includes an update to the sysv/upstart service files that pins the names of the pid and log files to "signalfx-agent" on system reboot (see https://signalfuse.atlassian.net/browse/INT-689).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,6 +249,41 @@ jobs:
       - store_artifacts:
           path: ~/testresults/k8s_integration_tests.html
 
+  installer_tests:
+    <<: *executor
+    steps:
+      - checkout
+      - run: |
+          if ! scripts/changes-include-dir deployments/installer; then
+              echo "Installer code has not changed, skipping tests!"
+              touch ~/.skip
+              exit 0
+          fi
+      - *install_pytest
+      - run:
+          no_output_timeout: 30m
+          command: |
+            if [ -f ~/.skip ]; then
+                echo "Installer code has not changed, skipping tests!"
+                exit 0
+            fi
+            mkdir /tmp/scratch
+            mkdir ~/testresults
+            pytest \
+                -n auto \
+                -m "installer" \
+                --verbose \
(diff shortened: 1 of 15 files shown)


tests/packaging/images/Dockerfile.debian-9-stretch
```diff
@@ -1,7 +1,7 @@
 FROM debian:stretch-slim
 
 RUN apt-get update &&\
-    apt-get install -yq ca-certificates procps systemd wget libcap2-bin apt-transport-https
+    apt-get install -yq ca-certificates procps systemd wget libcap2-bin apt-transport-https systemd-sysv
```
keitwb: What is this for? We shouldn't really be using sysv on Debian 9 since it already has systemd.

.circleci/config.yml
```diff
@@ -249,6 +249,41 @@ jobs:
       - store_artifacts:
           path: ~/testresults/k8s_integration_tests.html
 
+  installer_tests:
+    <<: *executor
+    steps:
+      - checkout
+      - run: |
+          if ! scripts/changes-include-dir deployments/installer tests/packaging; then
```
keitwb: Scope the changes to the specific installer script pytest module instead of anything in the packaging test dir.  Apparently the script supports handling filenames instead of just dirs, so perhaps rename the script to `changes-include-path` or something less specific to directories.

tests/packaging/images/Dockerfile.debian-9-stretch
jchengsfx: I can revert that... I added that package to install /sbin/shutdown when I was testing running "shutdown" in the container for the system restart tests as a workaround for the docker restart issues.

jchengsfx:
Also includes an update to the sysv/upstart service files that pins the names of the pid and log files to "signalfx-agent" on system reboot.

PR #391, opened 2018-08-15
add processlist monitor

charliesignalfx:


internal/monitors/processlist/processlist.go
```diff
@@ -0,0 +1,113 @@
...
+	_, err = compressor.Write(in)
+	_ = compressor.Close()
+	return
+}
+
+func toTime(secs float64) (response string) {
+	minutes := int(secs / 60)
+	seconds := int(math.Mod(secs, 60.0))
+	sec := seconds
+	dec := (seconds - sec) * 100
```
keitwb: Since `sec == seconds` due to the preceding line, this will always be 0.

internal/monitors/processlist/processlist.go
```diff
@@ -0,0 +1,113 @@
...
+			if err != nil {
+				logger.Error(err)
+				return
+			}
+
+			// escape and compress the process list
+			escapedBytes := bytes.Replace(processList.Bytes(), []byte{byte('\\')}, []byte{byte('\\'), byte('\\')}, -1)
+			cbytes, err := compressBytes(escapedBytes)
+			if err != nil {
+				logger.Error(err)
```
keitwb: `logger.WithError(err).Error("Couldn't compress process list")` would be better.

internal/monitors/processlist/processlist.go
```diff
@@ -0,0 +1,113 @@
...
+	var ctx context.Context
+	ctx, m.cancel = context.WithCancel(context.Background())
+
+	utils.RunOnInterval(
+		ctx,
+		func() {
+			// get the process list
+			processList, err := ProcessList()
+			if err != nil {
+				logger.Error(err)
```
keitwb: `logger.WithError(err).Error("Couldn't get process list")` would be better.

internal/monitors/processlist/processlist.go
```diff
@@ -0,0 +1,113 @@
...
+	_ = compressor.Close()
+	return
+}
+
+func toTime(secs float64) (response string) {
+	minutes := int(secs / 60)
+	seconds := int(math.Mod(secs, 60.0))
+	sec := seconds
+	dec := (seconds - sec) * 100
+	response = fmt.Sprintf("%02d:%02d.%02d", minutes, sec, dec)
```
keitwb: There is no way you are going to convince me that named return values are better than just doing `return fmt.Sprintf(...` here.

internal/monitors/processlist/processlist.go
```diff
@@ -0,0 +1,113 @@
...
+type Config struct {
+	config.MonitorConfig `acceptsEndpoints:"false"`
+}
+
+func init() {
+	monitors.Register(monitorType, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+func compressBytes(in []byte) (buf bytes.Buffer, err error) {
+	compressor := zlib.NewWriter(&buf)
```
keitwb: This would probably make very little practical difference in the wild, but you can instantiate the zlib writer once and stick it on an instance field and then use the `Reset` method each read interval to avoid doing an allocation per read cycle.

internal/monitors/processlist/processlist_windows.go
```diff
@@ -0,0 +1,172 @@
...
+// ProcessList takes a snapshot of running processes and returns a byte buffer
+func ProcessList() (*bytes.Buffer, error) {
+	processes := &bytes.Buffer{}
+	processes.WriteString("{")
+	defer processes.WriteString("}") // always close the associative array
+
+	// Get all processes
+	ps, err := getAllProcesses()
+	if err != nil {
+		logger.Debugf("no processes returned %v", err)
```
keitwb: This indicates something pretty bad going on right since there should be at least one process running in the system?

internal/monitors/processlist/processlist_windows.go
```diff
@@ -0,0 +1,172 @@
...
+		if percent, inMap := cpuPercentages[p.ProcessID]; inMap {
+			cpuPercent = float64(percent)
+		}
+
+		// Memory Percent
+		var memPercent float64
+		if systemMemory, err := mem.VirtualMemory(); err == nil {
+			memPercent = 100 * float64(p.WorkingSetSize) / float64(systemMemory.Total)
+		} else {
+			logger.Debugf("unable to collect system memory total", err)
```
keitwb: no `%s` in the format string to stick the err in.  Also doesn't this warrant an error message?

internal/monitors/processlist/processlist_windows.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	// index position to stop appending commas to the list of processes
+	stop := len(ps) - 1
+
+	// iterate over each process and build an entry for the process list
+	for index, p := range ps {
+		username, err := getUsername(p.ProcessID)
+		if err != nil {
+			logger.Debugf("unable to collect use name for process %v %v", p, err)
```
keitwb: Typo with `use`

internal/monitors/processlist/processlist_windows.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// PerfProcProcess is a performance process struct used for wmi calls
+// https://msdn.microsoft.com/en-us/library/aa394323(v=vs.85).aspx
+type PerfProcProcess struct {
+	IDProcess            uint32
+	PercentProcessorTime uint64
+}
+
+// getCPUPercentages is set as a package variable so we can mock it durring testing
```
keitwb: `durring`

docs/monitors/processlist.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+
+Monitor Type: `processlist`
+
+[Monitor Source Code](https://github.com/signalfx/signalfx-agent/tree/master/internal/monitors/processlist)
+
+**Accepts Endpoints**: No
+
+**Multiple Instances Allowed**: Yes
```
keitwb: This should probably not have multiple instances (it is controlled by a struct tag on the embedded monitor config in the monitor config struct definition).

internal/monitors/processlist/processlist.go
```diff
@@ -0,0 +1,114 @@
...
+	_, err = zlibCompressor.Write(in)
+	_ = zlibCompressor.Close()
+	return
+}
+
+// toTime returns the given seconds as a formatted string "min:sec.dec"
+func toTime(secs float64) string {
+	minutes := int(secs) / 60
+	seconds := math.Mod(secs, 60.0)
+	dec := math.Mod(seconds, 1.0) * 100
```
keitwb: This is quite suspect -- the remainder of `seconds/1.0` would always be 0 right since 1 is a factor of all real numbers.  Did you mean to use `math.Modf`?

PR #397, opened 2018-08-20
Add Python runner that runs collectd plugins

keitwb:
This is fairly basic right now and is based on the work done last year
called "neopy".

It works only on Linux right now but shouldn't be too hard to port to
Windows.


internal/monitors/collectd/redis/redis.tmpl
```diff
@@ -6,7 +6,7 @@ LoadPlugin "python"
   <Module redis_info>
     Host "{{.Host}}"
     Port {{.Port}}
-    Instance "{{.Name}}[monitorID={{.MonitorID}}]"
+	Instance "{{if .Name }}{{.Name}}{{else}}{{.Host}}:{{.Port}}{{end}}[monitorID={{.MonitorID}}]"
```
rmfitzpatrick: nit: looks like you're not expanding tabs in this update.

Dockerfile
```diff
@@ -338,6 +326,7 @@ RUN apt update &&\
 ENV SIGNALFX_BUNDLE_DIR=/bundle \
     TEST_SERVICES_DIR=/go/src/github.com/signalfx/signalfx-agent/test-services \
     AGENT_BIN=/go/src/github.com/signalfx/signalfx-agent/signalfx-agent \
+	PYTHONPATH=/go/src/github.com/signalfx/signalfx-agent/python \
```
rmfitzpatrick: nit: tab

Dockerfile
```diff
@@ -197,25 +196,16 @@ RUN pip install yq &&\
 
 RUN apt install -y libffi-dev libssl-dev build-essential python-dev libcurl4-openssl-dev
 
-#COPY scripts/install-dd-plugin-deps.sh /opt/
-
-#RUN mkdir -p /opt/dd &&\
-    #cd /opt/dd &&\
-    #git clone --depth 1 --single-branch https://github.com/DataDog/dd-agent.git &&\
```
rmfitzpatrick: I noticed neopy/datadog.py is removed w/ no port as well.  Is removing dd support an intended part of this or is it provided through another mechanism?

keitwb: The neopy thing had been broken since the v3 release and nobody is using it (it was my hackathon project last year).  It would be pretty easy to reenable if desired though.

PR #402, opened 2018-08-23
Combined win_perf_counters tests into single parametrized test

jchengsfx:


tests/helpers/util.py
```diff
@@ -84,13 +84,13 @@ def run_agent_with_fake_backend(config_text, fake_services):
...
 
         def pull_output():
-            if sys.platform == 'win32':
-                b, _ = proc.communicate()
-                output.write(b)
-                return
             while True:
                 # If any output is waiting, grab it.
-                ready, _, _ = select.select([proc.stdout], [], [], 0)
+                if sys.platform != 'win32':
```
keitwb: Since this is run in a separate thread I think it can actually just block on the read below and doesn't need any of the select/ready logic.

jchengsfx: Agreed.

tests/monitors/telegraf/win_perf_counters_test.py
```diff
@@ -0,0 +1,108 @@
...
+            "win_net.Bytes_Sent_persec",
+            "win_net.Bytes_Total_persec",
+            "win_net.Current_Bandwidth",
+            "win_net.Packets_persec",
+        ],
+    },
+}
+
+
+@pytest.mark.parametrize("measurement", sorted(params.keys()))
```
rmfitzpatrick: Would using fixtures make this a bit more maintainable?

```python
@pytest.fixture
def win_cpu_config():
    config = config_template.substitute(...)
    return config, include_total, metrics

@pytest.fixture(params=('win_cpu_config',))
def win_config(request)
    yield request.getfixturevalue(request.param)

def test_win_perf_counter(win_config):
    config, include_total, metrics = win_config
    ...
```

PR #403, opened 2018-08-23
fix issue preventing the dogstatsd listener from listening for metrics

charliesignalfx:


docs/monitors/collectd-signalfx-metadata.md
```diff
@@ -28,10 +28,11 @@ Monitor Type: `collectd/signalfx-metadata`
...
-| `dogStatsDPort` | no | `unsigned integer` | Set this to a non-zero value to enable the DogStatsD listener as part of this monitor.  The listener will accept metrics on the DogStatsD format, and sends them as SignalFx datapoints to our backend.  **Note: The listener emits directly to SignalFx and will not be subject to filters configured with the SignalFx Smart Agent.  Internal stats about the SignalFx Smart Agent will not reflect datapoints set through the DogStatsD listener** (**default:** `0`) |
+| `dogStatsDPort` | no | `unsigned integer` | Set this to a non-zero value to enable the DogStatsD listener as part of this monitor.  The listener will accept metrics on the DogStatsD format, and sends them as SignalFx datapoints to our backend.  **Note: The listener emits directly to SignalFx and will not be subject to filters configured with the SignalFx Smart Agent.  Internal stats about the SignalFx Smart Agent will not reflect datapoints set through the DogStatsD listener** |
```
keitwb: This will technically allow a zero value as well right?  Not sure there is any practical value in mentioning it to users since they probably want a fixed port though.

tests/helpers/assertions.py
```diff
@@ -95,6 +95,8 @@ def has_log_message(output, level="info", message=""):
             return True
     return False
 
+def regex_search_matches_output(output, search):
```
keitwb: can you call it `get_output` instead of `output`?

PR #415, opened 2018-09-07
Allow multiple monitors per port in docker label config

keitwb:
This should be completely backwards compatible with the existing scheme


internal/observers/docker/docker.go
```diff
@@ -84,9 +83,45 @@ const (
...
-// considered for this endpoint, and thus, no agent configuration can be used
-// anyway.
+// container.  If a `monitorType` label is present, **no discovery rules will
+// be considered for this endpoint**, and thus, no agent configuration can be
+// used anyway.
+//
+// ### Multiple Monitors per Port
+// If you want to configure multiple monitors per port, you can specify the
+// port name in the form `<port number>-<port name>` instead of just the port
+// number.  For example, if you had two different Prometeus exporters running
```
charliesignalfx: typo: Prometheus

tests/observers/docker_test.py
```diff
@@ -71,11 +71,38 @@ def test_docker_observer_labels_partial():
...
+                    "agent.signalfx.com.config.80-nginx2.intervalSeconds": "1",
+                    "agent.signalfx.com.config.80-nginx2.extraDimensions": "{app: other}",
+                }):
+            assert wait_for(p(has_datapoint_with_dim, backend, "plugin", "nginx")), "Didn't get nginx datapoints"
+            assert wait_for(p(has_datapoint_with_dim, backend, "app", "nginx")), "Didn't get extra dims"
+            assert wait_for(p(has_datapoint_with_dim, backend, "app", "other")), "Didn't get extra dims"
+        # Let nginx be removed by docker observer and collectd restart
+        time.sleep(5)
+        backend.datapoints.clear()
+        assert ensure_always(lambda: not has_datapoint_with_dim(backend, "plugin", "nginx"), 10)
```
rmfitzpatrick: Unrelated to your changes, but is this actually testing something or ensuring valid `run_service` teardown?  If the latter, should it be moved to the exit logic for that util instead of used in every test?

internal/observers/docker/labels.go
```diff
@@ -9,18 +9,23 @@ import (
 )
 
 var labelConfigRegexp = regexp.MustCompile(
-	`^agent.signalfx.com.` +
+	`^agent.signalfx.com\.` +
```
rmfitzpatrick: is this a breaking change?  I noticed you had to update the test labels.

tests/observers/docker_test.py
keitwb: It's ensuring that when you stop a service (nginx in this case) that the agent discovers that it has gone down and then stops monitoring it.  The way the test knows if the monitoring has stopped is by whether it gets datapoints or not.

internal/observers/docker/labels.go
keitwb: It was always documented with the dots but was allowing anything through inadvertently.  So anybody using slashes was going against the docs.

PR #417, opened 2018-09-10
Print datapoints and events in cleaner format

jchengsfx:
https://signalfuse.atlassian.net/browse/INT-778

### DIFF
--- a/tests/helpers/util.py
+++ b/tests/helpers/util.py
@@ -15,8 +15,7 @@ import netifaces as ni
 import yaml
 
 from . import fake_backend
-from signalfx.generated_protocol_buffers \
-    import signal_fx_protocol_buffers_pb2 as sf_pbuf
+from .formatting import print_dp_or_event
 
 AGENT_BIN = os.environ.get("AGENT_BIN", "/bundle/bin/signalfx-agent")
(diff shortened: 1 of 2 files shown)


tests/helpers/util.py
```diff
@@ -71,6 +73,65 @@ def run_agent(config_text):
             yield [fake_services, get_output, c]
 
 
+def print_dp_or_event(dp_or_event):
```
keitwb: I would put this in another file called `formatting.py` or something like that.  The util module is already pretty overloaded with stuff.

tests/helpers/util.py
```diff
@@ -71,6 +73,65 @@ def run_agent(config_text):
...
+
+    def _get_pretty_props(props):
+        str_props = ""
+        for prop in sorted(props, key=lambda prop: prop.key):
+            str_props += "\n - " + prop.key + " = " + _get_pretty_value(prop.value)
+        return str_props
+
+    assert type(dp_or_event) in [sf_pbuf.DataPoint, sf_pbuf.Event], "unsupported type '%s'!" % type(dp_or_event)
+    str_dp_or_event = ""
+    if type(dp_or_event) is sf_pbuf.DataPoint:
```
keitwb: `isinstance(dp_or_event, sf_pbuf.DataPoint)` I think

PR #419, opened 2018-09-11
Make applicable monitors run on the python runner

charliesignalfx:


internal/monitors/collectd/consul/consul.go
```diff
@@ -29,33 +33,73 @@ const monitorType = "collectd/consul"
...
 // Configure configures and runs the plugin in collectd
-func (am *Monitor) Configure(conf *Config) error {
-	return am.SetConfigurationAndRun(conf)
+func (m *Monitor) Configure(conf *Config) error {
+	conf.pyConf = &python.Config{
+		MonitorConfig: conf.MonitorConfig,
+		Host:          conf.Host,
+		Port:          conf.Port,
+		ModuleName:    "consul_plugin",
+		ModulePaths:   []string{filepath.Join(os.Getenv(constants.BundleDirEnvVar), "plugins", "collectd", "consul")},
```
keitwb: This filepath construction gets repeated a lot.  Can you make some kind of function, in some appropriate place (not the utils package), that can be called something like `makeCollectdPath("consul")` so that this is shorter and more centralized as far as where this directory actually lives.

internal/monitors/collectd/kong/kong.go
```diff
@@ -157,12 +164,172 @@ type Config struct {
...
+	if conf.ReportRouteIDs != nil {
+		conf.pyConf.PluginConfig["ReportRouteIDs"] = *conf.ReportRouteIDs
+	}
+	if conf.ReportHTTPMethods != nil {
+		conf.pyConf.PluginConfig["ReportHTTPMethods"] = *conf.ReportHTTPMethods
+	}
+	if conf.ReportStatusCodeGroups != nil {
+		conf.pyConf.PluginConfig["ReportStatusCodeGroups"] = *conf.ReportStatusCodeGroups
+	}
+	if conf.ReportStatusCodes != nil {
```
keitwb: Since `nil` isn't a valid value for collectd plugin config values, why don't we just use `nil` to mean not present and then you can just put all of these pointers into the `pluginConfig` literal map above and make this a lot cleaner.

If the python runner doesn't already, it will need to filter out nil values.  This should be safe to do since you can't put `nil`/`null` into a python plugin config anyway according to https://collectd.org/documentation/manpages/collectd-python.5.shtml#config.

internal/monitors/collectd/kong/kong.go
```diff
@@ -157,12 +164,172 @@ type Config struct {
...
+		conf.pyConf.PluginConfig["ReportStatusCodes"] = *conf.ReportStatusCodes
+	}
+	if len(conf.APIIDs) > 0 {
+		conf.pyConf.PluginConfig["APIIDs"] = map[string]interface{}{
```
keitwb: If it doesn't already, the thing in Python that processes the `#flatten` directives can easily handle the case where `values` is an empty array or null, in which case you can put all of these into the `PluginConfig` literal map above and get rid of the conditionals.

internal/monitors/collectd/rabbitmq/rabbitmq.go
```diff
@@ -54,12 +62,61 @@ type Config struct {
...
+	}
+
+	if conf.VerbosityLevel != "" {
+		conf.pyConf.PluginConfig["VerbosityLevel"] = conf.VerbosityLevel
+	}
+
+	// the python runner's templating system does not convert to map first
+	// this requires TitleCase template values.  For BrokerName we accept
+	// either upper or lower case values.  Converting the map to yaml
+	// and explicitly rendering the BrokerName will allow for upper or lower
```
keitwb: How exactly does this allow for upper casing?

internal/monitors/collectd/spark/spark.go
```diff
@@ -89,10 +100,37 @@ func (c *Config) Validate() error {
...
+
+	if conf.IsMaster {
+		conf.pyConf.PluginConfig["Master"] = "http://{{.Host}}:{{.Port}}"
+		conf.pyConf.PluginConfig["MasterPort"] = conf.Port
+	} else {
+		conf.pyConf.PluginConfig["WorkerPorts"] = conf.Port
+	}
+
+	if conf.ClusterType != sparkYarn {
+		conf.pyConf.PluginConfig["MetricsURL"] = "http://{{.Host}}"
```
keitwb: This won't get rendered as a template like this.

internal/monitors/collectd/spark/spark.go
```diff
@@ -89,10 +100,37 @@ func (c *Config) Validate() error {
...
+			"Host":            conf.Host,
+			"Port":            conf.Port,
+			"Cluster":         string(conf.ClusterType),
+			"Applications":    conf.CollectApplicationMetrics,
+			"EnhancedMetrics": conf.EnhancedMetrics,
+		},
+	}
+
+	if conf.IsMaster {
+		conf.pyConf.PluginConfig["Master"] = "http://{{.Host}}:{{.Port}}"
```
keitwb: Templating doesn't get rendered.

internal/monitors/collectd/healthchecker/healthchecker.go
```diff
@@ -71,10 +76,38 @@ func (c *Config) Validate() error {
...
+		PluginConfig: map[string]interface{}{
+			"Instance": conf.Name,
+		},
+	}
+
+	if conf.TCPCheck {
+		conf.pyConf.PluginConfig["URL"] = conf.Host
+		conf.pyConf.PluginConfig["TCP"] = conf.Port
+	} else {
+		conf.pyConf.PluginConfig["URL"] = "http{{if .UseHTTPS}}s{{end}}://{{.Host}}:{{.Port}}:{{.Path}}"
```
keitwb: Templating in here.

PR #420, opened 2018-09-11
add ms sql server monitor

charliesignalfx:


Gopkg.lock
```diff
@@ -1077,108 +888,29 @@
...
   name = "k8s.io/kubernetes"
   packages = ["pkg/kubelet/apis/stats/v1alpha1"]
-  pruneopts = "NUT"
   revision = "f01a2bf98249a4db383560443a59bed0c13575df"
   version = "v1.9.5"
 
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  input-imports = [
```
keitwb: There is some thrash from different versions of go dep.  Did you use the version in the dev image?

charliesignalfx: I did

PR #425, opened 2018-09-17
Add tail monitor

charliesignalfx:


internal/monitors/telegraf/monitors/tail/tail.go
```diff
@@ -0,0 +1,138 @@
...
+	var ctx context.Context
+	ctx, m.cancel = context.WithCancel(context.Background())
+
+	// start the tail plugin
+	if err = plugin.Start(ac); err != nil {
+		return err
+	}
+
+	// wait for the plugin to be stopped
+	<- ctx.Done()
```
keitwb: Configure isn't designed to block.

PR #430, opened 2018-09-19
Enable python plugins in windows

charliesignalfx:


internal/monitors/collectd/templating_notwindows.go
```diff
@@ -0,0 +1,150 @@
+// +build !windows
```
keitwb: What does this file reference that won't build on windows?

PR #433, opened 2018-09-26
Support K8S v1.11 integration tests with kubeadm

jchengsfx:


tests/conftest.py
```diff
@@ -164,23 +166,24 @@ def teardown():
...
     port = None
-    if worker_id in ("master", "gw0"):
+    k8s_container = request.config.getoption("--k8s-container")
+    if not k8s_container and worker_id in ("master", "gw0"):
         minikube.start_registry()
         port = minikube.registry_port
     print("\nWaiting for registry to be ready ...")
     assert wait_for(
-        p(container_is_running, minikube.client, "registry"), timeout_seconds=60
+        p(container_is_running, minikube.client, "registry"), 60, 2
```
keitwb: Can you use keyword args here and everywhere that passes both the timeout and interval?  It is much more readable that way even if it is a bit verbose.

PR #437, opened 2018-10-01
Install and use python 3.6.3 for circleci integration tests

jchengsfx:


.circleci/config.yml
```diff
@@ -169,7 +170,7 @@ jobs:
           export AGENT_BIN=$(pwd)/bundle/bin/signalfx-agent
           export BUNDLE_DIR=$(pwd)/bundle
           export TEST_SERVICES_DIR=$(pwd)/test-services
-          sudo -E pytest \
+          sudo -E $PYENV_ROOT/shims/pytest \
```
keitwb: Why do we need this??

jchengsfx: `$PYENV_ROOT/shims/` is auto added to PATH for the circleci user by pyenv, but sudo protects PATH.

.circleci/config.yml
```diff
@@ -250,6 +252,19 @@ jobs:
       - store_artifacts:
           path: ~/testresults/k8s_integration_tests.html
 
+  int_test_lint:
```
jchengsfx: @keitwb Did you want to add this job to the workflow?

keitwb: Yes

PR #442, opened 2018-10-02
clean up build directions so that go vet runs on darwin

charliesignalfx:


keitwb:
Please do rebase on latest master on here and let tests pass before merging

PR #445, opened 2018-10-03
Enable pylint and black

jchengsfx:


.circleci/config.yml
```diff
@@ -64,32 +64,32 @@ jobs:
...
-         export AGENT_VERSION=latest
-         make image
-     - run: |
-         mkdir -p /tmp/workspace
-         docker save -o /tmp/workspace/image.tar quay.io/signalfx/signalfx-agent-dev:latest
-     - persist_to_workspace:
-         root: /tmp/workspace
-         paths:
-           - image.tar
+      - checkout
```
keitwb: This doesn't line up anymore

tests/helpers/kubernetes/utils.py
```diff
@@ -62,45 +52,66 @@ def run_k8s_monitors_test(  # pylint: disable=too-many-locals,too-many-arguments
...
-    if expected_dims is not None:
-        if observer:
-            expected_dims = expected_dims.union(get_observer_dims_from_selfdescribe(observer))
-        expected_dims = expected_dims.union({"kubernetes_cluster"})
-
-    with run_k8s_with_agent(agent_image, minikube, monitors, observer, namespace, yamls, yamls_timeout) as [
-        backend,
-        agent,
-    ]:
+    args = dict(
```
keitwb: I don't like this at all.  If the whole point is to try and get around the too-many-arguments rule then just disable it and let it be.

.circleci/config.yml
jchengsfx: I indented everything with an extra space because yamllint was complaining: `wrong indentation: expected 6 but found 5  (indentation)`

tests/helpers/kubernetes/utils.py
jchengsfx: Ok, I'll revert.

PR #453, opened 2018-10-09
Openstack integration test

jchengsfx:


Makefile
```diff
@@ -175,3 +186,28 @@ lint-python:
 	pip install -q -r python/test-requirements.txt
 	LC_ALL=C.UTF-8 LANG=C.UTF-8 black --config python/pyproject.toml python/
 	cd python && pylint -j4 `find . -maxdepth 1 -mindepth 1 -type d -o -name "*.py"`
+
+.PHONY: devstack
+devstack:
```
keitwb: Too much code for a Makefile.  Put this in a script and have the Makefile simply call it.

Makefile
```diff
@@ -175,3 +186,28 @@ lint-python:
...
+	docker exec devstack start-devstack.sh
+	docker exec devstack su - stack -c 'cd /opt/stack/devstack && ./unstack.sh'
+	docker export devstack > /tmp/devstack.tar
+	docker import /tmp/devstack.tar quay.io/signalfx/devstack:latest
+	docker rm -fv devstack
+	rm -f /tmp/devstack.tar
+
+.PHONY: run-devstack
+run-devstack:
+	docker run -d --privileged \
```
keitwb: Same thing as above.

keitwb:
What is the benefit of running this in a separate test job in CircleCI?  This doesn't seem fundamentally any different than any other monitor check except that we have a custom test image in Docker, but that is pushed up to quay.io separate from the tests.

Makefile
```diff
@@ -103,10 +103,21 @@ run-dev-image:
...
+		-m "not packaging and not installer and not k8s and not openstack" \
 		-n 4 \
 		--verbose \
-		--html=test_output/results.html \
+		--html=test_output/integration_results.html \
+		--self-contained-html \
+		tests
+
+.PHONY: run-openstack-tests
+run-openstack-tests:
```
keitwb: Get rid of this.

Makefile
keitwb: I would actually just remove this make target and have it in a script if it ever needs to be run.

jchengsfx:
> What is the benefit of running this in a separate test job in CircleCI? This doesn't seem fundamentally any different than any other monitor check except that we have a custom test image in Docker, but that is pushed up to quay.io separate from the tests.

The devstack services currently take about 6-7 minutes to start up on my mac.  I need to gather metrics for how this behaves and affects the current circleci workflow.

keitwb:
> > What is the benefit of running this in a separate test job in CircleCI? This doesn't seem fundamentally any different than any other monitor check except that we have a custom test image in Docker, but that is pushed up to quay.io separate from the tests.
> 
> The devstack services currently take about 6-7 minutes to start up on my mac. I need to gather metrics for how this behaves and affects the current circleci workflow.

Wow ok that is slow.  

I would rather do this by having a conditional within the existing integration test job that only runs this test if the code for the openstack integration changed (e.g. `changes-include-dir tests/monitors/openstack internal/monitors/collectd/openstack test-services/openstack`) that is used to determine whether the `not openstack` selector gets populated or not.  This should be an exceedingly small proportion of the integration test runs and thus won't incur the cost of an additional CircleCI executor.  In the rare cases where Openstack is updated the integration tests will just have to take a bit longer to run.

keitwb:
The other issue I have is how much duplication exists in the circle config file for running the integration tests, but if we remove it that won't be an issue.

jchengsfx:
> The other issue I have is how much duplication exists in the circle config file for running the integration tests, but if we remove it that won't be an issue.

Agreed.  I just wanted to run it as a separate job temporarily for debugging purposes.

Makefile
jchengsfx: > Too much code for a Makefile. Put this in a script and have the Makefile simply call it.

Moved to scripts/make-devstack-image and scripts/run-devstack-image.

PR #454, opened 2018-10-10
Add JMX troubleshooting docs

keitwb:


docs/monitors/collectd-genericjmx.md
```diff
@@ -26,6 +26,64 @@ monitors:
...
+  ...
+```
+
+This should work as long as the agent is allowed to access port 5000 on the
+Java app's host (i.e. there is no firewall blocking it).  Note that this
+does not enable authentication or encryption, but these can be added if
+desired.
+
+Assuming you have the `host` config set to `172.17.0.3` and the port set to
+`5000` (this is a totally arbitrary port and your JMX app will probably just
```
charliesignalfx: "probably just something different" - grammar needs to be adjusted

PR #459, opened 2018-10-11
use alternate repo for golint in appveyor lint job

charliesignalfx:
golint made changes that require us to use their mirror repo


keitwb:
Can you elaborate?

PR #461, opened 2018-10-11
disable ipv6 in host observer until support in all applicable observers

charliesignalfx:


internal/observers/host/host.go
```diff
@@ -85,6 +85,10 @@ func (o *Observer) discover() []services.Endpoint {
 
 	var endpoints []services.Endpoint
 	for _, c := range conns {
+		// TODO: Add support for ipv6 to all observers
+		if c.Family == syscall.AF_INET6 {
+			continue
+		}
 		isIPSocket := c.Family == syscall.AF_INET || c.Family == syscall.AF_INET6
```
keitwb: You just need to remove the right hand side of the OR here.

keitwb: and maybe rename the var to `isIPV4Socket`

PR #465, opened 2018-10-17
Adding config options for API server to support multi-realm

keitwb:


deployments/docker/agent.yaml
```diff
@@ -1,6 +1,7 @@
 # *Required* The access token for the org that you wish to send metrics to.
 signalFxAccessToken: ${SFX_ACCESS_TOKEN}
 ingestUrl: ${SFX_INGEST_URL}
+apiUrl: ${SFX_API_URL}
```
jchengsfx: deployments/docker/README.md should be updated for this new env var.

deployments/installer/install.sh
```diff
@@ -26,10 +27,14 @@ parse_args_and_install() {
       --test)
         stage="test"
         ;;
-      --ingest)
+      --ingest-url)
```
jchengsfx: Should this be noted as a breaking change?

jchengsfx:
Should deployments/ecs/agent.yaml also be updated for apiUrl?

deployments/docker/agent.yaml
keitwb: Added.

keitwb:
> Should deployments/ecs/agent.yaml also be updated for apiUrl?

Yes, added.

deployments/installer/install.sh
keitwb: No, because nobody should be using the installer script in an automated way.  That is an abuse of the installer script.

I updated the usage output of the installer.sh script so that will tell people what the flag is when they go to use the script after it is updated.

PR #468, opened 2018-10-23
Conviva monitor

bjsignalfx:


internal/monitors/conviva/configsetter.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"context"
+	"strconv"
+	"strings"
+)
+
+func setConfigFields(m *Monitor, conf *Config)  {
+	conf.metriclensMetricNames = map[string][]string{
+		"quality_metriclens": {
+			"total_attempts",
```
keitwb: Some of these sound like counters, not gauges.  Why are they all treated as gauges?

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/signalfx/golib/datapoint"
+	"github.com/signalfx/golib/sfxclient"
+	"github.com/signalfx/signalfx-agent/internal/core/common/dpmeta"
+	"github.com/signalfx/signalfx-agent/internal/core/config"
+	"github.com/signalfx/signalfx-agent/internal/monitors"
+	"github.com/signalfx/signalfx-agent/internal/monitors/types"
+	"github.com/signalfx/signalfx-agent/internal/utils"
+	Logger "github.com/sirupsen/logrus"
```
keitwb: `log` is the conventional alias in the rest of the agent code.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+// `real-time(live)` video playing experience metrics from Conviva.
+//
+// Only `live` metrics types are supported. See
+// https://community.conviva.com/site/global/apis_data/experience_insights_api/index.gsp#metrics
+// if Conviva developer community member. The live metrics are gauge type metrics. They are converted
+// to SignalFx gauges with dimensions for the metric name, account name and filter name. For metriclens,
+// the metriclens dimensions are converted to SignalFx dimensions in addition. The values for these
+// dimensions are derived from the values of the associated metriclens dimension entities.
+//
+// TODO: doc about the default behavior
```
keitwb: Don't merge until this is fleshed out.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+		Transport: &http.Transport{
+			MaxIdleConnsPerHost: 100,
+		},
+	}
+	setConfigFields(m, conf)
+	maxNumOfGoroutines := len(conf.MetricConfigs) * 100
+	if maxNumOfGoroutines > 500 {
+		maxNumOfGoroutines = 500
+	}
+	maxNumOfGoroutinesChan := make(chan struct{}, maxNumOfGoroutines)
```
keitwb: Name this something that makes it clear it is being used as a semaphore.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+		for _, metricConfig := range conf.MetricConfigs {
+			if strings.Contains(metricConfig.Metric, "metriclens") {
+				getSendMetriclensMetrics(maxNumOfGoroutinesChan, m, metricConfig, conf)
+			} else {
+				getSendMetrics(maxNumOfGoroutinesChan, m, metricConfig, conf)
+			}
+		}
+	}, time.Duration(conf.IntervalSeconds) * time.Second)
+	return nil
+}
```
keitwb: Blank line after this line.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	defer res.Body.Close()
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	var payload map[string]interface{}
+	err = json.Unmarshal(body, &payload)
+	if res.StatusCode != 200 {
+		return nil, fmt.Errorf("Response status code: %d. Reason: %s.", res.StatusCode, payload["reason"])
```
keitwb: This is not safe to assume that payload will be non-nil when making this error message.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+	res, err := m.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	var payload map[string]interface{}
```
keitwb: It is much better to pass in an object of a struct type that conforms to the json structure you expect instead of deserializing to `interface{}` objects and doing type assertions.

See the way the kubelet monitor does it in `internal/monitors/cadvisor/kubeletstats.go:158`

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+//TODO: implement functionality for simple series and label series
+func getSendMetrics(maxNumOfGoroutinesChan chan struct{}, m *Monitor, metricConfig *MetricConfig, conf *Config) {
+	ctx, cancel := context.WithTimeout(m.ctx, time.Duration(conf.IntervalSeconds) * time.Second)
+	select {
+	case <- ctx.Done():
+		cancel()
+		logger.Error(ctx.Err())
+		return
+	case maxNumOfGoroutinesChan <- struct{}{}:
+		url := "https://api.conviva.com/insights/2.4/metrics.json?metrics=" + metricConfig.Metric +
```
keitwb: What is the difference between this and the call in the other method below?

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+			res, err := get(m, conf, url)
+			if err != nil {
+				logger.Error(err)
+				return
+			}
+			timestamps := res[metricConfig.Metric].(map[string]interface{})["timestamps"].([]interface{})
+			timeSeriesByFilterId := res[metricConfig.Metric].(map[string]interface{})["filters"].(map[string]interface{})
+			dps := make([]*datapoint.Datapoint, 0)
+			for filterId, timeSeries := range timeSeriesByFilterId {
+				for i, metricValue := range timeSeries.([]interface{}) {
```
keitwb: We probably don't want historical data, only the latest metrics when they are queried.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,254 @@
...
+//      filters:
+//        - All Traffic
+// ```
+
+var logger = Logger.WithFields(Logger.Fields{"monitorType": monitorType})
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig
+	Username                    string                       `yaml:"pulse_username" validate:"required"`
```
keitwb: Use camelCase config key names, not snake_case.

internal/monitors/conviva/configsetter.go
```diff
@@ -0,0 +1,153 @@
...
+	setAccounts(m, conf)
+	setFilters(m, conf)
+	setDimensions(m, conf)
+	setMetriclensFilters(m, conf)
+}
+
+func setAccounts(m *Monitor, conf *Config)  {
+	jsonResponse, err := get(m, conf, "https://api.conviva.com/insights/2.4/accounts.json")
+	if err != nil {
+		logger.Errorf("Get accounts request failed %v\n", err)
```
keitwb: Propagate the errors back to the calling function all the way back to the main Configure method on the monitor so that errors are clearer.

Also, for robustness, I would make it do this configuration on the first interval before it collects metrics in case the conviva API server is down temporarily when the agent starts up.  Have a check for whether it has been dynamically configured, and if not, have it do it before fetching metric values.  That way, if the Conviva server is down for an extended period, this monitor will start working after it comes back up automatically without user intervention.

keitwb:
You will need to run the code through golint to make it pass CI. Also check the vet errors.  You also need to run `make docs` in the dev image and commit the changes in this PR to update the rendered docs.

internal/monitors/conviva/configsetter.go
bjsignalfx: You would think that but looking at the total_attempts plot in pulse (conviva's web portal), total_attempts is not monotonically increasing!

internal/monitors/conviva/conviva.go
bjsignalfx: It the difference between iterating through this:
{
    "avg_bitrate": {
        "timestamps": [],
        "meta": {
            "status": 0
        },
        "type": "time_series",
        "filters": {
            "19457": []
        }
    }
}

As opposed to this:
{
    "quality_metriclens": {
        "tables": {
            "19457": {
                "rows": [
                    [],
		    …
                    ]
                ],
                "total_row": []
            }
        },
        "meta": {
            "status": 0
        },
        "type": "table",
        "xvalues": [ ]
    }
}

I will look into refactor the code in the future and make it more DRY compliant. Any suggestions?

internal/monitors/conviva/metricconfig.go
```diff
@@ -0,0 +1,111 @@
...
+	// Metriclens dimension names.
+	MetriclensDimensions   []string `yaml:"metriclensDimensions"`
+	accountID              string
+	filterMap              map[string]string
+	metriclensDimensionMap map[string]float64
+	isInitialized          bool
+}
+
+func (m *MetricConfig) init(accountService *AccountService) {
+	if !m.isInitialized {
```
keitwb: Wrap this function in a mutex since it could possibly take longer than the metric interval to complete and should never be done twice concurrently.

internal/monitors/conviva/accountservice.go
```diff
@@ -0,0 +1,223 @@
...
+	accounts       []*Account
+	ctx            context.Context
+	timeout        *time.Duration
+	httpClient     *HTTPClient
+}
+
+// NewAccountService factory function creating AccountService
+func NewAccountService(ctx context.Context, timeout *time.Duration, httpClient *HTTPClient) AccountService {
+	service := accountServiceImpl{ctx: ctx, timeout: timeout, httpClient: httpClient,}
+	service.init()
```
keitwb: Do the init with all of the HTTP requests on the initial read callback in case Conviva is having an outage when somebody starts up the agent.  Don't assume this will always be a one time thing that will always succeed.

internal/monitors/conviva/accountservice.go
```diff
@@ -0,0 +1,225 @@
...
+	"fmt"
+	"sync"
+	"time"
+)
+
+//const (
+//	metricLensURLFormat = "https://api.conviva.com/insights/2.4/metrics.json?metrics=%s&account=%s&filter_ids=%s&metriclens_dimension_id=%d"
+//)
+
+var anyErrors bool
```
keitwb: No global variables.  This monitor could conceivably be initialized twice.

internal/monitors/conviva/accountservice.go
```diff
@@ -0,0 +1,225 @@
...
+	getID(accountName string)                                                string
+	getFilters(accountName string)                                           map[string]string
+	getMetricLensFilters(accountName string)                                 map[string]string
+	getFilterID(accountName string, filterName string)                       string
+	getMetricLensDimensionID(accountName string, metricLensDimension string) float64
+}
+
+type accountServiceImpl struct {
+	defaultAccount *account
+	accounts       []*account
```
keitwb: Why don't you just make `accounts` a `map[string]*account` and make the key the account name so that you don't have to do all of those loops in the get* methods?

internal/monitors/conviva/accountservice.go
```diff
@@ -0,0 +1,225 @@
...
+	for _, a := range s.accounts {
+		a.metricLensDimensions = map[string]float64{}
+		waitGroup.Add(1)
+		go func(a *account) {
+			ctx, cancel := context.WithTimeout(s.ctx, *s.timeout)
+			defer waitGroup.Done()
+			defer cancel()
+			if err := (*s.client).Get(ctx, &a.metricLensDimensions, "https://api.conviva.com/insights/2.4/metriclens_dimension_list.json?account="+a.id); err != nil {
+				logger.Error(err)
+				mutex.Lock(); anyErrors = true; mutex.Unlock()
```
keitwb: no global vars.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,292 @@
...
+			MaxIdleConnsPerHost: 100,
+		},
+	}, conf.Username , conf.Password)
+
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+	semaphore := make(chan struct{}, maxGoroutinesPerInterval(conf.MetricConfigs))
+	interval := time.Duration(conf.IntervalSeconds) * time.Second
+	service := newAccountService(m.ctx, &m.timeout, &m.client)
+	utils.RunOnInterval(m.ctx, func() {
+		for _, metricConf := range conf.MetricConfigs {
```
keitwb: Run this all through `go fmt`, there are some formatting issues.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,292 @@
...
+					dps[i].Timestamp = now
+					dps[i].Meta[dpmeta.NotHostSpecificMeta] = true
+					m.Output.SendDatapoint(dps[i])
+				}
+			}(contextTimeout, m, conf, dim, fmt.Sprintf(metricLensURLFormat, conf.Metric, conf.accountID, strings.Join(conf.filterIDs(), ","), int(dimID)))
+		}
+	}
+}
+
+//func main(){
```
keitwb: Delete commented code.

internal/monitors/conviva/metricconfig.go
```diff
@@ -0,0 +1,111 @@
...
+	},
+	"audience_metriclens": {
+		"conviva.audience_metriclens.concurrent_plays",
+		"conviva.audience_metriclens.plays",
+		"conviva.audience_metriclens.ended_plays",
+	},
+}
+
+// metricConfig for configuring individual metric
+type metricConfig struct {
```
keitwb: If it is part of the user config struct then it should be exported.

internal/monitors/conviva/accountservice.go
```diff
@@ -0,0 +1,225 @@
+package conviva
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+)
+
+//const (
```
keitwb: No commented code.

internal/monitors/conviva/accountservice.go
```diff
@@ -0,0 +1,225 @@
...
+				if name == metricLensDimension {
+					return id
+				}
+			}
+		}
+	}
+	return 0
+}
+
+func (s *accountServiceImpl) init() {
```
keitwb: This function should return an error if it couldn't be initialized.

internal/monitors/conviva/accountservice.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// NewAccountService factory function creating AccountService
+func newAccountService(ctx context.Context, timeout *time.Duration, client *httpClient) accountService {
+	service := accountServiceImpl{ctx: ctx, timeout: timeout, client: client, mutex: &sync.RWMutex{},}
+	return &service
+}
+
+func (s *accountServiceImpl) getDefault() *account {
+	s.init()
```
keitwb: This should check init's return value for an error and return early if it errored out.  There should be logging somewhere that an error occurred.

internal/monitors/conviva/metricconfig.go
```diff
@@ -0,0 +1,111 @@
...
+	filterMap              map[string]string
+	metriclensDimensionMap map[string]float64
+	isInitialized          bool
+}
+
+func (m *metricConfig) init(service *accountService) {
+	if !m.isInitialized {
+		// setting account id and default account if necessary
+		if m.Account == "" {
+			m.Account = (*service).getDefault().Name
```
keitwb: The `(*service)` is unnecessary, you can just do `service.getDefault()`.

internal/monitors/conviva/metricconfig.go
bjsignalfx: This seems to work fine as long as the variables of that struct are exported.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,292 @@
...
+	metricLensURLFormat = "https://api.conviva.com/insights/2.4/metrics.json?metrics=%s&account=%s&filter_ids=%s&metriclens_dimension_id=%d"
+)
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig
+	// Conviva Pulse username required with each API request.
+	Username       string          `yaml:"pulseUsername" validate:"required"`
+	// Conviva Pulse password required with each API request.
+	Password       string          `yaml:"pulsePassword" validate:"required"`
```
jchengsfx: Need to add `neverLog:"true"` to mask the password in the agent output.

internal/monitors/conviva/accountsservice.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	return s.defaultAccount, nil
+}
+
+func (s *accountsServiceImpl) getMetricLensDimensionMap(accountName string) (map[string]float64, error) {
+	if err := s.init(); err != nil {
+		return nil, err
+	}
+	if a := s.accounts[accountName]; a != nil {
+			return a.metricLensDimensions, nil
```
keitwb: Run all your files through `go fmt`.  Something is off here.

internal/monitors/conviva/conviva.go
```diff
@@ -88,7 +88,9 @@ const monitorType = "conviva"
 var logger = log.WithFields(log.Fields{"monitorType": monitorType})
 
 const (
-	metricLensURLFormat = "https://api.conviva.com/insights/2.4/metrics.json?metrics=%s&account=%s&filter_ids=%s&metriclens_dimension_id=%d"
+	metricURLFormat     = "https://api.conviva.com/insights/2.4/metrics.json?metrics=%s&account=%s&filter_ids=%s"
+	metricLensURLFormat = metricURLFormat + "&metriclens_dimension_id=%d"
+		//metricLensURLFormat = "https://api.conviva.com/insights/2.4/metrics.json?metrics=%s&account=%s&filter_ids=%s&metriclens_dimension_id=%d"
```
keitwb: Remove commented code.

internal/monitors/conviva/metricconfig.go
```diff
@@ -0,0 +1,223 @@
...
+		"conviva.audience_metriclens.ended_plays",
+	},
+}
+
+// metricConfig for configuring individual metric
+type metricConfig struct {
+	// Conviva customer account name. The default account is fetched used if not specified.
+	Account                     string   `yaml:"account"`
+	Metric                      string   `yaml:"metric" default:"quality_metriclens"`
+	Filters                     []string `yaml:"filters"`                     // Filter names. The default is `All Traffic` filter
```
keitwb: Field comments go above the field definition, not to the right.

internal/monitors/conviva/metricconfig.go
```diff
@@ -0,0 +1,223 @@
...
+	MetricLensDimensions        []string `yaml:"metricLensDimensions"`        // MetricLens dimension names. The default is names of all MetricLens dimensions of the account
+	ExcludeMetricLensDimensions []string `yaml:"excludeMetricLensDimensions"` // MetricLens dimension names to exclude.
+	accountID                   string
+	filters                     map[string]string  // id:name map of filters derived from the configured Filters
+	metricLensDimensionMap      map[string]float64 // name:id map of MetricLens dimensions derived from configured MetricLensDimensions
+	isInitialized               bool
+	filtersWarmup               map[string]string // id:name map of filters in filters_warmup status on response
+	filtersNotExist             map[string]string // id:name map of filters in filters_not_exist status on response
+	filtersIncompleteData       map[string]string // id:name map of filters in filters_incomplete_data status on response
+	mutex                       *sync.RWMutex
```
keitwb: Make mutex a value not a pointer.

internal/monitors/conviva/accountsservice.go
```diff
@@ -0,0 +1,233 @@
...
+	getMetricLensDimensionID(accountName string, metricLensDimension string) (float64, error)
+}
+
+type accountsServiceImpl struct {
+	defaultAccount      *account
+	accounts            map[string]*account
+	ctx                 context.Context
+	timeout             *time.Duration
+	client              httpClient
+	mutex               *sync.RWMutex
```
keitwb: If you make mutex a value you don't need to initialize it in `newAccountsService`.  The default value for mutex is an unlocked mutex.

internal/monitors/conviva/conviva.go
```diff
@@ -0,0 +1,276 @@
...
+func init() {
+	monitors.Register(monitorType, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Configure monitor
+func (m *Monitor) Configure(conf *Config) error {
+	if conf.MetricConfigs == nil {
+		conf.MetricConfigs = []*metricConfig{{Metric: "quality_metriclens"}}
+	}
+	for _, metricConf := range conf.MetricConfigs {
```
keitwb: Get rid of this after you make mutex a value on the struct instead of a pointer.

internal/monitors/conviva/metricconfig.go
```diff
@@ -0,0 +1,223 @@
...
+			logger.Errorf("Metric %s MetricLens dimension(s) exclusion failure. %+v", mc.Metric, err)
+			return
+		}
+		mc.isInitialized = true
+	}
+}
+
+// setting account id and default account if necessary
+func (mc *metricConfig) setAccount(service accountsService) error {
+	if strings.TrimSpace(mc.Account) == "" {
```
keitwb: If you just do the trim on line 82 above this line then you can just check for `mc.Account == ""` instead of doing the trim twice.

keitwb:
@jchengsfx What is the status on the integration tests?

internal/monitors/conviva/metricconfig.go
```diff
@@ -0,0 +1,231 @@
...
+			if allFilters, err = service.getMetricLensFilters(mc.Account); err != nil {
+				return err
+			}
+		} else {
+			if allFilters, err = service.getFilters(mc.Account); err != nil {
+				return err
+			}
+		}
+		mc.Filters = make([]string, 0, len(allFilters))
+		mc.filters = make(map[string]string, len(allFilters))
```
keitwb: This is incredibly confusing have two fields that only differ by the initial letter casing.  Can you make the unexported one have a more descriptive name?

keitwb: Say, `filtersByID` or something

PR #475, opened 2018-10-26
Update es monitor

asuresh4:
Adding an optional cluster config option and updates to tests


internal/monitors/collectd/elasticsearch/elasticsearch.go
```diff
@@ -36,6 +36,8 @@ type Config struct {
 	Port                 uint16 `yaml:"port" validate:"required"`
 	// AdditionalMetrics to report on
 	AdditionalMetrics []string `yaml:"additionalMetrics"`
+	// Cluster name to which the node belongs
```
keitwb: I would make this a bit more descriptive and clarify that it is optional and can be any value they want and that it is used to populate the plugin_instance dimension.

asuresh4: Updated the description to be more verbose.

PR #477, opened 2018-10-31
Prop filter

mstumpfx:


internal/core/propfilters/filter_test.go
```diff
@@ -0,0 +1,461 @@
...
+func TestFilters(t *testing.T) {
+	t.Run("Filter a single property name", func(t *testing.T) {
+		f, _ := New([]string{"pod-template-hash"}, []string{"*"},
+			[]string{"*"}, []string{"*"})
+
+		properties := map[string]string{"pod-template-hash": "123", "replicaSet": "abc"}
+		filteredProperties := f.FilterProperties(properties)
+
+		spew.Dump(f)
+		assert.True(t, len(filteredProperties) == 1)
```
keitwb: Use the specific assertion functions for length and equals instead of just using assert.True for everything.

keitwb: Also, assert.Equal works with maps I think so that you can just have one assertion and it will cover both length and content.

internal/core/writer/properties.go
```diff
@@ -59,39 +59,45 @@ func newDimensionPropertyClient(conf *config.WriterConfig) *dimensionPropertyCli
...
 		// for the same dim props, but that's ok.
-		dpc.history.Add(dimProps.Dimension, dimProps)
+		dpc.history.Add(filteredDimProps.Dimension, filteredDimProps)
 		atomic.AddInt64(&dpc.TotalPropUpdates, int64(1))
 	}
 	return nil
 }
 
+// applyFilter filters out properties from filter sets and returns
+func applyFilter(dimProps *types.DimProperties, propFilters *propfilters.FilterSet) *types.DimProperties {
```
keitwb: Get rid of this function and just do the equivalent in line 74.  This is unnecessary redirection that just makes it harder to understand.

internal/utils/filter/matching.go
```diff
@@ -53,3 +53,13 @@ func anyRegexMatches(s string, res []*regexp.Regexp) bool {
 	}
 	return false
 }
+
+// stripNegation checks if a string is prefixed with "!"
+// and will returned the stripped string and true if so
+// else, return original value and false
+func stripNegation(value string) (string, bool) {
+	if strings.HasPrefix(value, "!") {
+		return value[1:], true
```
keitwb: What happens if somebody puts in a filter value of only `!`?

PR #481, opened 2018-11-08
Integration tests for conviva monitor

jchengsfx:


keitwb:
@bjsignalfx take a look and make sure the tests look right.  If so, go ahead and merge into your branch and then I will do a final review of your branch.

bjsignalfx:
It looks good except I changed the metricConfig field metric to metricParameter to be in line with Conviva documentation.

PR #485, opened 2018-11-09
Integration test updates

jchengsfx:
* Handle flakiness in some integration tests.
* Enable parallelism for integration tests in circleci.


scripts/lint-pytest
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+RC=0
+
+pip install -q -r tests/requirements.txt || exit 1
```
keitwb: I would not do the pip install here or in the script below.  This is mainly to make it quick to run these scripts locally before committing and not have it run through a pip install every time.  CircleCI and the dev image should install these things separately in a separate step and be cached.

tests/conftest.py
```diff
@@ -240,6 +240,7 @@ def teardown():
                 env={"PULL_CACHE": "yes", "AGENT_IMAGE_NAME": agent_image_name, "AGENT_VERSION": agent_image_tag},
                 stderr=subprocess.STDOUT,
                 check=True,
+                cwd=os.path.join(os.path.dirname(os.path.realpath(__file__)), ".."),
```
keitwb: Can you pull this out to a global constant called something like `REPO_ROOT_DIR` or something so that it is more explicit what it is for?

scripts/lint-pytest
jchengsfx: I added this in these scripts because we're using different pylint versions for the python files and the pytest files for python v.2.7 and v3.6 compatibility. So this just explicitly ensures that the correct version is installed when running pylint locally in the dev image.

.circleci/config.yml
```diff
@@ -32,7 +32,14 @@ commands:
     steps:
       - attach_workspace:
           at: /tmp/workspace
-      - run: docker load -i /tmp/workspace/image.tar
+      - run:
+          name: Load agent image
+          command: |
+            if [ -f ~/.skip ]; then
```
keitwb: For conciseness sake, I would make this `[ -f ~/.skip ] && echo "Found ~/.skip, skipping docker image load!" >&2 && exit 0` so that it takes up less vertical space and makes the docker load command more prominent.  It is a common enough idiom in shell scripts that it shouldn't be too cryptic.

scripts/lint-python
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+RC=0
+
+pip install -q -e python/ || exit 1
+
+LC_ALL=C.UTF-8 LANG=C.UTF-8 black --config python/pyproject.toml python/
+(( RC += $? ))
+
+bash -ec 'cd python && pylint -j4 `find . -maxdepth 1 -mindepth 1 -type d -o -name "*.py"`'
```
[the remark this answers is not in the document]
jchengsfx: It just saves an extra step of having to cd back to the cwd.  I can just use pushd/popd instead.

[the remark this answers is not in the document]
keitwb: Why would you need to cd back anyway?  It won't change the dir of whatever script invokes this since the whole thing runs in a separate subshell.

keitwb:
Lets try and get this in soon.  It has been lingering for really long now.

keitwb:
Is there anything outstanding to do with it?

jchengsfx:
> Is there anything outstanding to do with it?

I have a few more minor commits to push after I finish testing locally, but should be ready soon.

jchengsfx:
> > Is there anything outstanding to do with it?
> 
> I have a few more minor commits to push after I finish testing locally, but should be ready soon.

Also, I was waiting for https://github.com/signalfx/signalfx-agent/pull/530 to merge since I will need to resolve conflicts based on the root test dir changes.

keitwb:
I'm about to merge #530 soon, let try and get this in right after we cut the 4.0 release.

PR #486, opened 2018-11-12
Conviva monitor bugfix

bjsignalfx:


internal/monitors/conviva/conviva.go
```diff
@@ -169,33 +168,35 @@ func (m *Monitor) fetchMetrics(contextTimeout time.Duration, semaphore chan stru
...
 						case <-ctx.Done():
 							logger.Error(ctx.Err())
 							return
 						default:
-							dp := sfxclient.GaugeF(
-								metricName,
-								map[string]string{"account": metricConf.Account, "filter": metricConf.filtersMap[filterID]},
-								metricValue)
-							// Checking the type of series and setting dimensions accordingly
+							dp := sfxclient.GaugeF(metricName, map[string]string{"account": metricConf.Account, "filter": metricConf.filtersMap[filterID]}, metricValue)
```
keitwb: Make all of this that creates the datapoint into a separate function.  This function is doing too much.

internal/monitors/conviva/conviva.go
```diff
@@ -169,33 +168,35 @@ func (m *Monitor) fetchMetrics(contextTimeout time.Duration, semaphore chan stru
...
+			timestamp := time.Now()
 			dps := make([]*datapoint.Datapoint, 0)
 			for metricParameter, series := range res {
 				metricConf.logFilterStatuses(series.Meta.FiltersWarmup, series.Meta.FiltersNotExist, series.Meta.FiltersIncompleteData)
 				metricName := "conviva." + metricParameter
 				for filterID, metricValues := range series.FilterIDValuesMap {
+				L:
 					for i, metricValue := range metricValues {
 						select {
 						case <-ctx.Done():
```
keitwb: This is really unnecessary since the stuff in the default case is non-blocking.  Just let it make the datapoints and don't worry about trying to catch shutdown with such fine precision.  The Done channel is mainly for stopping goroutines.

internal/monitors/conviva/conviva.go
```diff
@@ -169,33 +168,35 @@ func (m *Monitor) fetchMetrics(contextTimeout time.Duration, semaphore chan stru
...
+							dp := sfxclient.GaugeF(metricName, map[string]string{"account": metricConf.Account, "filter": metricConf.filtersMap[filterID]}, metricValue)
+							dps = append(dps, dp)
+							dp.Meta[dpmeta.NotHostSpecificMeta] = true
 							switch series.Type {
+							// Getting the latest time_series metric value and timestamp only
 							case "time_series":
-								dp.Timestamp = time.Unix(int64(0.001*series.Timestamps[i]), 0)
+								dp.Value = datapoint.NewFloatValue(metricValues[len(metricValues)-1])
+								// Series timestamps are in milliseconds
+								dp.Timestamp = time.Unix(series.Timestamps[len(series.Timestamps)-1]/1000, 0)
```
keitwb: Is there really no way to just request the latest value only? Check the docs closely.

internal/monitors/conviva/conviva.go
```diff
@@ -169,33 +168,35 @@ func (m *Monitor) fetchMetrics(contextTimeout time.Duration, semaphore chan stru
...
-								metricValue)
-							// Checking the type of series and setting dimensions accordingly
+							dp := sfxclient.GaugeF(metricName, map[string]string{"account": metricConf.Account, "filter": metricConf.filtersMap[filterID]}, metricValue)
+							dps = append(dps, dp)
+							dp.Meta[dpmeta.NotHostSpecificMeta] = true
 							switch series.Type {
+							// Getting the latest time_series metric value and timestamp only
 							case "time_series":
-								dp.Timestamp = time.Unix(int64(0.001*series.Timestamps[i]), 0)
+								dp.Value = datapoint.NewFloatValue(metricValues[len(metricValues)-1])
```
keitwb: This is very confusing.  What is the original value that you were setting on the dp (`metricValue`) and why it is now being overridden?

bjsignalfx: In the case of time series I'm only interested in the latest metric value.

PR #488, opened 2018-11-13
Make access token optional

keitwb:
It isn't always required in situations where you are sending to a network-local proxy


mpetazzoni:
@keitwb Did you update the docs accordingly?

PR #492, opened 2018-11-13
Fix array config type in ES monitor

asuresh4:


keitwb:
Please squash when merging

PR #501, opened 2018-11-26
Enable pytest caching

jchengsfx:


.circleci/config.yml
```diff
@@ -33,14 +33,12 @@ commands:
...
+          key: v1-pytest-{{ checksum "tests/requirements.txt" }}
       - run:
           name: Install pytest and dependencies
           command: |
-            if [ -f ~/.skip ]; then
-                echo "Found ~/.skip, skipping pytest install!"
-                exit 0
-            fi
-            pyenv install 3.6.3
+            pyenv install -s 3.6.3
```
keitwb: Use full flags in scripts so it is clearer.  See https://superuser.com/a/174568

PR #511, opened 2018-12-07
Allow restricting the kubernetes_cluster monitor and k8s-api observer…

keitwb:
… to a single namespace

### DIFF
--- a/docs/monitors/kubernetes-cluster.md
+++ b/docs/monitors/kubernetes-cluster.md
@@ -37,6 +37,7 @@ Monitor Type: `kubernetes-cluster`
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
 | `alwaysClusterReporter` | no | `bool` | If `true`, leader election is skipped and metrics are always reported. (**default:** `false`) |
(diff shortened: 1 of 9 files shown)


docs/monitors/kubernetes-cluster.md
```diff
@@ -37,6 +37,7 @@ Monitor Type: `kubernetes-cluster`
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
 | `alwaysClusterReporter` | no | `bool` | If `true`, leader election is skipped and metrics are always reported. (**default:** `false`) |
+| `namespace` | no | `string` | If specified, only resources within the given namespace will be monitored.  If omitted (blank) all supported resources across all namespaces will be monitored. |
```
rmfitzpatrick: Is there an expectation of supporting multiple namespaces or are wildcards possible?

keitwb: No, because the K8s client-go library doesn't support multiple namespaces so the simplest thing to do is just to have a monitor do all or one and then instantiate multiple monitor instances if you want to monitor multiple namespaces but not all.

PR #512, opened 2018-12-12
Update metricproxy to latest

keitwb:
This will get us the updates to the trace endpoint logic around Jaeger process tags.


keitwb:
We should seriously consider refactoring metricproxy/gateway code to be more reusable without pulling in such a ridiculous amount of dependencies.

PR #513, opened 2018-12-12
Use pre-built images for hadoop tests

jchengsfx:
Reduces time and flakiness from building the hadoop test image.


jchengsfx:
> Add some doc or make target to make and push this test image.

Added the docker build command to the docstring in the test.

PR #519, opened 2018-12-19
Update go to v1.11.4

jchengsfx:


Dockerfile
```diff
@@ -307,6 +309,11 @@ COPY whitelist.json /lib/whitelist.json
 WORKDIR /
 
 
+# Workaround to utilize the global GO_VERSION argument
+# since "COPY --from" doesn't support variables.
+FROM golang:${GO_VERSION}-stretch as golang-ignore
```
keitwb: ok, this is pretty hacky but it seems like the best way without resorting to templating the Dockerfile.

PR #526, opened 2018-12-31
Fix regression in Redis key length config

### LINKED ISSUE #521: Redis ListLength expects DBIndex/KeyPattern dict but redis_info.py expects tuple
The `ListLength` struct (https://github.com/signalfx/signalfx-agent/blob/master/internal/monitors/collectd/redis/redis.go#L59) has keys `DBIndex` and `KeyPattern`.

However, `redis_info.py` (https://github.com/signalfx/redis-collectd-plugin/blob/master/redis_info.py#L245) expects a 2-item list (`(len(node.values)) == 2`). This causes the integration to break - you get a warning "redis_info plugin: monitoring length of keys requires both database index and key value".

If I had to guess, the fix could be to modify https://github.com/signalfx/signalfx-agent/blob/master/internal/monitors/collectd/redis/redis.go#L115 to just convert `conf.SendListLengths` into this tuple form.

keitwb:
It didn't get converted properly when we switched to the custom Python runner.

Fixes #521

### DIFF
--- a/internal/monitors/collectd/redis/redis.go
+++ b/internal/monitors/collectd/redis/redis.go
@@ -86,6 +86,14 @@ func (c *Config) PythonConfig() *python.Config {
 	return c.pyConf
 }
 
+func (c *Config) sendListLengthsTuples() [][]interface{} {
+	var out [][]interface{}
+	for _, ll := range c.SendListLengths {
+		out = append(out, []interface{}{ll.DBIndex, ll.KeyPattern})
+	}
+	return out
+}
+
 // Monitor is the main type that represents the monitor
 type Monitor struct {
 	python.PyMonitor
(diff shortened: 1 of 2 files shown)


internal/monitors/collectd/redis/redis.go
```diff
@@ -86,6 +86,14 @@ func (c *Config) PythonConfig() *python.Config {
 	return c.pyConf
 }
 
+func (c *Config) sendListLengthsTuples() [][]interface{} {
+	var out [][]interface{}
+	for _, ll := range c.SendListLengths {
+		out = append(out, []interface{}{ll.DBIndex, ll.KeyPattern})
```
bjsignalfx: Remind me how yaml list elements databaseIndex and keyPattern get mapped to struct fields DBIndex and KeyPattern.

keitwb: Through the struct tags `yaml:"..."` on the config struct.

bjsignalfx: SendListLengths []ListLength `yaml:"sendListLengths"`?!

keitwb: Yes, that is it.

bjsignalfx: Ok, I'm sure it works. I just don't see how databaseIndex gets mapped to DBIndex unless it by position. We'll have chat about it when we meet.

PR #527, opened 2019-01-03
added solr collectd plugin

bjsignalfx:


internal/monitors/collectd/solr/solr.go
```diff
@@ -0,0 +1,78 @@
...
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"true"`
+	pyConf               *python.Config
+	Host                 string `yaml:"host" validate:"required"`
+	Port                 uint16 `yaml:"port" validate:"required"`
+	// Cluster name of this solr cluster.
+	Cluster string `yaml:"cluster"`
+	// EnhancedMetrics boolean to indicate whether stats from /metrics are needed
+	EnhancedMetrics *bool `yaml:"enhancedMetrics" default:"false"`
+	// IncludeMetric metric name from the /admin/metrics endpoint to include (valid when EnhancedMetrics is "false")
+	IncludeMetric string `yaml:"includeMetric"`
```
keitwb: This will only allow for one metric to include since yaml doesn't support repeated keys like the collectd config format does.  You need to make this a `[]string` and then make the value passed to collectd via the `PluginConfig` map be this:

```
"IncludeMetric": {
  "#flatten": true,
  "values": conf.IncludeMetric,
}
```

internal/monitors/collectd/solr/solr.go
```diff
@@ -0,0 +1,78 @@
...
+	// ExcludeMetric metric name from the /admin/metrics endpoint to exclude (valid when EnhancedMetrics is "true")
+	ExcludeMetric string `yaml:"excludeMetric"`
+	// Dimension space-separated key-value pair for a user-defined dimension
+	Dimension string `yaml:"dimension"`
```
keitwb: This is unnecessary since the agent has the ability to add dimensions via the `extraDimensions` key that is inherent to every monitor.  Remove this and don't pass it to the plugin.

internal/monitors/collectd/solr/solr.go
bjsignalfx: If I just change conf.IncludeMetric to []string as you suggest, why won't "IncludeMetric": conf.IncludeMetric work?

keitwb: Because the collectd plugin doesn't support arrays/lists for the IncludeMetric config option.  It expects a bunch of separate instances of IncludeMetric on separate lines.

bjsignalfx: Thanks nvm. I'm looking at what you did for redis and I understand.

internal/monitors/collectd/solr/solr.go
bjsignalfx: Ok

internal/monitors/collectd/solr/solr.go
```diff
@@ -0,0 +1,106 @@
...
+// Sample YAML configuration:
+//
+// ```yaml
+// monitors:
+// - type: collectd/solr
+//   host: 127.0.0.1
+//   port: 8983
+// ```
+//
+// Sample YAML configuration with list lengths:
```
keitwb: "list lengths"?? Did you copy/paste this from the Redis monitor?

internal/monitors/collectd/solr/solr.go
```diff
@@ -0,0 +1,106 @@
...
+		ModulePaths:   []string{collectd.MakePath("solr")},
+		TypesDBPaths:  []string{collectd.MakePath("types.db")},
+		PluginConfig: map[string]interface{}{
+			"Host":            conf.Host,
+			"Port":            conf.Port,
+			"Cluster":         conf.Cluster,
+			"EnhancedMetrics": conf.EnhancedMetrics,
+			"IncludeMetric": map[string]interface{}{
+				"#flatten": true,
+				"values":   [][]string{conf.IncludeMetrics},
```
keitwb: There shouldn't be two levels of slicing here or below.

internal/monitors/collectd/solr/solr.go
```diff
@@ -0,0 +1,106 @@
...
+//
+// ```yaml
+// monitors:
+// - type: collectd/solr
+//   host: 127.0.0.1
+//   port: 8983
+//   includeMetrics:
+//   - solr.core_deleted_docs
+//     solr.core_index_size
+//   excludeMetrics:
```
keitwb: The docs say that `excludeMetrics` is valid when `enhancedMetrics` is `true`, but it is not true here, so this is confusing.

scripts/docs/metric-doc-to-agent
```diff
@@ -22,8 +22,8 @@ for f in ${INTEGRATIONS_REPO%/}/$INTEGRATION/docs/*.md; do
     continue
   fi
   metric_name=$(basename $f .md)
-  desc=$(head -n4 $f | yq r - 'brief')
-  type=$(head -n4 $f | yq r - 'metric_type')
+  desc=$(head -n4 $f | yq -r ."brief")
+  type=$(head -n4 $f | yq -r ."metric_type")
```
jchengsfx: The yq command syntax is dependent on whether you have yq or python-yq installed on your system. Maybe we should try to handle both, or only allow one or the other?

tests/monitors/solr/solr_test.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_solr_monitor():
+    with run_service("solr") as solr_container:
+        host = container_ip(solr_container)
+        config = MONITOR_CONFIG.substitute(host=host)
+        assert wait_for(p(tcp_socket_open, host, 8983), 60), "service not listening on port"
+
+        with run_agent(config) as [backend, get_output, _]:
+            assert wait_for(p(has_datapoint_with_dim, backend, "plugin", "solr")), "Didn't get solr datapoints"
```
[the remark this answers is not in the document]
jchengsfx: You can also add a check for `excludeMetrics`, for example:
`assert ensure_always(lambda: not has_datapoint_with_metric_name(backend, "solr.core_max_docs"))`

internal/monitors/collectd/solr/solr.go
bjsignalfx: Yes

tests/monitors/solr/solr_test.py
```diff
@@ -3,33 +3,27 @@
...
-    solr.core_index_size
-  excludeMetrics:
-  - solr.core_max_docs
-    solr.core_num_docs
-"""
-)
-
 
 def test_solr_monitor():
+    monitor_config = string.Template(
```
keitwb: If you are going to put config in the test function itself, just use an f-string and textwrap.dedent like some of the other tests do.

scripts/docs/metric-doc-to-agent
keitwb: Yeah, I would rather not change this, but would rather update the dev image to install the right version of yq that works with this.

jchengsfx: > Yeah, I would rather not change this, but would rather update the dev image to install the right version of yq that works with this.

This script seems to be intended to run directly on the local system (outside of the dev image) since it also requires the integrations repo.

bjsignalfx: Ok, reverted scripts/docs/metric-doc-to-agent

PR #530, opened 2019-01-04
Add ability to map endpoint fields to config dynamically

keitwb:


keitwb:
@ksatrasala This is the thing you needed.  Take a look and see if it is powerful enough for what you need.

docs/auto-discovery.md
```diff
@@ -68,6 +71,30 @@ There are no implicit rules built into the agent, so each rule must be specified
 manually in the config file, in conjunction with the monitor that should monitor the
 discovered service.
 
+## Endpoint Config Mapping
+
+Sometimes it might be useful to use certain live attribute of a discovered
```
rmfitzpatrick: certain live attribute?

internal/core/config/monitor.go
```diff
@@ -26,6 +26,11 @@ type MonitorConfig struct {
 	// monitor(s) created from this configuration. To specify metrics from this
 	// monitor should be high-resolution, add the dimension `sf_hires: 1`
 	ExtraDimensions map[string]string `yaml:"extraDimensions" json:"extraDimensions"`
+	// A set of mappings from a configuration option on this montitor to
```
rmfitzpatrick: montitor

internal/core/config/monitor.go
```diff
@@ -26,6 +26,11 @@ type MonitorConfig struct {
...
+	// attributes of a discovered endpoint.  They keys are the config option on
```
rmfitzpatrick: they keys

tests/basic/config_mappings.py
```diff
@@ -0,0 +1,29 @@
...
+
+CONFIG = """
+observers:
+  - type: docker
+monitors:
+  - type: collectd/kafka
+    discoveryRule: container_name == "kafka-discovery" && port == 7099
+    configEndpointMappings:
+      extraDimensions: 'Get(container_labels, "com.signalfx.extraDimensions")'
+      clusterName: 'Get(container_labels, "com.signalfx.cluster")'
```
rmfitzpatrick: any harm in testing contains?

keitwb: Nope, but it isn't really relevant to the config mapping.  I should probably have a test for boolean values though.

PR #531, opened 2019-01-07
Add logging and better 404 message if invalid trace forwarder path is…

keitwb:
… used

### DIFF
--- a/internal/monitors/traceforwarder/forwarder.go
+++ b/internal/monitors/traceforwarder/forwarder.go
@@ -2,6 +2,7 @@ package traceforwarder
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"time"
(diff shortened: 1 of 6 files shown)


internal/monitors/traceforwarder/forwarder.go
```diff
@@ -60,3 +63,9 @@ func setupHandler(ctx context.Context, router *mux.Router, chainType string, sin
 	pathSetup(router, handler, signalfx.DefaultTracePathV1)
 	return internalMetrics
 }
+
+func notFoundHandler(w http.ResponseWriter, r *http.Request) {
+	errMsg := fmt.Sprintf("Trace span request received on invalid path '%s'. You should send to the same path that you would on the Smart Gateway.", r.URL.Path)
+	logger.Error(errMsg)
```
mpetazzoni: Is this a throttled logger? I can see quickly overloading the log output of the local host's agent if applications are misconfigured.

keitwb: Good point, let me see how to best handle that.

PR #532, opened 2019-01-08
Update SignalFx gateway and golib to latest

keitwb:
This gets the fix for the timestamp/duration parsing in traces


keitwb:
Failure is Kong test due to image update, not anything code related in this change.

PR #534, opened 2019-01-09
Add vagrant files and resources for windows dev environment

charliesignalfx:


charliesignalfx:
I am merging this because the test failures are due to an unrelated and known issue.  There is already a fix for it in master.  I will rebase charlie-dev following this merge and everything should be good.

PR #537, opened 2019-01-11
Make internal-metrics monitor accept endpoints

keitwb:


mpetazzoni:
@keitwb anything to look at here on the build failures? That seemed like a super trivial change.

PR #539, opened 2019-01-17
system utilization monitor

charliesignalfx:


docs/monitors/telegraf-win_perf_counters.md
```diff
@@ -48,13 +48,13 @@ The **nested** `objects` config object has the following fields:
...
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
-| `objectName` | no | `string` |  |
-| `counters` | no | `list of string` |  |
-| `instances` | no | `list of string` |  |
-| `measurement` | no | `string` |  |
-| `warnOnMissing` | no | `bool` |  (**default:** `false`) |
-| `failOnMissing` | no | `bool` |  (**default:** `false`) |
-| `includeTotal` | no | `bool` |  (**default:** `false`) |
+| `objectName` | no | `string` | the name of a windows performance counter object |
```
keitwb: Titlecase these descriptions.

internal/monitors/telegraf/monitors/winperfcounters/win_perf_counters_windows.go
```diff
@@ -17,16 +17,18 @@ import (
...
+func GetPlugin(conf *Config) *telegrafPlugin.Win_PerfCounters {
 	plugin := factory().(*telegrafPlugin.Win_PerfCounters)
 
 	// copy top level struct fields
 	deepcopier.Copy(conf).To(plugin)
 
 	// Telegraf has a struct wrapper around time.Duration, but it's defined
 	// in an internal package which the gocomplier won't compile from
-	plugin.CountersRefreshInterval.Duration = time.Duration(conf.CountersRefreshInterval) * time.Second
+	plugin.CountersRefreshInterval.Duration = time.Duration(conf.CountersRefreshInterval) * time.Millisecond
```
keitwb: The description in the config struct says "number of nanoseconds".  Can you encode the unit in the config option name (e.g. `CountersRefreshIntervalMillis`) or use `time.Duration` which will be parsed as a string (see https://github.com/signalfx/signalfx-agent/blob/master/internal/monitors/traceforwarder/monitor.go#L41 for an example) so that the user can specify the units themselves.

internal/monitors/utilization/perfcounter/logical_disk_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/signalfx/signalfx-agent/internal/neotest"
+)
+
+func TestLogicalDiskMeasurement(t *testing.T) {
+	var monitorType = "signalfx-system-utilization"
+	var Measurement = "win_logical_disk"
+	var percentFreeSpace = float32(80)
+	var utilization = float64(100 - percentFreeSpace)
+	var freeMegabytes = float32(1024)
+	var usedBytes = ((float64(freeMegabytes) * megabytesToBytes) * 100 / (100 - utilization)) - (float64(freeMegabytes) * megabytesToBytes)
```
keitwb: If you just make `freeMegabytes` a `float64` to begin with there is no need for the conversions.  There is no need to micro-optimize stack usage in a unit test.

internal/monitors/utilization/perfcounter/perfcounter.go
```diff
@@ -0,0 +1,322 @@
...
+type LogicalDiskImpl struct{ *BasePerfCounter }
+
+// ProcessMeasurement processes a given measurement and returns logical disk
+func (l *LogicalDiskImpl) ProcessMeasurement(ms *measurement.Measurement, monitorType string, sendDatapoint func(dp *datapoint.Datapoint)) (errors []error) {
+	if len(ms.Fields) == 0 {
+		errors = append(errors, fmt.Errorf("no fields on logical disk measurement '%s'", ms.Measurement))
+		return
+	}
+
+	var dimensions map[string]string
```
keitwb: This declaration is unnecessary.

internal/monitors/utilization/perfcounter/perfcounter.go
```diff
@@ -0,0 +1,322 @@
...
+		includeTotal:         true,
+		warnOnMissing:        true,
+		objectName:           "PhysicalDisk",
+		instanceDimensionKey: "plugin_instance",
+		metricType:           datapoint.Gauge,
+		counters:             []string{"Disk Reads/sec", "Disk Writes/sec"},
+		instances:            []string{"_Total"},
+		metricNameMapping: map[string]string{
+			// perfcounter: "Disk Reads/sec"; perfcounter reporter: ""; collectd: "disk_ops.read"
+			"Disk_Reads_persec": "disk.read.per_second",
```
keitwb: This is ambiguous.  Does this mean _read operations_, _read bytes_, _read sectors_, etc?

internal/monitors/utilization/utilization.go
```diff
@@ -0,0 +1,45 @@
...
+// ```
+
+var logger = log.WithFields(log.Fields{"monitorType": monitorType})
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `acceptsEndpoints:"false"`
+	// (Windows Only) number of seconds that wildcards in counter paths should
+	// be expanded and how often to refresh counters from configuration.
+	CountersRefreshInterval int `yaml:"counterRefreshInterval" default:"300000"`
```
keitwb: If the `300,000` is really seconds as the comments say, this would be about 3.5 days.  See comment above about this field in another config struct.

internal/monitors/utilization/utilization.go
```diff
@@ -0,0 +1,45 @@
...
+// ```yaml
+// monitors:
+//  - type: signalfx-system-utilization
+// ```
+
+var logger = log.WithFields(log.Fields{"monitorType": monitorType})
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `acceptsEndpoints:"false"`
```
keitwb: I would make this `singleInstance` (https://github.com/signalfx/signalfx-agent/blob/master/internal/monitors/metadata/hostmetadata/hostmetadata.go#L74) unless you know of a good use case for having multiple instances of this.

internal/monitors/utilization/perfcounter/logical_disk_test.go
charliesignalfx: It's float32 because that value is set in a Measurement further down and it's intended to match exactly what is passed into this code from the perf counter monitor.

internal/monitors/utilization/utilization_windows.go
```diff
@@ -0,0 +1,126 @@
...
+	memory := perfcounter.Memory()
+	processor := perfcounter.Processor()
+	logicalDisk := perfcounter.LogicalDisk()
+	physicalDisk := perfcounter.PhysicalDisk()
+	networkIntf := perfcounter.NetworkInterface()
+	pagefile := perfcounter.PageFile()
+	processMeasurements := func(measurements []*measurement.Measurement) {
+		for _, measurement := range measurements {
+			var err []error
+			switch measurement.Measurement {
```
keitwb: This is kind of a code smell having so much repetition of the same set of 6 components three times in this function, esp. this switch statement.  You could eliminate the switch statement with some interface use but it doesn't seem easy given how the code is structured.

docs/monitors/system-utilization.md
```diff
@@ -1,34 +1,44 @@
...
 
-# signalfx-system-utilization
+# system-utilization
 
 
 This monitor reports utilization metrics for Windows
 Hosts.  It is used to drive the Windows Smart Agent host navigator view and dashboard
 content.
 
+Windows Performance Counters
```
keitwb: Formatting is off, either remove this line or make it a subheader with `##`.  You can use Markdown in the source Go comments.

docs/monitors/system-utilization.md
```diff
@@ -1,34 +1,44 @@
...
+Performance Counters.  The best way to mitigate this limitation is to increase
+the reporting interval on this monitor to collect more frequently.
+
 Sample YAML configuration:
 
 ```yaml
 monitors:
- - type: signalfx-system-utilization
+ - type: system-utilization
+   intervalSeconds: 2
```
keitwb: I wouldn't make this the sample because we don't have hi-res dimension handling implemented yet.

docs/monitors/telegraf-win_perf_counters.md
```diff
@@ -39,22 +39,22 @@ Monitor Type: `telegraf/win_perf_counters`
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
 | `objects` | no | `list of object (see below)` |  |
-| `counterRefreshInterval` | no | `integer` | number of nanoseconds that wildcards in counter paths should be expanded and how often to refresh counters from configuration (**default:** `60`) |
-| `useWildCardExpansion` | no | `bool` | if `true`, instance indexes will be included in instance names, and wildcards will be expanded and localized (if applicable).  If `false`, non partial wildcards will be expanded and instance names will not include instance indexes. (**default:** `false`) |
-| `printValid` | no | `bool` | print out the configurations that match available performance counters (**default:** `false`) |
+| `counterRefreshInterval` | no | `int64` | Number of nanoseconds that wildcards in counter paths should be expanded and how often to refresh counters from configuration (**default:** `5s`) |
```
keitwb: I just realized the type generated for `time.Duration` is misleading.  It is technically an int64 underneath but is parsed as a string, which is what should go here.  We'll just merge this as is and fix the doc generation later.

internal/monitors/telegraf/monitors/winperfcounters/win_perf_counters.go
```diff
@@ -41,34 +43,34 @@ func init() {
...
+	// Include the total instance when collecting performance counter metrics
 	IncludeTotal bool `yaml:"includeTotal" default:"false"`
 }
 
 // Config for this monitor
 type Config struct {
 	config.MonitorConfig `acceptsEndpoints:"false" deepcopier:"skip"`
 	Object               []PerfCounterObj `yaml:"objects" default:"[]"`
-	// number of nanoseconds that wildcards in counter paths should be expanded
+	// Number of nanoseconds that wildcards in counter paths should be expanded
```
keitwb: Is this really nanoseconds?

docs/development.md
```diff
@@ -120,7 +120,11 @@ non-test code.  Test code can fail if there is a good reason.
 
 ### Windows
 
-We develop on a [VirtualBox](https://www.virtualbox.org/) Windows Server 2008 [Vagrant](https://www.vagrantup.com/).
+We develop on a [VirtualBox](https://www.virtualbox.org/) Windows Server 2008 [Vagrant](https://www.vagrantup.com/).  You might want to develop on Windows Server 2012+ if you're using the evaluation boxes because the Windows Server 2008 evaluation only has a 10 day trial that can be renewed up to 5 times.
```
keitwb: Formatting is weird here.

docs/development.md
```diff
@@ -120,7 +120,11 @@ non-test code.  Test code can fail if there is a good reason.
...
+To renew the Windows Server 2008 evaluation you must manually reset the activation
+period by using the slmgr.vbs script from the command prompt and restart the vm.
+
+    slmgr.vbs –rearm
```
keitwb: Use backticks so this formats right.

PR #542, opened 2019-01-21
Hashicorp Vault as a Remote Config Source

keitwb:


docs/config-schema.md
```diff
@@ -249,6 +250,18 @@ The **nested** `consul` config object has the following fields:
...
 
+## vault
+The **nested** `vault` config object has the following fields:
+
+
+
+| Config option | Required | Type | Description |
+| --- | --- | --- | --- |
+| `vaultAddr` | no | string | The Vault Address.  Can also be provided by the standard Vault envvar `VAULT_ADDR`.  This option takes priority over the envvar if provided. |
+| `vaultToken` | no | string | The vault token, can also be provided by it the standard Vault envvar `VAULT_TOKEN`.  This option takes priority over the envvar if provided. |
```
jchengsfx: Capitalize `vault` for consistency?

rmfitzpatrick: Is this the standard precedence order for the agent?  In general, my expectations are cli arg > env var > file.  This way automated jobs can override on an environmental basis.

internal/core/config/sources/vault/keys.go
```diff
@@ -0,0 +1,41 @@
...
+	return nil
+}
+
+// The config path is of the form `/path/to/secret[keys.to.data]`
+func splitConfigPath(pathAndKey string) (string, string, error) {
+	if !strings.HasSuffix(pathAndKey, "]") {
+		return "", "", errInvalidPathFormat
+	}
+
+	parts := strings.SplitN(pathAndKey, "[", 2)
```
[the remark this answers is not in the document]
keitwb: No, in the secret data maps they could be valid characters since the secret data maps have string keys and they aren't any reserved chars.  Paths are more strict so it can be assumed there are no brackets in them.

docs/config-schema.md
keitwb: Yes, if it is in the agent config file, it is authoritative.

internal/core/config/sources/file/file.go
```diff
@@ -21,6 +21,13 @@ type Config struct {
 	PollRateSeconds int `yaml:"pollRateSeconds" default:"5"`
 }
 
+// New creates a new file remote config source from the target config
+func (c *Config) New() (types.ConfigSource, error) {
+	return New(time.Duration(c.PollRateSeconds) * time.Second), nil
+}
```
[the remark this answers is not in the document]
keitwb: It is done because there is no generic way to otherwise instantiate the config sources without listing them all out and having a bunch of duplicated code.  It allows https://github.com/signalfx/signalfx-agent/pull/542/files#diff-80d94174b8b7f9702d3f815d80ad347bL57 and all of the common validation logic to be shared.

PR #543, opened 2019-01-22
ASP.NET and .NET monitors

charliesignalfx:


tests/monitors/dotnet/dotnet_test.py
```diff
@@ -0,0 +1,56 @@
...
+    pytest.mark.windows,
+    pytest.mark.dotnet,
+]
+
+
+def test_dotnet():
+    config = dedent(
+        """
+        monitors:
+         - type: dotnet
```
keitwb: Does this just pull metrics for every .NET process on Windows or what?

charliesignalfx: yep it's exactly what our perfcounter monitor does now

PR #548, opened 2019-01-29
update filesystems monitor and update doc

charliesignalfx:
* add option to only pull physical disks in filesystem monitor
* add filtering doc
* add hostFSPath option to filesystems monitor


docs/monitors/net-io.md
```diff
@@ -28,7 +28,7 @@ Monitor Type: `net-io`
 
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
-| `interfaces` | no | `list of string` | The interfaces to include/exclude, is interpreted as a regex if surrounded by `/`. (**default:** `[* !/^lo\d*$/ !/^docker.*/ !/^t(un|ap)\d*$/ !/^veth.*$/ !/^Loopback*/]`) |
+| `interfaces` | no | `list of string` | The interfaces to include/exclude, is interpreted as a regex if surrounded by `/`.   For more information on filtering see [Generic Filters](https://github.com/signalfx/signalfx-agent/blob/master/docs/filtering.md#generic-filters). (**default:** `[* !/^lo\d*$/ !/^docker.*/ !/^t(un|ap)\d*$/ !/^veth.*$/ !/^Loopback*/]`) |
```
keitwb: This is a bit of a nit but I think this could be worded better.  The stuff about regex is redundant with the other doc.  Something like:

`The network interfaces to send metrics about.  This is a [filter set](https://github.com/...).`

Or something.

charliesignalfx: ok i'll use that ^

internal/monitors/filesystems/filesystems.go
```diff
@@ -77,6 +89,10 @@ func (m *Monitor) getCommonDimensions(partition *gopsutil.PartitionStat) map[str
 		"device":          strings.Replace(partition.Device, " ", "_", -1),
 		"plugin_instance": "",
 	}
+	if m.hostFSPath != "" {
+		// strip configured hostFSPath from mountpoint
+		dims["mountpoint"] = strings.Replace(partition.Mountpoint, m.hostFSPath, "", 1)
```
keitwb: Shouldn't this be `dims["mountpoint"] = strings.Replace(dims["mountpoint"],...)` to take advantage of the space to underscore conversion above?

internal/monitors/filesystems/filesystems.go
```diff
@@ -85,14 +84,10 @@ type Monitor struct {
 // returns common dimensions map according to reportInodes configuration
 func (m *Monitor) getCommonDimensions(partition *gopsutil.PartitionStat) map[string]string {
 	dims := map[string]string{
-		"mountpoint":      strings.Replace(partition.Mountpoint, " ", "_", -1),
+		"mountpoint":      strings.Replace(strings.Replace(partition.Mountpoint, m.hostFSPath, "", 1), " ", "_", -1),
```
keitwb: What happens if `m.hostFSPath` is `""`?  Is the inner `strings.Replace` a noop in that case?

charliesignalfx: yeah it is

keitwb: `If old is empty, it matches at the beginning of the string and after each UTF-8 sequence, yielding up to k+1 replacements for a k-rune string.`

Issue #549, opened 2019-01-29
Chef cookbook agent config yaml doesn't work with smart agent


pvanliew:
Currently, the chef cookbook will create an agent.yaml from conf attributes with json-ified yaml, that the java smart agent, in this case, can't use, and so does not start up.

Such as:
```
monitors: [{type: "trace-forwarder", listenAddress: "127.0.0.1:9080"}, {"type": "collectd/cpu"}]
signalFxAccessToken: "howdy"
traceEndpointUrl: "http://somewhere:8080/v1/trace"
writer: {"sendTraceHostCorrelationMetrics": true}
```

When correct yaml that the agent needs to start is:
```
monitors:
  - type: trace-forwarder
    listenAddress: 127.0.0.1:9080
  - type: collectd/cpu
signalFxAccessToken: howdy
traceEndpointUrl: http://somewhere:8080/v1/trace
writer:
  sendTraceHostCorrelationMetrics: true
```

keitwb:
Merged #550.

PR #550, opened 2019-01-29
Updates chef cookbook agent conf yaml templating.

pvanliew:
Addresses https://github.com/signalfx/signalfx-agent/issues/549 to make the resulting yaml actually yaml, without json for sub-keys, or quotes on values, and thus consumable by the smart agent.

yaml is able to be required by default in all supported distros.


keitwb:
Thanks @pvanliew.  The old thing was a corrupted form of YAML (and obviously not true JSON) that apparently sometimes worked, but you have found a clear case where it doesn't.

PR #554, opened 2019-01-30
k8s pod:service cache and refactoring cache key

mstumpfx:


internal/monitors/kubernetes/cluster/metrics/cache.go
```diff
@@ -232,9 +259,67 @@ func (dc *DatapointCache) AllDimProperties() []*atypes.DimProperties {
...
+	cacheKey := key.(types.UID)
+	podUIDs := dc.podServiceCache.GetPodUIDsForServiceUID(cacheKey)
+	dc.podServiceCache.DeleteServiceFromCache(cacheKey)
+	dc.updateServicePropForPods(podUIDs)
+}
+
+// updateServicePropForPods takes a list of pod UIDs, gets the matching
+// service for the pod, and adds the service property to the pod if one exists
+func (dc *DatapointCache) updateServicePropForPods(podUIDs []types.UID) {
+
```
keitwb: Remove blank line.

internal/monitors/kubernetes/utils/servicepodcache.go
```diff
@@ -0,0 +1,239 @@
...
+// GetServiceNameForPod deterministically returns a single service
+// matched by a pods labels. Used by the DataPoint cache for
+// updating the "service" property, which only may have one value.
+// Selection method is sorting strings alphabetically and selecting the
+// first service off the top.
+func (psc *PodServiceCache) GetServiceNameForPod(pod *v1.Pod) (string, error) {
+	var service string
+	services := psc.getServiceNamesForPod(pod)
+	if len(services) == 0 {
+		return service, errors.New("no service matched pod")
```
keitwb: This is not very idiomatic for Go.  I would just make the return value for this func `*string` and then check for nil in the functions that call this to determine whether no service matched a pod or not.

keitwb: Errors are usually reserved for true errors, not things that are expected on a pretty regular basis.

internal/monitors/kubernetes/utils/servicepodcache.go
```diff
@@ -0,0 +1,239 @@
...
+// GetServiceNameForPodUID deterministically returns a single service
+// matched by a pods labels. Used by the DataPoint cache for
+// updating the "service" property, which only may have one value.
+// Selection method is sorting strings alphabetically and selecting the
+// first service off the top.
+func (psc *PodServiceCache) GetServiceNameForPodUID(podUID types.UID) (string, error) {
+	var service string
+	services := psc.getServiceNamesForPodUID(podUID)
+	if len(services) == 0 {
+		return service, errors.New("no service matched pod")
```
keitwb: Same thing.

internal/monitors/types/dims.go
```diff
@@ -17,3 +17,27 @@ type Dimension struct {
 	// Value of the dimension
 	Value string
 }
+
+// Copy creates a copy of the the given dimProps object
+func (dp *DimProperties) Copy() *DimProperties {
+	clonedDimension := Dimension{
```
keitwb: This copies by just doing `clonedDimension := dp.Dimension`

internal/monitors/kubernetes/utils/podcache.go
```diff
@@ -0,0 +1,101 @@
...
+		(reflect.DeepEqual(cachedOR, pod.OwnerReferences))
+}
+
+// AddPod adds or updates a pod in cache
+func (pc *PodCache) AddPod(pod *v1.Pod) {
+	if _, exists := pc.namespacePodUIDCache[pod.Namespace]; !exists {
+		pc.namespacePodUIDCache[pod.Namespace] = make(map[types.UID]bool)
+	}
+	pc.namespacePodUIDCache[pod.Namespace][pod.UID] = true
+	pc.podUIDNamespaceCache[pod.UID] = pod.Namespace
```
keitwb: Turn this inside out so that you have a single map that has a key of the uid and values of a struct that contain all of the necessary pod info that we need.

keitwb: Make that struct the thing that you pass to the dimPropsForPod function instead of a combination of pod uid and this podcache.

internal/core/writer/properties.go
```diff
@@ -81,7 +81,7 @@ func (dpc *dimensionPropertyClient) SetPropertiesOnDimension(dimProps *types.Dim
 		return nil
 	}
 
-	if !dpc.isDuplicate(filteredDimProps) {
+	if !dpc.isDuplicate(filteredDimProps.Copy()) {
```
keitwb: Take the copy earlier than this.

internal/core/writer/properties.go
```diff
@@ -76,7 +76,7 @@ func newDimensionPropertyClient(conf *config.WriterConfig) (*dimensionPropertyCl
 func (dpc *dimensionPropertyClient) SetPropertiesOnDimension(dimProps *types.DimProperties) error {
 	filteredDimProps := &(*dimProps)
 
-	filteredDimProps = dpc.PropertyFilterSet.FilterDimProps(filteredDimProps)
+	filteredDimProps = dpc.PropertyFilterSet.FilterDimProps(filteredDimProps).Copy()
```
keitwb: This will blow up if `filteredDimProps` is nil, but I don't think it can be nil at this point because of how `FilterDimProps` is implemented.  So I would remove the nil check below this line to make it less confusing about the potential states of `filteredDimProps`.

PR #558, opened 2019-02-01
suppress context deadline exceeded log message and report it as a deb…

charliesignalfx:
…ug statement

### DIFF
--- a/internal/monitors/cpu/cpu.go
+++ b/internal/monitors/cpu/cpu.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/shirou/gopsutil/cpu"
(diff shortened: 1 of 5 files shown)


internal/monitors/cpu/cpu.go
```diff
@@ -110,7 +119,11 @@ func (m *Monitor) emitDatapoints() {
 
 		// append errors
 		if err != nil {
-			logger.WithError(err).Errorf("failed to calculate utilization for cpu")
+			if strings.HasSuffix(err.Error(), "are < 0") {
```
keitwb: Split up the error into two distinct errors (one for usedDiff and one for totalDiff) and make them constants at the top of the module (the convention is to name them `var err... =`.  You don't need to put the value in the error message, nobody cares about the exact value.

PR #561, opened 2019-02-02
update doc to fix rst related failures in product docs

charliesignalfx:


molner:
@jchengsfx - at least part of the problem here appear to be setup failures. Any idea?

jchengsfx:
> @jchengsfx - at least part of the problem here appear to be setup failures. Any idea?

@molner - Looks to be issues with apt repos, so the required docker images required for the tests are failing to build.

PR #562, opened 2019-02-04
Rely on windows dll for per core cpu times instead of wmi

charliesignalfx:


internal/monitors/cpu/cpu_windows.go
```diff
@@ -0,0 +1,153 @@
...
+	if !perCore {
+		return gopsutilTimes(perCore)
+	}
+	// Underneath the hood gopsutil relies on a wmi query for per core cpu utilization information
+	// this wmi query has proven to be problematic under unclear conditions.  It will hang
+	// from time to time, and when executed frequently.  Many projects rely on gopsutil for this information.
+	// Some have issues open complaining about hanging wmi calls, but none have a clear solution.
+	// In general if you search for information about WMI it is the best of a bad situation.  It
+	// is known to be buggy and slow.  A more performant solution is to
+	// get the System Processor Performance Information from the ntQuerySystemInformationfunction.
```
keitwb: space before "function"

internal/monitors/cpu/cpu_windows.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// attempt to use the windows api to get information on per core cpu times
+	res, err := ntQuerySystemInformation()
+	if err == nil {
+		return res, err
```
keitwb: This is misleading returning `err` when you know it is nil.  Just return `res, nil` and it is clearer.

PR #566, opened 2019-02-06
Update docs for postgresql monitor

asuresh4:


internal/monitors/collectd/postgresql/postgresql.go
```diff
@@ -89,11 +96,11 @@ type Database struct {
...
-	// Queries used to generate metrics.  If no queries are specified, the
-	// default set will be used [`custom_deadlocks`, `backends`, `transactions`,
-	// `queries`, `queries_by_table`, `query_plans`, `table_states`,
-	// `query_plans_by_table`, `table_states_by_table`, `disk_io`,
-	// `disk_io_by_table`, `disk_usage`]
+	// Queries used to generate metrics. These will override the default set.
+	// If no queries are specified, the default set will be used
+	// [`custom_deadlocks`, `backends`, `transactions`, `queries`, `queries_by_table`,
+	// `query_plans`, `table_states`, `query_plans_by_table`, `table_states_by_table`,
+	//  `disk_io`, `disk_io_by_table`, `disk_usage`]
```
charliesignalfx: I think the extra space here is breaking the script that generates the md

internal/monitors/collectd/postgresql/postgresql.go
```diff
@@ -55,12 +55,19 @@ const monitorType = "collectd/postgresql"
...
+//       instancePrefix: "test"
 //  databases:
 //  - name: "test"
 //    username: "username2"
 //    password: "password2"
 //    queries:
 //    - "exampleQuery"
+//
+//
+//metricsToInclude:
```
keitwb: This needs some clarification as to why it is necessary.  I see that it has something to do with the `instancePrefix` but should this be a glob like `- gauge.test*` since _test_ is a prefix?

keitwb: Also this is a great example of why we need to get rid of the whitelist filtering and make it inherent to the monitor so that people wouldn't have to jump through these hoops to make a metric come through.

internal/monitors/collectd/postgresql/postgresql.go
keitwb: Yes, it is.

asuresh4: Removed the space

internal/monitors/collectd/postgresql/postgresql.go
asuresh4: @keitwb I've added a note about how the metric name is made. Could you take a look and let me know if the example is clear now?

keitwb:
Don't worry about merging master in.  You can go ahead an merge it.

PR #567, opened 2019-02-07
upgrade telegraf dependencies

charliesignalfx:


internal/monitors/telegraf/common/accumulator/accumulator.go
```diff
@@ -1,12 +1,21 @@
...
 
-	"github.com/signalfx/golib/datapoint"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/metric"
 	"github.com/signalfx/signalfx-agent/internal/monitors/telegraf/common/emitter"
 )
 
+func getTime(times []time.Time) time.Time {
+	if len(times) > 0 {
+		return times[0]
```
keitwb: The first item is the latest time available?

internal/monitors/telegraf/monitors/procstat/procstat.go
```diff
@@ -83,6 +92,11 @@ func (m *Monitor) Configure(conf *Config) (err error) {
 	// create the emitter
 	em := baseemitter.NewEmitter(m.Output, logger)
 
+	// use the agent's configured host sys to get cgroup information
+	if conf.CGroup != "" {
```
keitwb: Shouldn't this be the opposite check for `CGroup` being blank?  It looks like you are setting a default here.

internal/monitors/telegraf/common/accumulator/accumulator.go
```diff
@@ -57,9 +76,69 @@ func (ac *Accumulator) AddError(err error) {
...
+// Delivered -
+func (ta *TrackingAccumulator) Delivered() <-chan telegraf.DeliveryInfo {
+	return ta.done
+}
+
+func (ta *TrackingAccumulator) deliveryCallback(i telegraf.DeliveryInfo) {
+	select {
+	case ta.done <- i:
+	default:
+		ta.Accumulator.emit.AddDebug(fmt.Sprintf("unable to return delivery Status '%t' for ID '%v' channel is full", i.Delivered(), i.ID()))
```
keitwb: If you make `AddDebug`'s signature `(msg string, args ...interface{})` then you can just do the `Sprintf` in that method and use it like the normal `logrus.Debugf` function.

internal/monitors/telegraf/common/accumulator/accumulator.go
```diff
@@ -57,9 +76,69 @@ func (ac *Accumulator) AddError(err error) {
...
 // NewAccumulator returns a pointer to an Accumulator
 func NewAccumulator(e emitter.Emitter) *Accumulator {
 	return &Accumulator{
 		emit: e,
 	}
 }
+
+// TrackingAccumulator will signal after a metric has been processed.  It indicates
+// if the metric was accepted, rejected, or dropped
+type TrackingAccumulator struct {
```
keitwb: What is the point of this?  I don't see it used anywhere.

internal/monitors/telegraf/common/accumulator/accumulator.go
charliesignalfx: yeah this is what telegraf does too.  It addresses a design issue from their old metric model.

internal/monitors/telegraf/monitors/procstat/procstat.go
charliesignalfx: If the cgroup name is specified, we explicitly build it with Host Sys.  This is to address behavior that's hard coded in the underlying telegraf plugin.  https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L388

internal/monitors/telegraf/common/accumulator/accumulator.go
charliesignalfx: I'm still trying to understand that myself, they added it and so we have to conform to their interface.  It's a mechanism for input plugins to track the status of a metric they've emitted.

internal/monitors/telegraf/common/accumulator/accumulator.go
charliesignalfx: ok will update

internal/monitors/telegraf/monitors/procstat/procstat.go
keitwb: ok, shouldn't the value that gets reassigned to `conf.CGroup` somehow incorporate the original `conf.CGroup` value?  Otherwise this is just saying if you set the CGRoup option to something it will always be overridden by this hardcoded value.

charliesignalfx: yes, yes it should.  just fixed it.

PR #576, opened 2019-02-13
adding docker memory stats

mstumpfx:
if anyone wants to take a shot at giving these definitions they're here but not really defined so well. 
https://docs.docker.com/config/containers/runmetrics/#metrics-from-cgroups-memory-cpu-block-io


keitwb:
For the ones that are documented in the link, can you just copy/paste those descriptions to the metric docs?  That or a link to them maybe.  We need some kind of description.

PR #584, opened 2019-02-19
Remove duplicate config docs for metricsToExclude

cory-signalfx:
# Synopsis
Removes some duplicate documentation.

# Motivation
Duplicates!

# Notes
Is this generated by something? I couldn't find anything from grepping…

# Testing
None needed!

### DIFF
--- a/docs/config-schema.md
+++ b/docs/config-schema.md
@@ -7,7 +7,7 @@ by the `-config` flag to the agent binary (`signalfx-agent`).
 
 ## Config Schema
 
-  
+
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
 | `signalFxAccessToken` | no | string | The access token for the org that should receive the metrics emitted by the agent. |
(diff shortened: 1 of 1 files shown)


mstumpfx:
@cory-signalfx These are indeed auto-generated from the config definition files themselves (for this one see https://github.com/signalfx/signalfx-agent/blob/master/internal/core/config/config.go)

We use a script (https://github.com/signalfx/signalfx-agent/blob/master/scripts/docs/make-docs) to generate them.

I'm not sure this is a duplicate though either - theres `metricsToExclude` and `propertiesToExclude` which are two different options with similar nested config options.

There's some info in our development guide (https://github.com/signalfx/signalfx-agent/blob/master/docs/development.md) about this, although now I'm re-reading it and maybe not for doc generation so that's a gap.

P.s. welcome!

keitwb:
There is both a `metricsToInclude` and a `metricsToExclude` section.  `metricsToExclude` acts as a pure blacklist and `metricsToInclude` acts like an override to the blacklist in `metricsToExclude`.  Confusingly enough, `metricsToExclude` can also be specified at the monitor config block level, but not `metricsToInclude` (although this could be done pretty easily if the use case arose), mainly because `metricsToInclude` was added as a stop-gap solution until we could make our monitors inherently follow the whitelist and then get rid of the whitelist filter config altogether.

cory-signalfx:
Thanks for the explanation!

PR #589, opened 2019-02-21
Multi-realm Updates

mstumpfx:
A new realm config option that fills in the endpoint URLs automatically, and also doc changes.


keitwb:
This should not be merged until the alias for us0 gets activated (it was not as of now).

mstumpfx:
@keitwb should we update the installer script to pass through the realmConfig option instead of using ingest and api urls?

keitwb:
Yeah that's a good idea.

mstumpfx:
@keitwb multi-realm API endpoints are live so this is good to go

PR #591, opened 2019-02-22
Add simple load testing with Kubernetes

keitwb:
This involves reworking the fake backends to make them more performant,
as well as making the fake K8s unit test helper able to run standalone.

This can be expanded greatly in the future but this gets us a start


keitwb:
Still trying to get the new pytest requirements buildable in Windows.  vcredis2015 isn't installing there right with choco.

PR #593, opened 2019-02-22
Add Windows support to Chef Cookbook

keitwb:


keitwb:
@mitom Tagging you in case you have any feedback before we release.

PR #603, opened 2019-02-25
WIP - AWS Fargate Integration

seonsfx:


tests/observers/ecs_test.py
```diff
@@ -0,0 +1,89 @@
...
+      ecsmeta.remove()
+      redis.kill()
+      redis.remove()
+
+      # Let redis be removed by docker observer and collectd restart
+      time.sleep(5)
+      backend.datapoints.clear()
+      assert ensure_always(lambda: not has_datapoint_with_dim(backend, "ClusterName", "seon-fargate-test"), 10)
+
+
```
seonsfx: @keitwb Need help here. I have three integration test cases and they all run fine individually if I comment other two. However, when I run all three at once, it tells me there's conflict with docker container name. Looks like it tries to create same container several times even though the container was already created. Spent few hours and tried many different things, still can't figure out why.

deployments/fargate/README.md
```diff
@@ -0,0 +1,72 @@
+# Fargate Deployment
+
+## Create Task Definition
+To deploy the agent on AWS Elastic Container Service (ECS) you must first
+create the agent task definition.  To do this using the web admin console:
+
+ 1. Prepare task definition. You can either take signalfx-agent container
+ 	definition from the file [signalfx-agent-fargate-task.json](./signalfx-agent-fargate-task.json) or add
```
keitwb: Formatting here is off.

deployments/fargate/README.md
```diff
@@ -0,0 +1,72 @@
...
+	your existing fargate container definitions to the file.
+	- Make sure all your fargate containers to monitor has docker labels to specify ports to be monitored.
```
keitwb: > Make sure all ... *have* Docker labels...

deployments/fargate/README.md
```diff
@@ -0,0 +1,72 @@
+# Fargate Deployment
+
+## Create Task Definition
+To deploy the agent on AWS Elastic Container Service (ECS) you must first
```
keitwb: > To deploy the agent on AWS *Fargate* ...

deployments/fargate/README.md
```diff
@@ -0,0 +1,72 @@
...
+create the agent task definition.  To do this using the web admin console:
+
+ 1. Prepare task definition. You can either take signalfx-agent container
+ 	definition from the file [signalfx-agent-fargate-task.json](./signalfx-agent-fargate-task.json) or add
+	your existing fargate container definitions to the file.
+	- Make sure all your fargate containers to monitor has docker labels to specify ports to be monitored.
+```json
+	"dockerLabels": {
+		"agent.signalfx.com.port.6379": "true",
```
keitwb: Explain what this does.

deployments/fargate/README.md
```diff
@@ -0,0 +1,72 @@
...
+	 - `Task Definition (Revision)` -> `1` (or whatever the latest is in your case)
+	 - `Service Name` -> `signalfx-fargate` (or any good name that explains you service)
+     - `Task Definition (Revision)` -> `1` (or whatever the latest is in your case)
+     - `Number of tasks` is also required for fargate service configuration
+ 5. Leave everything else at default and click "Next step"
+ 6. The second step is configuring network. Fargate requires to run with `awsvpc`
+    network type.
+    After providing all required network settings, click "Next step"
+ 7. Leave everything on this next page at their defaults and click "Next step".
+ 8. Click "Create Service" and the agent should be deployed with other fargate
```
keitwb: So the service feature causes the agent to run automatically as a sidecar everywhere?

deployments/fargate/agent.yml
```diff
@@ -0,0 +1,65 @@
...
+observers:
+    # Enable the docker observer to discover other ECS tasks running on the
+    # same EC2 instance.
+  - type: ecs
+    # ECS tasks running the awsvpc network mode don't actually expose and IP
+    # address on the docker inspect API so we need to use the hostname instead.
+    labelsToDimensions: &labelMap
+      # We map the cluster label to ClusterName to match what our AWS
+      # CloudWatch integration uses.
+      com.amazonaws.ecs.cluster: ClusterName
```
keitwb: We don't need these since the task metadata is gotten from the ECS Metadata agent.

deployments/fargate/signalfx-agent-fargate-task.json
```diff
@@ -0,0 +1,53 @@
...
+                    "name": "CONFIG_URL",
+                    "value": "https://raw.githubusercontent.com/signalfx/signalfx-agent/master/deployments/ecs/agent.yaml"
+                }
+            ],
+            "dockerLabels": {
+                "app": "signalfx-agent"
+            },
+            "privileged": null,
+            "name": "signalfx-agent",
+            "image": "quay.io/signalfx/signalfx-agent:4.0.0"
```
keitwb: Update to 4.0.2

docs/monitors/ecs-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+| `excludedImages` | no | `list of string` | A list of filters of images to exclude.  Supports literals, globs, and regex. |
+
+
+
+
+## Metrics
+
+The following table lists the metrics available for this monitor. Metrics that are not marked as Custom are standard metrics and are monitored by default.
+
+| Name | Type | Custom | Description |
```
keitwb: Talk to @mstumpfx about how to get these in the whitelist properly.  Mark, they are the same as what is emitted by the `docker-container-stats` monitor.

internal/core/common/ecs/ecs.go
```diff
@@ -0,0 +1,40 @@
...
+	Family		string `json:"Family"`
+	Revision	string `json:"Revision"`
+	KnownStatus	string `json:"KnownStatus"`
+	Containers  []Container `json:"Containers"`
+}
+
+func (task *TaskMetadata) GetDimensions() map[string]string {
+	dims := make(map[string]string, 0)
+	if idx := strings.Index(task.ClusterName, "/"); idx >= 0 {
+		dims["ClusterName"] = task.ClusterName[idx+1: len(task.ClusterName)]
```
keitwb: Match the same dimension names used with the sample config for ECS.

keitwb: Also general naming convention for dimension names is lower case snake_case or kebab-case.

docs/monitors/ecs-metadata.md
mstumpfx: hm yeah I can help you with this @seonsfx - are we building new dashboards for this integration?

internal/monitors/ecs/ecs.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+
+	isRegistered := false
+
+	utils.RunOnInterval(m.ctx, func() {
+		if !isRegistered {
+			m.containers = map[string]ecs.Container{}
+			m.shouldIgnore = map[string]bool{}
+			m.fetchContainers()
```
keitwb: Make this return the set of containers and assign it to `m.containers` instead of implicitly sharing the variable through the struct instance.

internal/monitors/ecs/ecs.go
```diff
@@ -0,0 +1,262 @@
...
+	// Usually the container was filtered out by excludedImages
+	// or container metadata is not received.
+	shouldIgnore map[string]bool
+	imageFilter filter.StringFilter
+}
+
+// Configure the monitor and kick off volume metric syncing
+func (m *Monitor) Configure(conf *Config) error {
+	var err error
+	m.imageFilter, err = filter.NewBasicStringFilter(conf.ExcludedImages)
```
keitwb: Use the _exhausive_ string filter in that same package.  It is more powerful.

internal/monitors/ecs/ecs.go
keitwb: Same thing with the black list (`shouldIgnore`).

internal/monitors/ecs/ecs.go
```diff
@@ -0,0 +1,262 @@
...
+	return true
+}
+
+func (m *Monitor) fetchStatsForAll() {
+	_, cancel := context.WithTimeout(m.ctx, m.timeout)
+	
+	body, err := getMetadata(m.client, m.conf.StatsEndpoint)
+
+	if err != nil {
+		cancel()
```
keitwb: You should always cancel, even if there is no error.  It is just a small efficiency gain that frees the resources faster.

internal/monitors/ecs/ecs.go
```diff
@@ -0,0 +1,262 @@
...
+		logger.WithError(err).Error("Failed to read ECS stats - %v", err)
```
keitwb: You don't need to format the error into the error string if you use `WithError(err)`.

docs/monitors/ecs-metadata.md
seonsfx: @mstumpfx I understood so. Something very similar with name `ECS Metadata`

internal/observers/ecs/ecs.go
```diff
@@ -0,0 +1,293 @@
...
+
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		logger.WithError(err).Error("failed to read ECS metadata response body - %v", err)
+		return nil
+	}
+
+	var metadata ecs.TaskMetadata
+
+	if err := json.Unmarshal(body, &metadata); err != nil {
```
keitwb: You can do the decoding like this and it's a bit simpler:

```
if err := json.NewDecoder(response.Body).Decode(&metadata); err != nil {
```

then you don't need the intermediate `body` var or error check.

Dockerfile
```diff
@@ -300,7 +300,8 @@ COPY --from=python-plugins /usr/local/lib/python2.7/ /lib/python2.7
 COPY --from=python-plugins /usr/bin/python /bin/python
 
 COPY scripts/umount-hostfs-non-persistent /bin/umount-hostfs-non-persistent
-COPY deployments/docker/agent.yaml /etc/signalfx/agent.yaml
+ARG build_target=docker
+COPY deployments/${build_target}/agent.yaml /etc/signalfx/agent.yaml
```
keitwb: Did you mean to commit this?  I would rather it not be because people are going to have to override the fargate config in 99% of the cases to get useful application metrics so there is little point to making a build with some standard config.  It would be better to enhance the standard Docker agent.yaml file to work on both with envvar references used generously.

Makefile
```diff
@@ -44,6 +44,13 @@ endif
 image:
 	COLLECTD_VERSION=$(COLLECTD_VERSION) COLLECTD_COMMIT=$(COLLECTD_COMMIT) ./scripts/build
 
+.PHONY: local-image
+local-image:
```
keitwb: Remove.

deployments/fargate/signalfx-agent-fargate-task.json
```diff
@@ -0,0 +1,46 @@
...
+                "curl --fail $CONFIG_URL > /etc/signalfx/agent.yaml && exec /bin/signalfx-agent"
+            ],
+            "environment": [
+                {
+                    "name": "ACCESS_TOKEN",
+                    "value": "MY_ACCESS_TOKEN"
+                },
+                {
+                    "name": "CONFIG_URL",
+                    "value": "https://raw.githubusercontent.com/signalfx/signalfx-agent/master/deployments/ecs/agent.yaml"
```
keitwb: This should probably be from the `fargate` directory.

deployments/fargate/signalfx-agent-fargate-task.json
```diff
@@ -0,0 +1,46 @@
...
+                }
+            ],
+            "dockerLabels": {
+                "app": "signalfx-agent"
+            },
+            "name": "signalfx-agent",
+            "image": "quay.io/signalfx/signalfx-agent:4.0.2"
```
keitwb: Make sure this line gets updated by the script in `scripts/update-deployment-versions`

docs/observers/ecs.md
```diff
@@ -0,0 +1,148 @@
+<!--- GENERATED BY gomplate from scripts/docs/observer-page.md.tmpl --->
+
+# ecs
+
+ Queries the [ECS Task Metadata Endpoint version 2](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v2.html) for running containers.
+
+The Smart Agent needs to run in the same task with the containers need
```
keitwb: Remove _need_ at end of line.

docs/observers/ecs.md
```diff
@@ -0,0 +1,148 @@
...
+to be monitored to get access to the data through ECS task metadata endpoint.
+
+## Configuration from Labels
+You can configure monitors by putting special labels on your Docker
+containers in ECS task definition. You can either specify all of the
```
keitwb: in _the_ ECS task definition.

docs/observers/ecs.md
```diff
@@ -0,0 +1,148 @@
...
+`goruntime` endpoint getting the default interval of the agent.
+
+
+Observer Type: `ecs`
+
+[Observer Source Code](https://github.com/signalfx/signalfx-agent/tree/master/internal/observers/ecs)
+
+## Configuration
+
+This observer has no configuration options.
```
keitwb: Should probably allow the endpoint overrides like you did in the monitor.

internal/core/common/ecs/ecs.go
```diff
@@ -0,0 +1,40 @@
...
+func (task *TaskMetadata) GetDimensions() map[string]string {
+	dims := make(map[string]string, 0)
+	if idx := strings.Index(task.ClusterName, "/"); idx >= 0 {
+		dims["cluster"] = task.ClusterName[idx+1: len(task.ClusterName)]
+	} else {
+		dims["cluster"] = task.ClusterName
+	}
+	dims["ecs-task-arn"] = task.TaskARN
+	dims["ecs-task-group"] = task.Family
+	dims["ecs_task_version"] = task.Revision
```
keitwb: Inconsistent between kebab-case and snake_case.

keitwb: Also, the ECS standard label config uses snake_case so this one is right and the other two should change.

We use the titlecased `ClusterName` for the task's cluster name field so that it matches what CloudWatch metrics use.

internal/core/common/ecs/ecs.go
```diff
@@ -0,0 +1,40 @@
...
+	dims["known-status"] = task.KnownStatus
```
keitwb: This is risky putting as a dimensions.  It kind of sounds more like a property.  If it changes, it will produce a new timeseries for what is the same "thing".

internal/monitors/docker/docker.go
```diff
@@ -172,7 +174,21 @@ func (m *Monitor) fetchStats(container dockerContainer, labelMap map[string]stri
...
+		cancel()
+		// EOF means that there aren't any stats, perhaps because the container
+		// is gone.  Just return nothing and no error.
+		if err == io.EOF {
+			return
+		}
+		logger.WithError(err).Errorf("Could not parse docker stats for container id %s", container.ID)
+		return
+	}
+	stats.Body.Close()
```
keitwb: Make sure it gets closed in the case of errors above.

internal/core/common/ecs/ecs.go
```diff
@@ -0,0 +1,40 @@
...
+type TaskMetadata struct {
+	ClusterName	string `json:"Cluster"`
+	TaskARN		string `json:"TaskARN"`
+	Family		string `json:"Family"`
+	Revision	string `json:"Revision"`
+	KnownStatus	string `json:"KnownStatus"`
+	Containers  []Container `json:"Containers"`
+}
+
+func (task *TaskMetadata) GetDimensions() map[string]string {
```
keitwb: We also need to figure out whether it is appropriate to send an `AWSUniqueId` dimension with this or not.  Another question is whether the agent is going to discover the AWSUniqueId of the underlying host or not.  Do you see AWSUniqueId as a dimension already on stuff you have been testing with? If so, we might need to disable the auto determination of it from the EC2 metadata endpoint.

internal/monitors/ecs/ecs.go
```diff
@@ -0,0 +1,275 @@
...
+	isRegistered := false
+
+	utils.RunOnInterval(m.ctx, func() {
+		if !isRegistered {
+			task, err := fetchTaskMetadata(m.client, m.conf.MetadataEndpoint)
+			if err != nil {
+				logger.WithFields(log.Fields{
+					"error":    err,
+				}).Error("Could not receive ECS Task Metadata")
+				m.cancel()
```
keitwb: You don't want to cancel in this case, it will mean that it will never try again since your context is tied to the RunOnInterval loop..

keitwb: Just return and let it try again after `conf.IntervalSeconds`

internal/monitors/ecs/ecs.go
```diff
@@ -0,0 +1,275 @@
...
+	if (m.imageFilter != nil && m.imageFilter.Matches(container.Image)) ||
+		container.Type != "NORMAL" {
+		return ecs.Container{}, errors.New("The container has been excluded by image filter.")
+	}
+
+	return container, nil
+}
+
+func (m *Monitor) fetchStatsForAll() {
+	_, cancel := context.WithTimeout(m.ctx, m.timeout)
```
keitwb: If you don't use the context returned, there is no point in calling the `cancel` method, as it only acts on the context you are ignoring with `_` here.

keitwb: You probably don't need any context here.

internal/monitors/ecs/ecs.go
```diff
@@ -0,0 +1,275 @@
...
+
+func fetchTaskMetadata(client *http.Client, metadataEndpoint string) (ecs.TaskMetadata, error) {
+	body, err := getMetadata(client, metadataEndpoint)
+	if err != nil {
+		return ecs.TaskMetadata{}, errors.Wrapf(err, "Failed to read ECS container data")
+	}
+
+	task := ecs.TaskMetadata{}
+	if err := json.Unmarshal(body, &task); err != nil {
+		return ecs.TaskMetadata{}, errors.Wrapf(err, "ould not parse ecs metadata json")
```
keitwb: "ould"

scripts/common.sh
```diff
@@ -16,6 +16,7 @@ do_docker_build() {
   local operating_system=${5:-"linux"}
   local collectd_commit=${COLLECTD_COMMIT}
   local collectd_version=${COLLECTD_VERSION}
+  local build_target=${BUILD_TARGET:-"docker"}
```
keitwb: Remove.

scripts/common.sh
```diff
@@ -30,6 +31,7 @@ do_docker_build() {
     --build-arg GOOS=${operating_system} \
     --build-arg collectd_version=${collectd_version} \
     --build-arg collectd_commit=${collectd_commit} \
+    --build-arg build_target=${build_target} \
```
keitwb: Remove.

test-services/redis/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM redis
```
keitwb: If you aren't going to do anything useful with the image, just run this redis image directly in your tests with `run_container("redis")...` and don't make an unnecessary test-service.

Same thing for Mongo.

internal/monitors/docker/metircs_basic.go
```diff
@@ -0,0 +1,25 @@
+package docker
```
keitwb: File name is spelled wrong.

internal/core/config/monitor.go
```diff
@@ -51,6 +51,10 @@ type MonitorConfig struct {
 	// is useful when you have an endpoint whose identity is not particularly
 	// important since it acts largely as a proxy or adapter for other metrics.
 	DisableEndpointDimensions bool `yaml:"disableEndpointDimensions" json:"disableEndpointDimensions"`
+	// Whether it will send all custom and non-basic metrics as well.
+	// For the monitors which utilize this config, a list of basic metrics will be managed within the
+	// monitor module.
+	EnhancedMetrics bool `yaml:"enhancedMetrics" json:"enhancedMetrics" default:"false"`
```
keitwb: This is too generic of a place to put this.  It definitely doesn't apply to all monitors.  What I was saying about this struct is that it is embedded, just like you could also do with another struct that is common to ECS and Docker monitors that contains the shared enhanced metrics flags.  Like this:

```
type EnhancedMetricsConfig struct {
    EnableExtraBlockIOMetrics bool `yaml:"enableExtraBlockIOMetrics"`
   ...
}
```

That could be used like:

```
type Config struct{
    config.MonitorConfig
    docker.EnhancedMetricsConfig
...
}
```

internal/monitors/docker/conversion.go
```diff
@@ -23,25 +21,25 @@ var memoryStatCounters = map[string]bool{
...
-		// EOF means that there aren't any stats, perhaps because the container
-		// is gone.  Just return nothing and no error.
-		if err == io.EOF {
-			return nil, nil
-		}
-		return nil, err
-	}
-	cstats.Body.Close()
-
+func ConvertStatsToMetrics(container *dtypes.ContainerJSON, parsed *dtypes.StatsJSON, enhancedMetrics bool) ([]*datapoint.Datapoint, error) {
```
keitwb: Pass in that embedded config struct with all the flags and the split it out here to the various `convert*` functions.

internal/monitors/docker/conversion.go
```diff
@@ -23,25 +21,25 @@ var memoryStatCounters = map[string]bool{
...
 	var dps []*datapoint.Datapoint
 	dps = append(dps, convertBlkioStats(&parsed.BlkioStats)...)
 	dps = append(dps, convertCPUStats(&parsed.CPUStats, &parsed.PreCPUStats)...)
 	dps = append(dps, convertMemoryStats(&parsed.MemoryStats)...)
 	dps = append(dps, convertNetworkStats(&parsed.Networks)...)
 
+	if !enhancedMetrics {
```
keitwb: Get rid of this.

internal/core/config/monitor.go
seonsfx: That makes sense.

seonsfx:
@keitwb A change to not create datapoints for enhancedMetrics was made and committed.

test-services/ecsmeta/app/metadata.json
```diff
@@ -0,0 +1,112 @@
...
+                "com.amazonaws.ecs.cluster": "arn:aws:ecs:us-east-1:906383545488:cluster/seon-fargate-test",
+                "com.amazonaws.ecs.container-name": "~internal~ecs~pause",
+                "com.amazonaws.ecs.task-arn": "arn:aws:ecs:us-east-1:906383545488:task/01f5855c-2335-4f1d-bddb-049aaa418317",
+                "com.amazonaws.ecs.task-definition-family": "seon-fargate-test",
+                "com.amazonaws.ecs.task-definition-version": "3"
+            },
+            "DesiredStatus": "RESOURCES_PROVISIONED",
+            "KnownStatus": "RESOURCES_PROVISIONED",
+            "Limits": {
+                "CPU": 0,
```
keitwb: Send in this metric.

seonsfx: The metric will be sent as `cpu.limit`.

deployments/fargate/agent.yaml
```diff
@@ -0,0 +1,51 @@
...
+logging:
+  level: {"#from": "env:LOG_LEVEL", default: "info"}
+
+# observers are what discover running services in the environment
+observers:
+    # Enable the ecs observer to discover other ECS containers running on the
+    # same ECS task.
+  - type: ecs
+    labelsToDimensions: &labelMap
+      # We map the cluster label to ClusterName to match what our AWS
```
keitwb: This comment is wrong.

docs/observers/ecs.md
```diff
@@ -0,0 +1,153 @@
...
+
+# ecs
+
+ Queries the [ECS Task Metadata Endpoint version 2](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v2.html) for running containers.
+
+The Smart Agent needs to run in the same task with the containers to be
+monitored to get access to the data through ECS task metadata endpoint.
+
+## Configuration from Labels
+You can configure monitors by putting special labels on your Docker
```
keitwb: I think we should be more explicit here and say that you **must** put at least one special label to identify a port that is exposed on the container since ECS metadata doesn't contain any port information.

internal/core/common/ecs/ecs.go
```diff
@@ -0,0 +1,45 @@
...
+	Revision    string      `json:"Revision"`
+	KnownStatus string      `json:"KnownStatus"`
+	Containers  []Container `json:"Containers"`
+}
+
+// GetDimensions returns a set of dimensions based on the task metadata
+func (task *TaskMetadata) GetDimensions() map[string]string {
+	dims := make(map[string]string, 0)
+	if idx := strings.Index(task.ClusterName, "/"); idx >= 0 {
+		dims["ClusterName"] = task.ClusterName[idx+1 : len(task.ClusterName)]
```
keitwb: If `task.ClusterName` is `my-cluster/` what will happen?  I think it will panic since `idx+1` will be out of bounds to the slice.

keitwb: Also, why is the stuff after the first slash undesirable?

keitwb:
Now that the docker container stats monitor only sends in built-in content metrics by default, you can remove the section for it in https://github.com/signalfx/signalfx-agent/blob/master/whitelist.json#L394

internal/core/common/ecs/ecs.go
seonsfx: Every time I saw the cluster name provided by metadata API, task.ClusterName was having a format of `[AWS task ARN]/[Cluster name given by user]` and that was why I was taking whatever comes after the first slash. If the same happens all the time, when user set `my-cluster/` as the cluster name, it will still be parsed as `my-cluster/` after the task ARN part is removed.

seonsfx: I just did another round of research and there wasn't really any documentation that guarantee the task ARN will be a part of the cluster name, so now we should consider when the cluster name is solely what's given by user. I have three options to resolve this.

seonsfx: 1. Just use the whole thing ex) `arn:aws:ecs:us-east-1:906383545488:cluster/seon-fargate-test`
2. Use regex to extract only the resource name  ex) `seon-fargate-test`
3. TaskARN will be given thru another label so replace TaskARN in the cluster name with an empty string

keitwb:
Add `ecs_cluster_arn`.

Research how to construct `ecs_container_arn` to be consistent with what AWS API returns for `containerArn` in https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeTasks.html.

internal/core/common/ecs/ecs.go
keitwb: ok, yeah it's fine that the Task ARN isn't part of the cluster name because we already get that separately.

Keep this logic you have for `ClusterName`, but guard it to avoid panics in case of a malformed ClusterName field.

Then send in the `task.ClusterName` field unaltered as `ecs_cluster_arn`.

seonsfx:
For `container_arn`, just did an extensive research including what's believed to be ECS metadata API v1 and there's completely no clue how to get an arn for a container, except calling `DescribeTasks` through AWS CLI.

seonsfx:
For `docker-container-stats` section in `whitelist.json`, we're still relying on the portion for documentation(specifically where we show if it is custom metric or not). Do we still want to get rid of it?

keitwb:
@barbara-sfx The `agent.signalfx.com` value is just a namespace for Docker labels, not an actual URL for a realm.

deployments/fargate/README.md
```diff
@@ -0,0 +1,75 @@
...
+	- `Number of tasks` is also required for Fargate service configuration
+5. Leave everything else at default and click "Next step"
+6. The second step is configuring network. Fargate requires to run with `awsvpc`
+network type.
+After providing all required network settings, click "Next step"
+7. Leave everything on this next page at their defaults and click "Next step".
+8. Click "Create Service" and you will see new tasks starting up as many as you 
+instructed in Step 4 above. In each task, there will be an agent container that was 
+deployed along with other Fargate containers. As all the containers and the agent 
+startup, you should see infrastructure and docker metrics flowing soon.
```
keitwb: capitalize `Docker`

deployments/fargate/signalfx-agent-fargate-task.json
```diff
@@ -0,0 +1,54 @@
...
+        }
+    ],
+    "cpu": "256",
+    "memory": "512",
```
keitwb: Can probably reduce this by half.

internal/monitors/ecs/metricdoc.go
```diff
@@ -0,0 +1,78 @@
+package ecs
+
+// CUMULATIVE(blkio.io_service_bytes_recursive.async): Volume, in bytes, of asynchronous block I/O
+
+// CUMULATIVE(blkio.io_service_bytes_recursive.read): Volume, in bytes, of reads from block devices
+// CUSTOM(blkio.io_service_bytes_recursive.read) : false
```
keitwb: Remove spaces before the `:`, but leave the space after the `:`.

whitelist.json
```diff
@@ -131,25 +131,6 @@
             "gauge.jenkins.node.vm.memory.total.used"
         ]
     },
-    {
-        "monitorType": "collectd/docker",
```
keitwb: We have to leave this block for `collectd/docker` because it doesn't use any of the common logic that you refactored (it runs in a Python collectd plugin).

The `docker-container-stats` below is good to remove though.

deployments/fargate/agent.yaml
```diff
@@ -0,0 +1,54 @@
...
+logging:
+  level: {"#from": "env:LOG_LEVEL", default: "info"}
+
+# observers are what discover running services in the environment
+observers:
+    # Enable the ecs observer to discover other ECS containers running on the
+    # same ECS task.
+  - type: ecs
+    labelsToDimensions: &labelMap
+      # We map the Fargate docker label to ClusterName to match what our AWS
```
keitwb: This comment is incorrect.

PR #606, opened 2019-02-26
Turn up default writer settings to better accomodate heavy load out o…

keitwb:
…f the box


keitwb:
This is unnecessary with #609.

PR #608, opened 2019-02-26
Chef integration tests

jchengsfx:


tests/deployments/chef/chef_test.py
```diff
@@ -0,0 +1,57 @@
...
+    ("amazonlinux2", INIT_SYSTEMD),
+    ("centos6", INIT_UPSTART),
+    ("centos7", INIT_SYSTEMD),
+]
+
+
+@pytest.mark.parametrize("base_image,init_system", SUPPORTED_DISTROS)
+def test_chef(base_image, init_system):
+    if base_image in ("debian-7-wheezy", "ubuntu1404"):
+        pytest.skip("Wait for fix in debian init.d script to be released")
```
keitwb: Will release this soon.

.circleci/config.yml
```diff
@@ -319,52 +319,63 @@ jobs:
...
+            fi
+            if [ ! -d tests/deployments/$DEPLOYMENT ]; then
+                # no pytest tests to execute for deployment
+                touch ~/.skip
+            fi
+            cd deployments/$DEPLOYMENT
+            make dev-image
+            case "$DEPLOYMENT" in
+            chef)
+                CHEF_DEV="docker run --rm \
```
keitwb: I would swap the ordering here and run the linting stuff before the integration test for Chef.

keitwb: Since if you have lint errors it will require repushing whereas integration test failures might be transient and can be retried.

jchengsfx: This will run the linting stuff first.  `run_pytest` will run at the end of the job if linting is successful.

PR #609, opened 2019-02-27
Use ring buffer for datapoint writer

keitwb:
This should reduce allocations a bit and also avoid the goroutine blow
up that happens when datapoints can't be sent fast enough


keitwb:
Other main thing this provides is a simple way to prioritize new datapoints over old ones in the case of a slow output.  It also allows larger batches to build up in the case where all requests are busy without blocking monitor output or dropping datapoints.

PR #610, opened 2019-02-27
Test upgrade/downgrade with chef cookbook

jchengsfx:


keitwb:
I looked at the goroutine artifact output by the perf_tests and they seemed fine.  Not sure where it got the 501 number from.  We'll need to tune the perf tests to be more predictable but this can merge as is.

PR #614, opened 2019-02-28
K8s test updates

jchengsfx:


Makefile
```diff
@@ -136,17 +136,16 @@ ifdef K8S_SFX_AGENT
...
+		-n auto \
 		--verbose \
 		--exitfirst \
 		--k8s-observers=k8s-api,k8s-kubelet \
 		--html=test_output/k8s_results.html \
 		--self-contained-html \
-		$(agent_image_arg) $(k8s_version_arg) tests || true
-	docker rm -f -v minikube
+		$(agent_image_arg) $(k8s_version_arg) tests && \
+	docker rm -fv minikube
```
keitwb: I would put some kind of `echo` output saying that minikube has been left running since the tests failed or something to at least give the user a bit more feedback.

tests/helpers/kubernetes/minikube.py
```diff
@@ -1,208 +1,326 @@
...
+MINIKUBE_KUBEADM_VERSION = "0.34.1"
+MINIKUBE_IMAGE_TIMEOUT = int(os.environ.get("MINIKUBE_IMAGE_TIMEOUT", 300))
+MINIKUBE_KUBECONFIG_PATH = "/kubeconfig"
+K8S_API_PORT = 8443
+K8S_RELEASE_URL = "https://storage.googleapis.com/kubernetes-release/release/stable.txt"
+K8S_MIN_VERSION = "1.7.0"
+K8S_MIN_KUBEADM_VERSION = "1.11.0"
+
+
+def get_free_port():
```
keitwb: Note that this will actually bind the port and prevent it from being used without actually returning the socket that underlies it.

tests/helpers/kubernetes/minikube.py
```diff
@@ -1,208 +1,326 @@
...
+        try:
+            self.host_client.containers.get(options["name"]).remove(force=True, v=True)
+        except docker.errors.NotFound:
+            pass
+        options.setdefault("privileged", True)
+        options.setdefault(
+            "environment",
+            {"K8S_VERSION": self.k8s_version, "TIMEOUT": str(timeout), "KUBECONFIG_PATH": MINIKUBE_KUBECONFIG_PATH},
+        )
+        if tcp_socket_open("127.0.0.1", self.registry_port):
```
keitwb: I don't understand this conditional.  Shouldn't it be the opposite?

jchengsfx: `tcp_socket_open` returns true if the port is connectable (i.e. in use). So this just assigns a new port for the registry if for some reason the default port is in use.

keitwb: ok, yeah I should have called that function `tcp_socket_bound` in retrospect since "open" is easily confused with "free".

tests/helpers/kubernetes/minikube.py
jchengsfx: Updated to close the socket when returning the port.

Makefile
jchengsfx: Echo `minikube container is left running for debugging purposes` if tests fail; otherwise remove the minikube container.

keitwb:
Let's merge this soon since it improves the k8s test times so much.

PR #616, opened 2019-03-03
Reimplement dimension property handling

keitwb:
This introduces a delay before dimension updates are sent so that
quickly changing dimension property sets are combined into the latest.

This should also make the agent use less memory when property updates
back up since we are not using a separate goroutine per request.


keitwb:
@tedoc2000 this is the thing that we discussed a while back to reduce rapid back-to-back dimension updates by making the agent delay and consolidate them.

internal/core/propfilters/dimpropfilter.go
```diff
@@ -90,15 +90,11 @@ func (f *basicDimPropsFilter) FilterDimProps(dimProps *types.DimProperties) *typ
...
-			Properties: filteredProperties,
-			Tags:       dimProps.Tags,
-		}
+	return &types.DimProperties{
+		Dimension:  dimProps.Dimension,
+		Properties: filteredProperties,
+		Tags:       dimProps.Tags,
 	}
-
-	return nil
```
keitwb: @mstumpfx instead of it "nilling out" the dimension set and becoming a noop in the writer, I'm making it just return a dimension with an empty property list to handle the case where a dimension has one set of properties (not filtered) and then changes to another set of properties that are all filtered, so we still do the update to the backend.  I think there was also a bug before where if all the properties were filtered out, no tags would sync since we don't consider the number of tags.  Let me know if you can think of any other edge cases with this that I might not be handling.

mstumpfx: I guess we should consider a `tagsToExclude` config option if some of our monitors are going to start emitted tags by default.

keitwb: yeah I don't really want to go there yet.

PR #620, opened 2019-03-06
Run for the first time in a goroutine

keitwb:
This will make Configure not block as long


keitwb:
> How about:
> 
> func RunOnInterval(ctx context.Context, fn func(), interval time.Duration) {
> timer := time.NewTicker(interval)
> go func() {
> defer timer.Stop()
> for {
> select {
> case <-ctx.Done(): return
> default: fn(); <-timer.C
> }
> }
> }()
> }
> 
> I'm assuming channel timer.C gets written to at the specified intervals and thus execution at <-timer.C blocks at regular intervals waiting for the channel to get written into.

That would prevent the goroutine from stopping immediately upon the context cancellation and delay the agent shutdown.

PR #621, opened 2019-03-06
Sync deployment name and UID to kubernetes_pod_uid

mstumpfx:
This adds a new internal ReplicaSet cache for the Kubernetes cluster monitor.

Since we cache Pods and their OwnerReferences already (which may be a ReplicaSet), using a ReplicaSet cache we can look up a Pods deployment by looking at the pod's owners owner, aka a deployment.

New properties syncing to `kubernetes_pod_uid`: `deployment` and `deployment_uid`

Change in property syncing to `kubernetes_uid` for deployment metrics: `name` --> `deployment` so that we can filter in the UI pods and deployments using the same property

### DIFF
--- a/internal/monitors/kubernetes/cluster/metrics/cache.go
+++ b/internal/monitors/kubernetes/cluster/metrics/cache.go
@@ -32,6 +32,7 @@ type DatapointCache struct {
 	uidKindCache               map[types.UID]string
 	podCache                   *k8sutil.PodCache
 	serviceCache               *k8sutil.ServiceCache
+	replicaSetCache            *k8sutil.ReplicaSetCache
 	useNodeName                bool
 	nodeConditionTypesToReport []string
 }
@@ -44,6 +45,7 @@ func NewDatapointCache(useNodeName bool, nodeConditionTypesToReport []string) *D
 		uidKindCache:               make(map[types.UID]string),
 		podCache:                   k8sutil.NewPodCache(),
 		serviceCache:               k8sutil.NewServiceCache(),
+		replicaSetCache:            k8sutil.NewReplicaSetCache(),
 		useNodeName:                useNodeName,
 		nodeConditionTypesToReport: nodeConditionTypesToReport,
 	}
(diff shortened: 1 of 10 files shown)


internal/monitors/kubernetes/cluster/metrics/cache.go
```diff
@@ -210,16 +197,45 @@ func (dc *DatapointCache) handleAddService(svc *v1.Service) {
...
+		dimProps := dimPropsForPod(cachedPod, dc.serviceCache, dc.replicaSetCache)
+		if dimProps != nil {
+			dc.addDimPropsToCache(uid, dimProps)
 		}
 	}
 }
 
+// refreshPodsDimProps will call dimPropsForPods for all pods in
+// a given a list of pod UIDs
+func (dc *DatapointCache) refreshPodsDimProps(uids []types.UID) {
```
keitwb: Just expand this loop into the functions that call this. I don't like function names that differ by only a single `s` in the middle of them, it is too hard distinguish.

keitwb: You don't even need the intermediate variable `podUIDs` if you inline it.

tests/monitors/kubernetes_cluster/perf_test.py
```diff
@@ -225,3 +225,127 @@ def missing_service_tags():
...
+                            backend,
+                            dim_name="kubernetes_pod_uid",
+                            dim_value=pod_uid,
+                            prop_name="kubernetes_deployment_uid",
+                            prop_value=rs["dp_uid"],
+                        ):
+                            return False
+                    return True
+
+            assert wait_for(has_all_deployment_props, interval_seconds=2)
```
keitwb: This should also test for the `replicaSet` property.

tests/monitors/kubernetes_cluster/perf_test.py
```diff
@@ -225,3 +225,178 @@ def missing_service_tags():
...
+            time.sleep(30)
+
+            pprof_client.save_goroutines()
+
+            assert (
+                backend.datapoints_by_metric["sfxagent.go_num_goroutine"][-1].value.intValue < 100
+            ), "too many goroutines"
+
+            assert (
+                backend.datapoints_by_metric["sfxagent.go_heap_alloc"][-1].value.intValue < 2 * heap_usage_baseline
```
keitwb: Twice the baseline seems a bit high since technically everything that was put into heap for the monitor should be removed right?

mstumpfx: updated the test to await for heap 1.25x less than baseline

keitwb:
Can you repush this, something got stuck with CircleCI.

PR #627, opened 2019-03-12
Helm tests

jchengsfx:


.circleci/config.yml
```diff
@@ -477,36 +485,43 @@ workflows:
           k8s_version: "v1.13.0"
           requires:
             - build
+            - helm_check
       - k8s_integration_tests:
           name: "k8s_v1.12.0_integration_tests"
           k8s_version: "v1.12.0"
           requires:
             - build
+            - helm_check
```
keitwb: Why do the k8s integration tests require the helm check?

Makefile
```diff
@@ -118,8 +118,8 @@ run-dev-image:
 		-v $(CURDIR)/tmp/pprof:/tmp/pprof \
 		signalfx-agent-dev /bin/bash
 
-MARKERS ?= not packaging and not installer and not k8s and not windows_only and not deployment and not perf_test
 .PHONY: run-integration-tests
+run-integration-tests: MARKERS ?= not packaging and not installer and not k8s and not windows_only and not deployment and not perf_test
```
keitwb: What does this syntax mean?

tests/deployments/helm/helm_test.py
```diff
@@ -0,0 +1,112 @@
...
+            break
+    assert release, "chart '%s' not found in helm list output:\n%s" % (chart_release_name, output)
+    release_name = release.get("Name")
+    assert release_name, "failed to get name for release:\n%s" % yaml.dump(release)
+    return release_name + "-" + chart_name
+
+
+def install_helm_chart(minikube, namespace):
+    minikube.exec_cmd("helm install --namespace=%s --dry-run --debug %s" % (namespace, CONTAINER_CHART_DIR))
+    print(minikube.exec_cmd("helm install --namespace=%s --debug %s" % (namespace, CONTAINER_CHART_DIR)))
```
keitwb: Assign output form the exec_cmd to a var and then print that later if desired, but don't do  the exec_cmd within a print statement.

tests/deployments/helm/helm_test.py
```diff
@@ -0,0 +1,112 @@
...
+            release = rel
+            break
+    assert release, "chart '%s' not found in helm list output:\n%s" % (chart_release_name, output)
+    release_name = release.get("Name")
+    assert release_name, "failed to get name for release:\n%s" % yaml.dump(release)
+    return release_name + "-" + chart_name
+
+
+def install_helm_chart(minikube, namespace):
+    minikube.exec_cmd("helm install --namespace=%s --dry-run --debug %s" % (namespace, CONTAINER_CHART_DIR))
```
keitwb: What is the point of this dry run?  It doesn't even print the output of it.

tests/deployments/helm/nginx-k8s.yaml
```diff
@@ -0,0 +1,52 @@
+---
```
keitwb: Can we put this deployment spec in test-services with the other nginx stuff?  This would be useful to have just for general development to throw up on a k8s cluster manually.

tests/helpers/kubernetes/minikube.py
```diff
@@ -123,17 +125,26 @@ def kubeconfig_exists():
...
+    def exec_kubectl(self, command, namespace=None, check=True):
+        command = "kubectl %s" % command
+        if namespace:
+            command += " -n %s" % namespace
+        return self.exec_cmd(command, check)
+
     def get_cluster_version(self):
         version_yaml = self.exec_kubectl("version --output=yaml")
         assert version_yaml, "failed to get kubectl version"
+        print(version_yaml.strip())
```
keitwb: Still needed?

keitwb:
Looks good except for a few things noted, this will be very helpful to ensure Helm charts at least work out of the box since they are kind of a pain to test manually.

.circleci/config.yml
jchengsfx: Since the helm test would be included in the k8s test jobs, my thought was that the helm chart should be up-to-date and pass lint.

On a related note, do we want to run `helm_check` for every build, or only if `deployments/k8s` has changes?

Makefile
jchengsfx: Since both the `run-integration-tests` and `run-k8s-tests` targets use `MARKERS`, I mistakenly made `MARKERS` a global env var, which would incorrectly set the default markers for `run-k8s-tests` to `not packaging and not installer and not k8s and not windows_only and not deployment and not perf_test`.  This change just makes `MARKERS` target-specific to avoid the conflict.

tests/deployments/helm/nginx-k8s.yaml
jchengsfx: This is actually the same spec from `tests/monitors/nginx`. I can just delete this file and update the test to use the existing spec instead.

tests/deployments/helm/helm_test.py
jchengsfx: I'll update `exec_cmd` to print the output by default.

tests/deployments/helm/helm_test.py
jchengsfx: Will remove.

tests/helpers/kubernetes/minikube.py
jchengsfx: Will remove.

.circleci/config.yml
keitwb: It seems pretty clear to me that there are no external dependencies for the Helm chart outside of `deployments/k8s` so I think it is safe to skip it.

If making the helm_check a requirement of the k8s test jobs means they take longer to start up, I would rather them not be a requirement, but just run along side the k8s jobs, mainly just to keep the finish time on the k8s jobs as fast as possible.

jchengsfx: `helm_check` currently only takes ~15 seconds, so it won't add any time since the k8s tests need to wait ~5 minutes for the build job to complete anyways.

I'll update `helm_check` to only run if `deployments/k8s` has changes.

keitwb: ok cool

PR #630, opened 2019-03-14
update whitelist to include net-io monitor

mstumpfx:
Customer issue, in the next sprint or two I can go fix this the right way.

### DIFF
--- a/whitelist.json
+++ b/whitelist.json
@@ -172,6 +172,16 @@
         ]
     },
     {
+        "monitorType": "net-io",
+        "negated": true,
+        "metricNames": [
+            "if_errors.rx",
+            "if_errors.tx",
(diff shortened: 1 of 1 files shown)


whitelist.json
```diff
@@ -171,6 +171,16 @@
             "if_octets.tx"
         ]
     },
+    {
+        "monitorType": "collectd/net-io",
```
keitwb: No such monitor type.

keitwb:
Can you just add a flag to `internal/monitors/netio`'s config to enable the two additional metrics that aren't built-in and then we can avoid the whitelist altogether.

keitwb:
Are you going to reimplement this with the config flag to conditionalize the generation of the custom metrics?

PR #632, opened 2019-03-16
Collect more node condition metrics for kubernetes-cluster

chaosunsfx:


internal/monitors/kubernetes/cluster/metrics/node.go
```diff
@@ -40,9 +44,20 @@ func datapointsForNode(node *v1.Node, useNodeName bool) []*datapoint.Datapoint {
 		dims["machine_id"] = node.Status.NodeInfo.MachineID
 	}
 
-	return []*datapoint.Datapoint{
-		sfxclient.Gauge("kubernetes.node_ready", dims, nodeConditionValue(node, v1.NodeReady)),
+	datapoints := make([]*datapoint.Datapoint, 0)
+	for _, nodeConditionTypeValue := range nodeConditionTypesToReport {
+		v1NodeConditionTypeValue := v1.NodeConditionType(nodeConditionTypeValue)
+		nodeConditionMetric, ok := nodeConditionTypeToMetric[v1NodeConditionTypeValue]
```
keitwb: I would rather the metric name be constructed programatically (e.g. `fmt.Sprintf("kubernetes.node_%s", strcase.ToSnake(string(v1NodeConditionTypeValue)))`, we already use the `github.com/iancoleman/strcase` package in the agent so the dependency is free) instead so that we will automatically be able to support configuring conditions added in newer K8s versions or custom conditions.

internal/monitors/kubernetes/cluster/clusterstate.go
```diff
@@ -19,20 +19,22 @@ import (
...
+	reflectors                 map[string]*cache.Reflector
+	namespace                  string
+	nodeConditionTypesToReport []string
+	cancel                     func()
 
 	metricCache *metrics.DatapointCache
 }
 
-func newState(clientset *k8s.Clientset, metricCache *metrics.DatapointCache, namespace string) *State {
+func newState(clientset *k8s.Clientset, metricCache *metrics.DatapointCache, namespace string, nodeConditionTypesToReport []string) *State {
```
keitwb: Pass the `nodeConditionTypesToReport` list into the constructor of the DatapointCache at https://github.com/signalfx/signalfx-agent/blob/master/internal/monitors/kubernetes/cluster/monitor.go#L126 (like `useNodeName`) so that you don't have to pass it down through this function.

Issue #629, opened 2019-03-13
Chef breaks service management on centos 7, redhat 7 and amazon 2


mitom:
After installing the agent via chef on redhat 7, centos 7 and amazon 2 the agent is started, but it runs outside of the service manager. It can then be started a 2nd time by using the service manager. The instance launched by chef can not be stopped using the service manager, only by killing it.

```
root@ip-10-0-2-189 ~]# service signalfx-agent status
Not running
[root@ip-10-0-2-189 ~]# ps aux | grep signalfx-agent
signalf+  2581  0.4 21.8 281656 221656 ?       Ssl  13:15   0:00 /usr/bin/signalfx-agent
signalf+  2691  0.7  2.1 784008 22160 ?        Sl   13:15   0:01 /usr/lib/signalfx-agent/lib64/ld-linux-x86-64.so.2 /usr/lib/signalfx-agent/bin/collectd -f -C /var/run/signalfx-agent/collectd/global/collectd.conf
root      3628  0.0  0.0 112640   964 pts/0    S+   13:18   0:00 grep --color=auto signalfx-agent
[root@ip-10-0-2-189 ~]# service signalfx-agent restart
Not running
Starting signalfx-agent
Started.  Logs will go to /var/log/signalfx-agent.log
[root@ip-10-0-2-189 ~]# service signalfx-agent status
Running with pid 3695
[root@ip-10-0-2-189 ~]# ps aux | grep signalfx-agent
signalf+  2581  0.4 21.8 281656 221656 ?       Ssl  13:15   0:00 /usr/bin/signalfx-agent
signalf+  2691  0.7  2.1 784008 22160 ?        Sl   13:15   0:01 /usr/lib/signalfx-agent/lib64/ld-linux-x86-64.so.2 /usr/lib/signalfx-agent/bin/collectd -f -C /var/run/signalfx-agent/collectd/global/collectd.conf
root      3695  0.0  0.2  77064  2132 pts/0    S    13:18   0:00 su -s /bin/sh -c exec "$0" "$@" signalfx-agent -- /usr/bin/signalfx-agent
signalf+  3696  1.3  3.0 145144 31048 ?        Ssl  13:18   0:00 /usr/bin/signalfx-agent
[...]

PR #633, opened 2019-03-17
Service redirect

jchengsfx:
For issue #629


jchengsfx:
> @jchengsfx is the simplest solution to fix this for existing package installations to just delete the `/etc/init.d/signalfx-agent` directory?

I believe so. Users may also need to unregister the service from sysv with `chkconfig --del signalfx-agent` or `update-rc.d -f signalfx-agent remove`.

PR #634, opened 2019-03-18
Use Go modules and upgrade Go to 1.12

keitwb:


docs/development.md
```diff
@@ -60,20 +60,17 @@ run for that commit.
...
 and consistent.
 
-If you add another Go package dependency, you should manually add it to the
-[Gopkg.toml](../Gopkg.toml) and specify a dependency version.  Then run `dep
-ensure` and commit the new vendored source to the repository.  Please make
-commits to the vendor directory separate from commits to the application code
-for easier code review.
+If you add another Go package dependency, you can just run `go get
+<package>@<optional_version>`.  Then run `go mod tidy && go mod vendor` and
+commit the new vendored source to the repository in the same commit that
```
jaysfx: So commit the vendored code along with our own code changes? Not in a separate commit?

keitwb: Yes, Github is pretty good about pushing the vendor stuff down to the bottom and given the concept of a "commit" it should really include everything needed in one atomic unit instead of being spread across multiple units with a hard dependency between them.

Mainly though Github has just gotten smarter about it.

jaysfx: I'd make the argument that the vendored code could live in a separate commit _before_ the code commit that uses the new dependency. That way the project is always in a buildable state. I'd probably prefer it to be separated out but not a huge deal either way for me.

PR #639, opened 2019-03-19
Set api/ingest endpoints and update package versions for chef tests

jchengsfx:


keitwb:
I think Appveyor got stuck queuing the jobs, but it isn't really important for this PR anyway.

PR #640, opened 2019-03-19
Add IAM auth method to Vault remote config

keitwb:


keitwb:
The vendor additions are due to us having to pull in a lot more of Vault and the AWS Golang SDK now.

tests/config_sources/vault_test.py
```diff
@@ -272,3 +272,81 @@ def test_vault_kv_poll_refetch():
...
+
+def test_vault_iam_token_with_explicit_credentials():
+    run_vault_iam_test(
+        {"awsAccessKeyId": "MY_ACCESS_KEY", "awsSecretAccessKey": "MY_SECRET_KEY", "role": "dev-role-iam"}
+    )
+
+
+def test_vault_iam_token_with_env_credentials():
+    old_access_key = os.environ.get("AWS_ACCESS_KEY_ID")
+    os.environ["AWS_ACCESS_KEY_ID"] = "MY_ACCESS_KEY"
```
jaysfx: I haven't tested it but I think this is a little cleaner:

```python
with monkeypatch.context() as m:
   m.setenv("AWS_ACCESS_KEY_ID", "MY_ACCESS_KEY")
   m.setenv("AWS_SECRET_ACCESS_KEY", "MY_SECRET_KEY")

   ...
```

keitwb: Nice, I was looking for something like this and couldn't find it.  Will update.

PR #643, opened 2019-03-20
Support ARM (aarch64/arm64) architectures

jadeblaquiere:
Build process adapted to aarch64 build. rpm/deb packages install and work on Amazon A1 instances under Amazon Linux and Ubuntu 18.


internal/monitors/pyrunner/linux.go
```diff
@@ -21,7 +23,11 @@ func procAttrs() *syscall.SysProcAttr {
 }
 
 func pythonBinaryExecutable() string {
-	return filepath.Join(os.Getenv(constants.BundleDirEnvVar), "lib64", "ld-linux-x86-64.so.2")
+	if strings.Compare(runtime.GOARCH, "arm64") == 0 {
+		return filepath.Join(os.Getenv(constants.BundleDirEnvVar), "lib", "ld-linux-aarch64.so.1")
+	} else {
```
keitwb: From golint:

`if block ends with a return statement, so drop this else and outdent its block`

internal/monitors/pyrunner/linux.go
```diff
@@ -21,7 +23,11 @@ func procAttrs() *syscall.SysProcAttr {
 }
 
 func pythonBinaryExecutable() string {
-	return filepath.Join(os.Getenv(constants.BundleDirEnvVar), "lib64", "ld-linux-x86-64.so.2")
+	if strings.Compare(runtime.GOARCH, "arm64") == 0 {
```
keitwb: The idiomatic way of comparing strings in Go is with just regular `==`, none of that Java confusion.

internal/monitors/collectd/collectd.go
```diff
@@ -468,6 +470,10 @@ func (cm *Manager) rerenderConf(writeHTTPPort int) error {
 
 func (cm *Manager) makeChildCommand() (*exec.Cmd, io.ReadCloser) {
 	loader := filepath.Join(cm.conf.BundleDir, "lib64/ld-linux-x86-64.so.2")
+	if strings.Compare(runtime.GOARCH, "arm64") == 0 {
```
keitwb: Same thing about `==`.

keitwb:
@jchengsfx can you take a look at the K8s failures?  They are consistent but I don't understand what changed in this branch to trigger them.

jchengsfx:
> @jchengsfx can you take a look at the K8s failures? They are consistent but I don't understand what changed in this branch to trigger them.

For some reason, the global env vars that we have set for all jobs in the circleci project settings aren't getting set in the job. Need to investigate further.

internal/monitors/pyrunner/linux.go
jadeblaquiere: done

internal/monitors/pyrunner/linux.go
jadeblaquiere: done

internal/monitors/collectd/collectd.go
jadeblaquiere: done

keitwb:
I do want to make sure AppVeyor passes to make sure this doesn't mess up anything on Windows.

keitwb:
Joe, can you check the git commit user config on your machine and squash and amend all of this code to specify a more identifiable author?  The current commits are using names like 'root' and 'Default EC2 user' and it will be confusing in the git history.

jadeblaquiere:
closing this request, will resubmit from a clean branch

PR #648, opened 2019-03-20
Add ES monitor

asuresh4:


internal/monitors/elasticsearch/client/client.go
```diff
@@ -0,0 +1,80 @@
...
+type esClient struct {
+	host       string
+	port       string
+	scheme     string
+	username   string
+	password   string
+	httpClient *http.Client
+}
+
+type EShttpClient interface {
```
keitwb: `ESHttpClient`

internal/monitors/elasticsearch/client/indexstats.go
```diff
@@ -0,0 +1,176 @@
...
+	Successful *int64 `json:"successful"`
+	Failed     *int64 `json:"failed"`
+}
+
+type IndexStats struct {
+	Primaries IndexStatsGroups `json:"primaries"`
+	Total     IndexStatsGroups `json:"total"`
+}
+
+type IndexStatsGroups struct {
```
keitwb: If you don't want to have to put a comment on all of these, just make them unexported (lowercase first letter).

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+	// populate the plugin_instance dimension
+	Cluster string `yaml:"cluster"`
+	// Version of the elasticsearch cluster
+	Version string `yaml:"version"`
+	// EnableClusterHealth enables reporting on the cluster health
+	EnableClusterHealth *bool `yaml:"enableClusterHealth" default:"true"`
+	// Node stat groups to collect enhanced metrics from. By default the monitor
+	// collects a subset of stats (see documentation for details) across the jvm,
+	// process, threadpool, transport, http indices groups. Only valid options for
+	// this option are transport, http, process, jvm, indices, thread_pool
```
keitwb: Implement `func (c *Config) Validate() error {...}` and return an error if the user provides an invalid option.  That will give them clearer feedback if they make a typo.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+	monitors.Register(monitorType, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Configure monitor
+func (m *Monitor) Configure(conf *Config) error {
+	fmt.Println(conf)
+	esClient := client.NewESClient(conf.Host, conf.Port, conf.Protocol, conf.Username, conf.Password)
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+
+	threadPools := []string{"search", "index"}
```
keitwb: Why not just have this be the default value for `conf.ThreadPools`?  You can do that with the struct tag `default:"[\"search\", \"index\"]"`

keitwb: Also it is impossible for the user to disable these thread pools the way you have it.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+	threadPools = append(threadPools, conf.ThreadPools...)
+
+	utils.RunOnInterval(m.ctx, func() {
+		var nodeStatsOutput client.NodeStatsOutput
+		err := esClient.GetNodeAndThreadPoolStats(&nodeStatsOutput)
```
keitwb: Why not just have this method return `nodeStatsOutput` instead?  Go makes it easy to return two values.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+
+		if err != nil {
+			logger.Errorf("Failed to GET node stats", err)
```
keitwb: Specify the URL it is attempting to fetch for better error diagnosis by the user.

keitwb: Also for logging errors do this: `logger.WithError(err).Error("...")`

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		pluginInstanceDimension, err := prepareClusterDimension(conf.Cluster, nodeStatsOutput.ClusterName)
+
+		if err != nil {
+			logger.Errorf("Failed to create plugin_instance dimension", err)
+			return
+		}
+
+		dps := client.GetNodeStatsDatapoints(&nodeStatsOutput, pluginInstanceDimension, threadPools, conf.NodeStatsGroups)
```
keitwb: Since this can potentially fail, make it return an error back up to here and then log it, and try not to do much logging within the method called.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+	clusterName := userProvidedClusterName
+
+	if clusterName == "" {
+		if queriedClusterName == nil {
+			return nil, errors.New("Failed to get cluster name from node")
+		}
+		clusterName = *queriedClusterName
+	}
+
+	dims["plugin_instance"] = clusterName
```
keitwb: We should probably also provide an alias dimension called `cluster` that is the same value so that it easier to work with than `plugin_instance`, which is super vague.

keitwb: Also put a comment here saying that we send `plugin_instance` for compatibility with built-in content only, in case somebody is tempted to delete it later.

internal/monitors/elasticsearch/client/nodestatsfetcher.go
```diff
@@ -0,0 +1,164 @@
...
+	var out []*datapoint.Datapoint
+	for _, nodeStats := range nodeStatsOutput.NodeStats {
+		out = append(out, GetNodeStatsDatapointsHelper(nodeStats, defaultDims, selectedThreadPools, nodeStatsGroups)...)
+	}
+	return out
+}
+
+func GetNodeStatsDatapointsHelper(nodeStats NodeStats, defaultDims map[string]string, selectedThreadPools []string, nodeStatsGroups []string) []*datapoint.Datapoint {
+	var dps []*datapoint.Datapoint
+	dps = append(dps, nodeStats.Jvm.fetchJVMStats(utils.StringInSlice(jvmStatsGroup, nodeStatsGroups), defaultDims)...)
```
keitwb: Convert `nodeStatsGroups` to a `map[string]bool`, which is like a set.  That way all you have to do is pass `nodeStatsGroupsSet[jvmStatsGroup]` which is O(1) instead of O(n) to the number of groups.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+	Cluster string `yaml:"cluster"`
+	// Version of the elasticsearch cluster
+	Version string `yaml:"version"`
+	// EnableClusterHealth enables reporting on the cluster health
+	EnableClusterHealth *bool `yaml:"enableClusterHealth" default:"true"`
+	// Node stat groups to collect enhanced metrics from. By default the monitor
+	// collects a subset of stats (see documentation for details) across the jvm,
+	// process, threadpool, transport, http indices groups. Only valid options for
+	// this option are transport, http, process, jvm, indices, thread_pool
+	NodeStatsGroups []string `yaml:"enhancedNodeStats" default:"[]"`
```
keitwb: Call it the same thing in Yaml and Go (with different casing obviously).

Also I am not really convinced a list of strings is better than just having a set of bool flags on the monitor (e.g. `enhancedJVMMetrics`, `enhancedProcessMetrics`, etc.).  Bools are easier to work with and you get automatic error validation.  What are your reasons for making it a list?

internal/monitors/elasticsearch/client/nodestatsfetcher.go
```diff
@@ -0,0 +1,164 @@
...
+		prepareGaugeHelper("gauge.process.open_file_descriptors", dims, processStats.OpenFileDescriptors),
+	}...)
+
+	return out
+}
+
+func fetchThreadPoolStats(enhanced bool, threadPools map[string]ThreadPoolStats, defaultDims map[string]string, selectedThreadPools []string) []*datapoint.Datapoint {
+	var out []*datapoint.Datapoint
+	for threadPool, stats := range threadPools {
+		if !utils.StringInSlice(threadPool, selectedThreadPools) {
```
keitwb: Same thing here about using a "set" with a map to bools to make this constant time lookup.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+type Config struct {
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"true"`
+	Host                 string `yaml:"host" validate:"required"`
+	Port                 string `yaml:"port" validate:"required"`
+	// Username used to access elasticsearch stats api
+	Username string `yaml:"username"`
+	// Password used to access elasticsearch stats api
+	Password string `yaml:"password" neverLog:"true"`
+	// Protocol used to connect: http or https
+	Protocol string `yaml:"protocol" default:"http"`
```
keitwb: Just have a bool flag for `UseHTTPS` since there are only two possibilities.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+const monitorType = "elasticsearch"
+
+var logger = log.WithFields(log.Fields{"monitorType": monitorType})
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"true"`
+	Host                 string `yaml:"host" validate:"required"`
+	Port                 string `yaml:"port" validate:"required"`
+	// Username used to access elasticsearch stats api
```
keitwb: API all caps.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+	Username string `yaml:"username"`
+	// Password used to access elasticsearch stats api
+	Password string `yaml:"password" neverLog:"true"`
+	// Protocol used to connect: http or https
+	Protocol string `yaml:"protocol" default:"http"`
+	// Cluster name to which the node belongs. This is an optional config that
+	// will override the cluster name fetched from a node and will be used to
+	// populate the plugin_instance dimension
+	Cluster string `yaml:"cluster"`
+	// Version of the elasticsearch cluster
```
keitwb: In public-facing docs, _Elasticsearch_ should be properly cased.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,119 @@
...
+	timeout time.Duration
+}
+
+func init() {
+	monitors.Register(monitorType, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Configure monitor
+func (m *Monitor) Configure(conf *Config) error {
+	fmt.Println(conf)
```
keitwb: Remove.

internal/monitors/elasticsearch/client/client.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return nil
+}
+
+// Fetches response from the given URL
+func get(url string, username string, password string, httpClient http.Client) ([]byte, error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		fmt.Println(err)
```
keitwb: Remove all three prints in this func.

keitwb:
@jaysfx this is a good one to look at for the stuff you are doing with whitelisting.  It has a lot of groups and most metrics are custom.  We aren't going to block this until your stuff is ready, but should be able to be able to transition to the new unified system with backwards-compatbility so speak up if you see any hindrances to that.  @asuresh4 is writing it in such a way that the defaults to only sending whitelisted metrics with a "standard" config.

jaysfx:
Looks like it should map well to the model we discussed today. @asuresh4 make sure you add a `metadata.yaml` file (see other monitors as an example).

internal/monitors/elasticsearch/client/client.go
asuresh4: Done

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: I realized we had discussed about using separate bool options for each group. Updated to do so.

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: That's a good point. I've updated this.

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Updated so that the method returns `nodeStatsOutput`

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Added URL to the error message and updated the logger

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Added a comment about `plugin_instance` and also added a redundant `cluster` dimension

asuresh4: Added a comment about `plugin_instance` and also added a redundant `cluster` dimension.

internal/monitors/elasticsearch/client/nodestatsfetcher.go
asuresh4: Using `bool` options now. One per group, as mentioned in an earlier comment.

internal/monitors/elasticsearch/client/nodestatsfetcher.go
asuresh4: Updated this so, the input list is converted into map before using it inside `nodestatsfetcher`

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Updated

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Updated

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Done

internal/monitors/elasticsearch/client/client.go
asuresh4: Done

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Corrected existing occurrences. Will do it right as I add a more docs content.

internal/monitors/elasticsearch/client/indexstats.go
```diff
@@ -130,15 +149,30 @@ type QueryCache struct {
 	Evictions         *int64 `json:"evictions"`
 }
 
+// FilterCache stats available across all supported versions
+// of Elasticsearch. Known as QuertyCache after ES 2.0.
+type FilterCache struct {
+	MemorySizeInBytes *int64 `json:"memory_size_in_bytes"`
+	//TotalCount        *int64 `json:"total_count"`
```
keitwb: Why are these commented out?

internal/monitors/elasticsearch/client/client.go
```diff
@@ -35,44 +43,43 @@ func NewESClient(host string, port string, scheme string, username string, passw
...
 // Method to fetch node stats
-func (c *esClient) GetNodeAndThreadPoolStats(result *NodeStatsOutput) error {
+func (c *esClient) GetNodeAndThreadPoolStats() (*NodeStatsOutput, string, error) {
 	url := fmt.Sprintf("%s://%s:%s/%s", c.scheme, c.host, c.port, nodeStatsEndpoint)
 
 	body, err := get(url, c.username, c.password, *c.httpClient)
 
 	if err != nil {
-		return err
+		return nil, url, err
```
keitwb: Instead of returning the URL up the call stack, just wrap the error message so that it includes the URL.  That way, in your [main RunOnInterval function](https://github.com/signalfx/signalfx-agent/pull/648/files/0dd9b359126222ad238e10d8b9ad8dc99fb58f5d..bc625c7fa992f2b96d8460d98e54fd6226fd3185#diff-d839d4d48563fddb8b5872435d15c20bR72), you don't need to actually deal directly with the URL.

So like here it would be like: `return nil, fmt.Errorf("could not get url %s: %v", url, err)`.

Sorry, didn't realize the URL wasn't readily available in that caller.

internal/monitors/elasticsearch/client/indexstats.go
asuresh4: Removed commented lines

internal/monitors/elasticsearch/client/client.go
asuresh4: Moved this to the method that actually fetches the URL

internal/monitors/elasticsearch/client/clusterstatsfetcher.go
```diff
@@ -0,0 +1,61 @@
...
+		prepareGaugeHelper("gauge.cluster.relocating-shards", defaultDims, clusterStatsOutput.RelocatingShards),
+		prepareGaugeHelper("gauge.cluster.unassigned-shards", defaultDims, clusterStatsOutput.UnassignedShards),
+	}...)
+	return out
+}
+
+// Map cluster status to a numeric value
+func getMetricValueFromClusterStatus(s *string) *int64 {
+	out := new(int64)
+	status := *s
```
keitwb: Since this could theoretically be `nil`, add a condition that returns `-1` if it is such, meaning "unknown".

keitwb: It would panic on this line if it were actually `nil`.

internal/monitors/elasticsearch/client/clusterstatsfetcher.go
```diff
@@ -0,0 +1,61 @@
...
+	case clusterStatusGreen:
+		*out = 0
+		break
+	case clusterStatusYellow:
+		*out = 1
+		break
+	case clsuterStatusRed:
+		*out = 2
+		break
+	default:
```
keitwb: default is unnecessary.

internal/monitors/elasticsearch/client/clusterstatsfetcher.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Map cluster status to a numeric value
+func getMetricValueFromClusterStatus(s *string) *int64 {
+	out := new(int64)
+	status := *s
+
+	switch status {
+	case clusterStatusGreen:
+		*out = 0
+		break
```
keitwb: `break` is unnecessary in Go.

asuresh4:
@keitwb - yes, that's for index stats. I'm yet to push changes for that. Right now, I've pushed updates for node and cluster stats.

internal/monitors/elasticsearch/client/clusterstatsfetcher.go
asuresh4: Returning `nil` now. That would mean the metric would stop emitting if there's no known value for it.

docs/monitors/elasticsearch.md
```diff
@@ -0,0 +1,124 @@
+<!--- GENERATED BY gomplate from scripts/docs/monitor-page.md.tmpl --->
+
+# elasticsearch
+
+Monitors ElasticSearch instances.
```
keitwb: This needs to be a bit more descriptive.  Maybe talk about all the subtleties we have learned about master nodes and index stats, etc.

internal/monitors/elasticsearch/client/indexstatsfetcher.go
```diff
@@ -0,0 +1,472 @@
...
+			prepareGaugeHelper(fmt.Sprintf("gauge.indices.%s.suggest.current", aggregationType), defaultDims, suggest.Current),
+			prepareCumulativeHelper(fmt.Sprintf("counter.indices.%s.suggest.time", aggregationType), defaultDims, suggest.TimeInMillis),
+			prepareCumulativeHelper(fmt.Sprintf("counter.indices.%s.suggest.total", aggregationType), defaultDims, suggest.Total),
+		}...)
+	}
+
+	return out
+}
+
+func (percolate *Percolate) fetchIndexPercolateStats(enhanced bool, defaultDims map[string]string, aggregationType string) []*datapoint.Datapoint {
```
keitwb: Rename all these that start with `fetch`.  That strongly implies that the function is doing something externally to generate the metrics, but it is really just mapping the struct to datapoints.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+
+		dps = append(dps, client.GetClusterStatsDatapoints(clusterStatsOutput, defaultDimensions, conf.EnableEnhancedClusterHealthStats)...)
+
+		if !*conf.EnableIndexStats {
+			m.sendDatapoints(dps)
+			return
+		}
+
+		// Collect Index stats
+		if !conf.IndexStatsMasterOnly || nodeIsCurrentMaster {
```
keitwb: How will this ever be reached since you return on line 157 if the node isn't the current master?

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+		// If current node is not the master, collect only node level stats
+		if !nodeIsCurrentMaster {
+			m.sendDatapoints(dps)
+			return
+		}
+
+		// Collect cluster level stats
+		clusterStatsOutput, err := esClient.GetClusterStats()
+
+		if err != nil {
```
keitwb: Send the datapoints as they are returned by these methods called in this method.  Don't wait until the end to send all of them.  In the case of this error, you wouldn't even send the node stats if you couldn't fetch the cluster stats at all since you just return here, but also you would delay the sending of the node stats unnecessarily until you got the cluster stats.

Also, parallelize as many of these requests as possible to reduce the runtime of this method and to try and keep them as consistent as possible on the interval.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func isCurrentMaster(nodeID string, masterInfoOutput *client.MasterInfoOutput) (bool, error) {
+	var isCurrentMaster bool
+
+	masterNode := masterInfoOutput.MasterNode
+
+	if masterNode == nil {
+		logger.Warn("Unable to identify Elasticsearch cluster master node. Assuming current node is not the current master.")
+		return isCurrentMaster, nil
```
keitwb: Make this an explicit `return false, nil` instead of relying on the implicit zero value of the var.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+	return dims, nil
+}
+
+func isCurrentMaster(nodeID string, masterInfoOutput *client.MasterInfoOutput) (bool, error) {
+	var isCurrentMaster bool
+
+	masterNode := masterInfoOutput.MasterNode
+
+	if masterNode == nil {
+		logger.Warn("Unable to identify Elasticsearch cluster master node. Assuming current node is not the current master.")
```
keitwb: Use a throttled logger to log this since it is going to happen every collection cycle if it happens at all: https://github.com/signalfx/signalfx-agent/blob/master/internal/utils/log.go#L142

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+		return isCurrentMaster, nil
+	}
+
+	if nodeID == *masterNode {
```
keitwb: From here down can be condensed to just `return nodeID == *masterNode, nil`

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+	return isCurrentMaster, nil
+}
+
+func prepareNodeMetricsDimensions(nodeStats map[string]client.NodeStats) (map[string]string, error) {
+	var nodeID *string
+
+	// nodes will have exactly one entry, for the current node since the monitor hits "_nodes/_local" endpoint
+	for node := range nodeStats {
+		nodeID = new(string)
+		*nodeID = node
```
keitwb: There is a [util in golib](https://github.com/signalfx/golib/blob/master/pointer/pointer.go#L35) that does this kind of this already: `nodeID = pointer.String(node)`

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	return isCurrentMaster, nil
+}
+
+func prepareNodeMetricsDimensions(nodeStats map[string]client.NodeStats) (map[string]string, error) {
+	var nodeID *string
+
+	// nodes will have exactly one entry, for the current node since the monitor hits "_nodes/_local" endpoint
+	for node := range nodeStats {
```
keitwb: I would be more comfortable with this if you did a check that indeed there is only one key:

```
if len(nodeStats) != 1 {
    return nil, error.New("nodeStats did not have one entry")
```

Or something like that before trying to extract it.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,276 @@
...
+
+	if nodeID == *masterNode {
+		isCurrentMaster = true
+	}
+
+	return isCurrentMaster, nil
+}
+
+func prepareNodeMetricsDimensions(nodeStats map[string]client.NodeStats) (map[string]string, error) {
+	var nodeID *string
```
keitwb: This really doesn't need to be a pointer.  You can just check if it is `== ""` below if you need to check if it didn't come through.  It is safer than dealing with pointer deferencing just in case you missed something, especially when refactoring, that could cause a panic.

internal/monitors/elasticsearch/metadata.yaml
```diff
@@ -0,0 +1,182 @@
+- dimensions: null
+  doc: |
+    Monitors ElasticSearch instances.
+  metrics:
+  - description: The total number of get requests since node startup
```
keitwb: Go ahead and add a `group: ` field to all of these and extend the schema in https://github.com/signalfx/signalfx-agent/blob/master/internal/selfdescribe/monitors.go#L20 to support it.  It will save @jaysfx some work later and will be good to get while everything if fresh on your mind.

internal/monitors/elasticsearch/metricdocs.go
```diff
@@ -0,0 +1,89 @@
+package elasticsearch
+
+// COUNTER(counter.indices.get.total): The total number of get requests since node startup
```
keitwb: Don't need these special comments at all if you have the metadata yaml file.

tests/monitors/collectd_elasticsearch/elasticsearch_test.py
```diff
@@ -0,0 +1,205 @@
...
+            ), "Didn't get gauge.cluster.initializing-shards metric"
+            assert wait_for(
+                p(has_datapoint_with_metric_name, backend, "gauge.thread_pool.threads")
+            ), "Didn't get gauge.thread_pool.threads metric"
+            assert not has_log_message(get_output().lower(), "error"), "error found in agent output!"
+
+
+@pytest.mark.k8s
+@pytest.mark.kubernetes
+def test_elasticsearch_in_k8s(agent_image, minikube, k8s_observer, k8s_test_timeout, k8s_namespace):
```
keitwb: Delete the k8s test, it is unnecessary.

tests/monitors/elasticsearch/elasticsearch_test.py
```diff
@@ -144,39 +139,6 @@ def test_elasticsearch_with_threadpool():
...
-            ), "Didn't get gauge.cluster.initializing-shards metric"
-            assert wait_for(
-                p(has_datapoint_with_metric_name, backend, "gauge.thread_pool.threads")
-            ), "Didn't get gauge.thread_pool.threads metric"
-            assert not has_log_message(get_output().lower(), "error"), "error found in agent output!"
-
-
 @pytest.mark.k8s
 @pytest.mark.kubernetes
 def test_elasticsearch_in_k8s(agent_image, minikube, k8s_observer, k8s_test_timeout, k8s_namespace):
```
keitwb: Delete this k8s test.

internal/monitors/elasticsearch/metadata.yaml
keitwb: Also, is this list really exhaustive?  It seems too short given how much code there was.

internal/monitors/elasticsearch/metricdocs.go
asuresh4: Removed

internal/monitors/elasticsearch/metadata.yaml
asuresh4: This is not exhaustive. I made use of the file from the existing monitor. Looks like this is only a fraction of all the metrics. Will update

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Made it a string

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Added a check to confirm there's only 1 entry in the map

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Made this a string instead of a pointer

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Done

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Updated

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Sending datapoints as soon as they are collected now. I've updated so that fetching cluster stats and index stats can be done in parallel once the information required by both the groups of stats are collected. Please take a look and let me know what you think.

docs/monitors/elasticsearch.md
asuresh4: Added more details and examples

docs/monitors/elasticsearch.md
```diff
@@ -35,7 +35,7 @@ Monitor Type: `elasticsearch`
 | `enableEnhancedClusterHealthStats` | no | `bool` | Enable Cluster level stats. These stats report only from master Elasticserach nodes (**default:** `false`) |
 | `indexStatsIntervalSeconds` | no | `integer` | Interval to report IndexStats on (defaults to 60s) (**default:** `60`) |
 | `indexSummaryOnly` | no | `bool` | Collect only aggregated index stats across all indexes (**default:** `false`) |
-| `indexStatsMasterOnly` | no | `bool` | Collect index stats only from Master node. By default the monitor will collect these from master eligible nodes. (**default:** `false`) |
+| `indexStatsMasterOnly` | no | `bool` | Collect index stats only from Master node (defaults to true) (**default:** `true`) |
```
keitwb: You don't need to say `(defaults to true)`, the doc generator will add it for you.

asuresh4:
@jaysfx - i've added metric docs in metadata.yaml and added a comment `# group: <GROUP_NAME>` for each metric

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Resolving this since we decided to keep it the way it is, as the method only parses the output struct.

internal/monitors/elasticsearch/client/client.go
```diff
@@ -0,0 +1,181 @@
...
+		scheme:     scheme,
+		username:   username,
+		password:   password,
+		httpClient: httpClient,
+	}
+}
+
+// Method to collect index stats
+func (c *esClient) GetIndexStats() (*IndexStatsOutput, error) {
+	url := fmt.Sprintf("%s://%s:%s/%s", c.scheme, c.host, c.port, allIndexStatsEndpoint)
```
keitwb: This is a lot of repeated code.  You can just have a method that decodes the json to a target object like this:

```
func (c *esClient) fetchJSON(url string, obj interface{}) error {
	req, err := http.NewRequest("GET", url, nil)
	if err != nil {
		return nil, fmt.Errorf("could not get url %s: %v", url, err)
	}

	req.SetBasicAuth(username, password)
	res, err := httpClient.Do(req)

	if err != nil {
		return nil, fmt.Errorf("could not get url %s: %v", url, err)
	}
	defer res.Body.Close()

	if res.StatusCode != 200 {
		return nil, fmt.Errorf("Received status code that's not 200: %s, url: %s", string(body), url)
	}

       err := json.NewDecoder(res.Body).Decode(obj)
	if err != nil {
		return nil, fmt.Errorf("could not get url %s: %v", url, err)
	}

	return nil
}
```

And then call it like this:

```
func (c *esClient) GetIndexStats() (*IndexStatsOutput, error) {
     	url := fmt.Sprintf("%s://%s:%s/%s", c.scheme, c.host, c.port, allIndexStatsEndpoint)
        var indexStatsOutput IndexStatsOutput
        return &indexStatsOutput, c.fetchJSON(url, &indexStatsOutput)
}
```

That would eliminate most of the boilerplate and also avoid having to copy the response body to a temporary variable as well, which would help reduce GC churn possibly.

internal/monitors/elasticsearch/client/indexstatsfetcher.go
```diff
@@ -0,0 +1,482 @@
...
+		out = append(out, indexStats.Primaries.Recovery.getIndexRecoveryStats(enhancedStatsForIndexGroupsOption[RecoveryStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStats.Primaries.IDCache.getIndexIDCacheStats(enhancedStatsForIndexGroupsOption[IDCacheStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStats.Primaries.Suggest.getIndexSuggestStats(enhancedStatsForIndexGroupsOption[SuggestStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStats.Primaries.Percolate.getIndexPercolateStats(enhancedStatsForIndexGroupsOption[PercolateStatsGroup], defaultDimsForPrimaries)...)
+	}
+
+	defaultDimsForTotal := utils.MergeStringMaps(defaultDims, map[string]string{
+		"aggregation": Total,
+	})
+	out = append(out, indexStats.Total.Docs.getIndexDocsStats(enhancedStatsForIndexGroupsOption[DocsStatsGroup], defaultDimsForTotal)...)
```
keitwb: `indexStats.Total` and `indexStats.Primaries` are the same struc type so you can just assign one or the other to a variable and then call `.Docs`, `.Store`, etc. on that with the appropriate dimension set.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,414 @@
...
+//     port: 9200
+//     enableEnhancedHTTPStats: true
+//     enableEnhancedJVMStats: true
+//     enableEnhancedProcessStats: true
+//     enableEnhancedThreadPoolStats: true
+//     enableEnhancedTransportStats: true
+//     enableEnhancedNodeIndicesStats:
+//    	- indexing
+//    	- warmer
+//     - get
```
keitwb: Indentation is off.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,414 @@
...
+func init() {
+	monitors.Register(monitorType, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+type sharedInfo struct {
+	isInitialized        bool
+	nodeIsCurrentMaster  bool
+	defaultDimensions    map[string]string
+	nodeMetricDimensions map[string]string
+	lock                 *sync.RWMutex
```
keitwb: Don't use pointer for mutex field.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,414 @@
...
+		// For any reason the monitor is not able to fetch cluster and node information upfront we capture that
+		// to ensure that the stats are not sent in with faulty dimensions. The monitor will try to fetch this
+		// information again every MetadataRefreshInterval seconds
+		if err != nil {
+			logger.WithError(err).Errorf(fmt.Sprintf("Failed to get Cluster and node metadata upfront. Will try again in %d s.", conf.MetadataRefreshIntervalSeconds))
+			return
+		}
+
+		// The first time fetchNodeAndClusterMetadata succeeded the monitor can schedule fetchers for all the other stats
+		if !shared.isInitialized {
```
keitwb: just put `isInitialized` as a variable in the main body of the `Configure` method instead of making it a member of `shared`.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,414 @@
...
+				}
+			}, time.Duration(conf.IntervalSeconds)*time.Second)
+
+			// Collect Index stats
+			utils.RunOnInterval(m.ctx, func() {
+				defaultDimensions, _, nodeIsCurrentMaster := shared.getAllSharedInfo()
+
+				// if "IndexStatsMasterOnly" is true collect stats only from master,
+				// otherwise collect index stats from all nodes
+				if *conf.EnableIndexStats && (!*conf.IndexStatsMasterOnly || nodeIsCurrentMaster) {
```
keitwb: Don't even schedule this function if `*conf.EnableIndexStats` is `false`.  Config is immutable so you don't have to worry about it changing.

tests/monitors/elasticsearch/elasticsearch_test.py
```diff
@@ -144,62 +127,118 @@ def test_elasticsearch_with_threadpool():
             assert not has_log_message(get_output().lower(), "error"), "error found in agent output!"
 
 
+DEFAULT_METRICS = [
```
keitwb: Derive this from the metadata.yaml based on the `included` flag.  See https://github.com/signalfx/signalfx-agent/blob/master/tests/helpers/util.py#L281

asuresh4: Done. Added a method in util to return a list of metric names. Did the same for monitor dimensions

internal/monitors/elasticsearch/client/client.go
asuresh4: That's a good point. I've used this pattern now.

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Updated

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Done

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Done

internal/monitors/elasticsearch/elasticsearch.go
asuresh4: Updated so that we only check this once in configure

internal/monitors/elasticsearch/client/indexstatsfetcher.go
asuresh4: Assigning to a variable now

internal/monitors/elasticsearch/client/indexstatsfetcher.go
```diff
@@ -87,54 +87,56 @@ func getIndexStatsHelper(indexStats IndexStats, defaultDims map[string]string, e
...
+		out = append(out, indexStatsGroup.Segments.getIndexSegmentsStats(enhancedStatsForIndexGroupsOption[SegmentsStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStatsGroup.Translog.getIndexTranslogStats(enhancedStatsForIndexGroupsOption[TranslogStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStatsGroup.RequestCache.getIndexRequestCacheStats(enhancedStatsForIndexGroupsOption[RequestCacheStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStatsGroup.Recovery.getIndexRecoveryStats(enhancedStatsForIndexGroupsOption[RecoveryStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStatsGroup.IDCache.getIndexIDCacheStats(enhancedStatsForIndexGroupsOption[IDCacheStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStatsGroup.Suggest.getIndexSuggestStats(enhancedStatsForIndexGroupsOption[SuggestStatsGroup], defaultDimsForPrimaries)...)
+		out = append(out, indexStatsGroup.Percolate.getIndexPercolateStats(enhancedStatsForIndexGroupsOption[PercolateStatsGroup], defaultDimsForPrimaries)...)
 	}
 
+	indexStatsGroup := indexStats.Total
```
keitwb: You can just assign `indexStatsGroup` to either the total or primaries struct in a conditional the top of this method , along with the dimensions, and then avoid all the duplication of the `out = append(out ...` lines.

asuresh4: Moved the block to a separate method.

PR #649, opened 2019-03-21
Enable sendTraceHostCorrelationMetrics by default

keitwb:
There isn't much reason to disable this if doing tracing now that it is pretty well established that we are going to use it on the backend.


keitwb:
> LGTM for a quick fix. Any thoughts/comments on why this config is on `writer` and not the `trace-forwarder` monitor?

Span processing and host metric generation happens in the writer so allow for the possibility of other sources of trace spans in the agent (see JIRA).

PR #654, opened 2019-03-22
Increase number of goroutines in perf test to 200

mstumpfx:
This should make the test pass more consistently. It hovers around 120 to 160 for this test.


keitwb:
I incorporated this in #658.

PR #655, opened 2019-03-22
Windows sign and push

jchengsfx:
The Windows release process is temporary until we finalize the plan and environments for automated build/release.


scripts/windows/make.ps1
```diff
@@ -55,6 +56,11 @@ function bundle (
         signalfx-agent -AGENT_VERSION "$AGENT_VERSION" -AGENT_BIN "$buildDir\$AGENT_NAME\bin\signalfx-agent.exe"
     }
 
+    if ((Test-Path -Path "$scriptDir\SFX_Windows.pfx") -And $PFX_PASSWORD -ne "") {
```
keitwb: Make the user provide the path to the pfx file instead of forcing them to put it in a specific location, esp one that has a risk of being committed to a public repo.

Then, if the user has provided the path to a cert, throw an error if there is no password provided instead of just silently skipping signing.

PR #656, opened 2019-03-23
Support arm64/aarch64 architectures

jadeblaquiere:
Redo previous PR #643

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,11 @@ RUN apt update &&\
     apt install -y curl wget pkg-config parallel
 
 ARG GO_VERSION
+ARG TARGET_ARCH
+
 ENV PATH=$PATH:/usr/local/go/bin
 RUN cd /tmp &&\
-    wget https://storage.googleapis.com/golang/go${GO_VERSION}.linux-amd64.tar.gz &&\
+    wget https://storage.googleapis.com/golang/go${GO_VERSION}.linux-${TARGET_ARCH}.tar.gz &&\
 	tar -C /usr/local -xf go*.tar.gz
 
 ENV GOPATH=/go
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -369,15 +384,15 @@ RUN curl -fsSL get.docker.com -o /tmp/get-docker.sh &&\
...
+    if [ `uname -m` != "aarch64" ]; then go get github.com/derekparker/delve/cmd/dlv; fi &&\
     go get github.com/tebeka/go2xunit
 
 # Get integration test deps in here
 COPY python/setup.py /tmp/
 RUN pip3 install -e /tmp/
 COPY tests/requirements.txt /tmp/
 RUN pip3 install --upgrade pip==9.0.1 && pip3 install -r /tmp/requirements.txt
-RUN wget -O /usr/bin/gomplate https://github.com/hairyhenderson/gomplate/releases/download/v2.3.0/gomplate_linux-amd64-slim &&\
+RUN wget -O /usr/bin/gomplate https://github.com/hairyhenderson/gomplate/releases/download/2.8.0/gomplate_linux-${TARGET_ARCH}-slim &&\
```
keitwb: If you change the version of gomplate here you need to change the version in https://github.com/signalfx/signalfx-agent/blob/master/.circleci/config.yml#L200 so that the docs check in CI is consistent with what developers are using.

jadeblaquiere: Updated to match.

keitwb:
Ok, that is what I was afraid of, gomplate 2.8.0 doesn't behave exactly the same and caused docs to change a lot in a way that kind of broke our doc gen script.  Does the original gomplate version not support ARM or something?

PR #659, opened 2019-03-25
Rework the status output to be more compact and useful

keitwb:
- There are several sub-sections that can be selected now, the default
 section being a short summary of the agent that fits on a single page.

 - Also adding a sampler that tracks activity of datapoints, events, and
 spans sent over the previous one minute, sampled every ten seconds.
 This is used by the status output to give a rough idea of the volume
 pushed currently through the agent.

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -50,12 +50,30 @@ func setCollectdVersionEnvvar() {
 func doStatus() {
 	set := flag.NewFlagSet("status", flag.ExitOnError)
 	configPath := set.String("config", defaultConfigPath, "agent config path")
+	set.Usage = func() {
+		fmt.Fprintf(set.Output(), "Usage: signalfx-agent status [all | monitors | config | endpoints]\n\n"+
+			"  The optional section arg can be one of the following:\n"+
+			"    all - Dump everything available\n"+
+			"    monitors - Show information about all active monitors\n"+
+			"    config - Show the fully resolved configuration currently in use by the agent\n"+
+			"    endpoints - Show the discovered endpoints available to discovery rules\n"+
+			"  If no section arg is provided, a short summary of the agent is output\n\n")
+		set.PrintDefaults()
+	}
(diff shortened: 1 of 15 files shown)


molner:
@keitwb - what are your plans on making folks (internally and externally) aware of this new capability?

keitwb:
> @keitwb - what are your plans on making folks (internally and externally) aware of this new capability?

The changes are fairly well documented in the status output itself (i.e. it shows the commands to run to get the various subsections), so hopefully it is self-teaching.   The main thing is that support needs to ask for `signalfx-agent status all` for support tickets instead of the previous `signalfx-agent status`, _if_ the customer is using a newer version of the agent.

keitwb:
[This](https://github.com/signalfx/signalfx-agent/pull/659/files#diff-450c3da688feeaa4f8c8e65cf5eb6c41R42) is what the user will see when they run the old plain `status` command.

cmd/agent/main.go
```diff
@@ -50,12 +50,30 @@ func setCollectdVersionEnvvar() {
...
 	set := flag.NewFlagSet("status", flag.ExitOnError)
 	configPath := set.String("config", defaultConfigPath, "agent config path")
+	set.Usage = func() {
+		fmt.Fprintf(set.Output(), "Usage: signalfx-agent status [all | monitors | config | endpoints]\n\n"+
+			"  The optional section arg can be one of the following:\n"+
+			"    all - Dump everything available\n"+
+			"    monitors - Show information about all active monitors\n"+
+			"    config - Show the fully resolved configuration currently in use by the agent\n"+
+			"    endpoints - Show the discovered endpoints available to discovery rules\n"+
+			"  If no section arg is provided, a short summary of the agent is output\n\n")
```
jaysfx: Could you use a raw string here instead to make it a little easier to read?

keitwb: Changed.

PR #661, opened 2019-03-25
Remove debian 7 (wheezy) from package tests since it is EOL

keitwb:


jchengsfx:
Need to also remove wheezy from tests/deployments/chef.

PR #663, opened 2019-03-25
Change custom to included in docs

jaysfx:
Submitting this separately to keep churn in other PRs down.

* Change metrics table to check metrics that are included by default vs. being ones that are not included
* Use a fancy unicode checkmark
* Remove some unneeded whitelist.json processing


mstumpfx:
cc @barbara-sfx

mstumpfx:
approved but barbara makes a good point, we should make sure we can use that unicode character in product docs

jaysfx:
@barbara-sfx Yep, I built the docs in the product-docs repo and it renders correctly

jaysfx:
Here's what it looks like:

Screen Shot 2019-03-25 at 6 42 51 PM

PR #664, opened 2019-03-25
Expvar

bjsignalfx:


docs/monitors/expvar.md
```diff
@@ -0,0 +1,206 @@
+<!--- GENERATED BY gomplate from scripts/docs/monitor-page.md.tmpl --->
+
+# expvar
+
+This monitor exports metrics derived from [expvar](https://golang.org/pkg/expvar/)
+variables in JSON objects retrieved from a HTTP endpoint.
```
keitwb: _an_ HTTP endpoint.

docs/monitors/expvar.md
```diff
@@ -0,0 +1,206 @@
...
+
+The monitor uses configured JSON paths to get metric and dimension values from the fetched JSON objects.
+[Memstats metrics](https://github.com/signalfx/integrations/blob/master/expvar/docs/expvar_metrics.md)
```
keitwb: Do not link to the integrations repo.  This repo and this document are the authoritative docs for this and we will be migrating this to the integrations repo very soon.

docs/monitors/expvar.md
```diff
@@ -0,0 +1,206 @@
...
+
+Below is a sample YAML configuration showing the configuration of custom metrics `xyz` and `h.i`. The `name`s
+of metrics and dimensionals are optional. The `JSONPath` is used for `name` instead if not provided.
+```
+monitors:
+- type: expvar
+  host: 172.12.0.6
+  path: /debug/vars
+  port: 8123
+    metrics:
```
keitwb: Indentation is off.

docs/monitors/expvar.md
```diff
@@ -0,0 +1,206 @@
...
+        dimensions:
+          - name: kl
+            value: mno
+  extraDimensions:
+    metric_source: expvar
+```
+
+Below is an example JSON object that the above configuration may apply to.
+```
+{
```
keitwb: Show a realistic expvar output with the standard fields instead of this highly abstract JSON document.

docs/monitors/expvar.md
```diff
@@ -0,0 +1,206 @@
...
+ "j" : 0.8
+}
+```
+
+Two data points will be created for the metric `a.b.c.d`. The first data point will have the value 4 and
+dimensions `a.b = c`, `a.b.c = 0` and `metric_source = expvar`. The second data point will have the value 9
+and dimensions `a.b = c`, `a.b.c = 1` and `metric_source = expvar`. The monitor creates dimensions such as
+`a.b.c` automatically to store array indexes.
+
+For metric `h.i`, a data point of value 1.2, dimensions `h = 0` and `metric_source = expvar` will be created.
```
keitwb: A dimension`h = 0`?  Why?

docs/monitors/expvar.md
```diff
@@ -0,0 +1,206 @@
...
+ "h": [
+   {
+     "i": 1.2
+   }
+ ],
+ "j" : 0.8
+}
+```
+
+Two data points will be created for the metric `a.b.c.d`. The first data point will have the value 4 and
```
keitwb: Why don't you make two datapoints named `a_b_c_d` and avoid the whole dimension thing?

keitwb: Do you know use cases where the index in the array actually matters in monitoring?

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	UseHTTPS bool `yaml:"useHTTPS"`
+	// If useHTTPS is true and this option is also true, the host's TLS cert will not be verified.
+	SkipVerify bool `yaml:"skipVerify"`
+	// Path to the expvar endpoint, usually `/debug/vars` (the default).
+	Path string `yaml:"path" default:"/debug/vars"`
+	// Excludes metrics memstats.alloc, memstats.by_size.size, memstats.by_size.mallocs and memstats.by_size.frees if false
+	EnhancedMetrics bool `yaml:"enhancedMetrics"`
+	// Metrics configurations
+	MetricConfigs []*MetricConfig `yaml:"metrics"`
+	url           *url2.URL
```
keitwb: Don't use unexported fields in the Config, config should be treated as immutable.  If you want to cache derived values, just set it on the Monitor instance instead.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+type MetricConfig struct {
+	// Metric name
+	Name string `yaml:"name"`
+	// JSON path of the metric value
+	JSONPath string `yaml:"JSONPath" validate:"required"`
+	// SignalFx metric type. Possible values are "gauge" or "counter"
+	Type string `yaml:"type" validate:"required"`
+	// Metric dimensions
+	DimensionConfigs []*DimensionConfig `yaml:"dimensions"`
+	metricType       datapoint.MetricType
```
keitwb: Same thing.  Just derive this as needed in the monitor, it isn't that expensive.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+const monitorType = "expvar"
+
+const (
+	gauge                                  = "gauge"
+	counter                                = "counter"
```
keitwb: I would change this to "cumulative" to match our app nomenclature.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	Host string `yaml:"host" validate:"required"`
+	// Port of the expvar endpoint
+	Port uint16 `yaml:"port" validate:"required"`
+	// If true, the agent will connect to the host using HTTPS instead of plain HTTP.
+	UseHTTPS bool `yaml:"useHTTPS"`
+	// If useHTTPS is true and this option is also true, the host's TLS cert will not be verified.
+	SkipVerify bool `yaml:"skipVerify"`
+	// Path to the expvar endpoint, usually `/debug/vars` (the default).
+	Path string `yaml:"path" default:"/debug/vars"`
+	// Excludes metrics memstats.alloc, memstats.by_size.size, memstats.by_size.mallocs and memstats.by_size.frees if false
```
keitwb: Phrase this, "If true, sends metrics ..."

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	"fmt"
+	"github.com/signalfx/golib/datapoint"
+	"github.com/signalfx/signalfx-agent/internal/core/config"
+	"github.com/signalfx/signalfx-agent/internal/monitors"
+	"github.com/signalfx/signalfx-agent/internal/monitors/types"
+	"github.com/signalfx/signalfx-agent/internal/utils"
+	log "github.com/sirupsen/logrus"
+	"io/ioutil"
+	"net/http"
+	url2 "net/url"
```
keitwb: `neturl "net/url"` would be a bit clearer.

keitwb: Also why do you even need to alias it in the first place?

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	dp.Timestamp = *now
+	dp.Dimensions["application_host"] = conf.Host
+	dp.Dimensions["application_expvar_port"] = strconv.FormatInt(int64(conf.Port), 10)
+	dp.Dimensions["plugin"] = monitorType
+	m.Output.SendDatapoint(dp)
+	return nil
+}
+
+func (m *Monitor) setContext() {
+	m.ctx, m.cancel = context.WithCancel(context.Background())
```
keitwb: Do this directly in `Configure`.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func (m *Monitor) setClient() {
+	m.client = &http.Client{
```
keitwb: Do this directly in `Configure`.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+		Transport: &http.Transport{
+			MaxIdleConnsPerHost: 100,
```
keitwb: You don't need anything remotely close to 100 idle connections when you are only doing one connection every ~10 seconds or so.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+	}
+	// Renaming auto created dimension 'memstats.BySize' to 'class'
+	if metricPath == memstatsBySizeSizeMetricPath || metricPath == memstatsBySizeMallocsMetricPath || metricPath == memstatsBySizeFreesMetricPath {
+		dp.Dimensions["class"] = dp.Dimensions[memstatsBySizeDimensionPath]
+		delete(dp.Dimensions, memstatsBySizeDimensionPath)
+	}
+	dp.Timestamp = *now
+	dp.Dimensions["application_host"] = conf.Host
+	dp.Dimensions["application_expvar_port"] = strconv.FormatInt(int64(conf.Port), 10)
```
keitwb: Is this really useful for charts and analytics?  If people really wanted to distinguish ports, they could do it via `extraDimensions`, either directly in the config or with container labels.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	dp.Dimensions["plugin"] = monitorType
```
keitwb: Don't set this, it is only used for legacy collectd integrations.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	if _map, ok := v.(map[string]interface{}); ok {
+		for _, dConf := range mConf.DimensionConfigs {
+			if len(dConf.splits) != 0 && len(dConf.splits) == i {
+				dp.Dimensions[dConf.Name] = mConf.keys[i]
+			}
+		}
+		setDatapoints(_map[mConf.keys[i+1]], mConf, dp, dpsMap, i+1)
+	} else if arr, ok := v.([]interface{}); ok {
+		_dp := dp
+		for _i, _v := range arr {
```
keitwb: Don't prefix vars with `_`.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+func getApplicationName(_map map[string]interface{}) (string, error) {
+	if cmdline, ok := _map["cmdline"].([]interface{}); ok && len(cmdline) > 0 {
+		if applicationName := strings.TrimSpace(slashLastRegexp.FindStringSubmatch(cmdline[0].(string))[0]); applicationName != "" {
+			return applicationName, nil
+		}
+	}
+	return "", fmt.Errorf("failed to get application name from the first array value of cmdline in map: %+v", _map)
+}
+
+func copyDims(dp *datapoint.Datapoint) map[string]string {
```
keitwb: This is already implemented in `utils.CloneStringMap`.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+	dimensions := map[string]string{}
+	for k, v := range dp.Dimensions {
+		dimensions[k] = v
+	}
+	return dimensions
+}
+
+var camelRegexp = regexp.MustCompile("(^[^A-Z]*|[A-Z]*)([A-Z][^A-Z]+|$)")
+
+func toSnakeCase(s string) string {
```
keitwb: This is already implemented in a library that the agent already depends on: `strcase.ToSnake`.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+			return err
+		}
+	}
+	// Renaming auto created dimension 'memstats.BySize' to 'class'
+	if metricPath == memstatsBySizeSizeMetricPath || metricPath == memstatsBySizeMallocsMetricPath || metricPath == memstatsBySizeFreesMetricPath {
+		dp.Dimensions["class"] = dp.Dimensions[memstatsBySizeDimensionPath]
+		delete(dp.Dimensions, memstatsBySizeDimensionPath)
+	}
+	dp.Timestamp = *now
+	dp.Dimensions["application_host"] = conf.Host
```
keitwb: This is not a great dimension because it will differ depending on where you monitor the app from.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+			}
+			return "http"
+		}(),
+		Host: conf.Host + ":" + fmt.Sprint(conf.Port),
+		Path: conf.Path,
+	}
+}
+
+func (conf *Config) setRunInterval() {
+	conf.runInterval = time.Duration(conf.IntervalSeconds) * time.Second
```
keitwb: Do this directly in Configure.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+					return fmt.Errorf("invalid dimension path %s. Dimension path not parent path of metric path %s", dConf.JSONPath, mConf.JSONPath)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func (conf *Config) setURL() {
+	conf.url = &url2.URL{
```
keitwb: DO this directly in Configure.

internal/monitors/expvar/expvar.go
```diff
@@ -0,0 +1,393 @@
...
+
+func init() {
+	monitors.Register(monitorType, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Validate validates configuration
+func (conf *Config) Validate() error {
+	if conf.MetricConfigs != nil {
+		for _, mConf := range conf.MetricConfigs {
+			_type := strings.TrimSpace(strings.ToLower(mConf.Type))
```
keitwb: Don't use underscore-prefixed var names.

docs/monitors/expvar.md
```diff
@@ -0,0 +1,206 @@
...
+Below is a sample YAML configuration showing the configuration of custom metrics `xyz` and `h.i`. The `name`s
+of metrics and dimensionals are optional. The `JSONPath` is used for `name` instead if not provided.
+```
+monitors:
+- type: expvar
+  host: 172.12.0.6
+  path: /debug/vars
+  port: 8123
+    metrics:
+      - name: xyz
```
keitwb: Show an example of looking up a specific index in an array.

keitwb:
Also, rebase with latest master.

docs/monitors/expvar.md
bjsignalfx: Avoiding dimension wouldn't work for something like PauseNs in https://golang.org/pkg/runtime/.

internal/monitors/expvar/expvar.go
bjsignalfx: 10?

keitwb: If you are only doing one request per cycle, just 1 will do.

keitwb: But I would just omit this option

PR #673, opened 2019-03-27
Add 'agent_version' attribute to Chef cookbook

keitwb:
This is used to derive package_version

Fixes #671.

PR #676, opened 2019-03-28
Fix/chef windows version

mitom:
#673 introduces a bug where `package_version` will be used before it's set for windows. It also introduces a bug for wrapping this cookbook as it's using `default` instead of `node` to consume `agent_version` for windows

cc @keitwb @jchengsfx


keitwb:
Thanks, will get this merged and released as soon as it passes CI.

PR #677, opened 2019-03-29
Salt-Formula: split into sub-states

cetanu:
This pull-request leaves the default behavior of the salt-formula unchanged.

This allows the requisite packages for the signalfx-agent to be installed by applying the `signalfx-agent.install` state.

Our use-case for this is to be able to install the packages on an AWS AMI, and to configure and run the service later on when machines are deployed via Cloudformation (which they receive their SignalFX API key from)


keitwb:
Thanks for the PR @cetanu.  It looks good to me.

PR #678, opened 2019-03-29
Python monitor runner

keitwb:
This is a simpler and more native alternative to collectd/python monitor


keitwb:
Still need to add tests but getting it up for review now.

PR #679, opened 2019-03-29
Generate go datastructures from metadata.yaml

jaysfx:
* Generates `generatedMetadata.go` for each `metadata.yaml` with sets
  for metrics, included metrics, and groups.
* Migrate some existing prometheus monitors to the new generated code.
* Change `metadata.yaml` so that top level is a map so that
  package-level configuration can go there with monitors moved to the
  key `monitors`


cmd/monitorCodeGen/main.go
```diff
@@ -0,0 +1,195 @@
...
+	"fmt"
+	"github.com/signalfx/signalfx-agent/internal/monitors"
+	log "github.com/sirupsen/logrus"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"text/template"
+)
+
+var tmpl = `// DO NOT EDIT. This file is auto-generated.
```
keitwb: See https://golang.org/s/generatedcode.

Makefile
```diff
@@ -1,12 +1,21 @@
...
 
 .PHONY: compileDeps
-compileDeps: templates internal/core/common/constants/versions.go
+compileDeps: templates code-gen internal/core/common/constants/versions.go
+
+.PHONY: code-gen
+code-gen: $(MONITOR_CODE_GEN)
+	$(MONITOR_CODE_GEN)
+
+$(MONITOR_CODE_GEN): $(wildcard cmd/monitorCodeGen/*.go)
```
keitwb: The camel-case is very striking when you're used to looking a Go code. Convention is all lowercase strings.

.gitignore
```diff
@@ -2,9 +2,11 @@ build/
 local-etc/
 /agent
 internal/monitors/**/template.go
+internal/monitors/**/generatedMetadata.go
```
keitwb: I'm not so sure the generated files should be ignored.  If we put them in the git repo, there should definitely be a CI check to ensure they are up to date, but it would be kind of nice to have them.  Same thing with the collectd template.go files, not sure why I ignored them to begin with.

keitwb: Here is a Reddit thread discussing it: https://www.reddit.com/r/golang/comments/8mbi47/do_you_commit_all_the_generated_code_in_your/

cmd/monitorCodeGen/main.go
```diff
@@ -0,0 +1,195 @@
...
+	Groups: groupSet,
+	GroupMetricsMap: groupMetricsMap,
+	SendAll: {{ .SendAll }},
+}
+{{end}}
+`
+
+// shouldRegenerate determines whether the metadata file needs regenerated based on its existence and timestamps.
+func shouldRegenerate(pkg *monitors.PackageMetadata) (bool, error) {
+	generatedMetadata := filepath.Join(pkg.PackagePath, "generatedMetadata.go")
```
keitwb: Can we just call it `metadata.go` or `metadata.gen.go` or something shorter?  Also same thing about camel cased file names.

cmd/monitorCodeGen/main.go
```diff
@@ -0,0 +1,195 @@
...
+		return false, fmt.Errorf("unable to stat %s", generatedMetadata)
+	}
+
+	// There's an existing generatedMetadata.go so check timestamps.
+	return statMetadataYaml.ModTime().After(statMetadataGenerated.ModTime()) ||
+		generatorStat.ModTime().After(statMetadataGenerated.ModTime()), nil
+}
+
+func generate() error {
+	pkgs, err := monitors.CollectMetadata("internal/monitors")
```
keitwb: There should be some mechanism somewhere to forcibly rewrite the generated metadata files, even if they exist and have newer timestamps than the source yaml.

cmd/monitorCodeGen/main.go
```diff
@@ -0,0 +1,195 @@
...
+{{- with .Alias}}
+	{{. | formatVariable}}: true,
+{{- end}}
+{{- end}}
+{{- end}}
+}
+
+var groupMetricsMap = map[string][]string {
+{{- range $group, $metrics := .groupMetricsMap}}
+	{{(printf "group.%s" $group) | formatVariable}}: []string {
```
keitwb: Why does it have a `group.` prefix?

cmd/monitorCodeGen/main.go
```diff
@@ -0,0 +1,195 @@
...
+		{{- range $metrics}}
+		{{. | formatVariable}},
+		{{- end}}
+	},
+{{- end}}
+}
+
+{{range .monitors}}
+var {{if gt (len $.monitors) 1 -}}
+{{- (printf "%s%s" .MonitorType "monitorMetadata") | formatVariable -}}
```
keitwb: "monitorMetadata" in this case should be capitalized right?

cmd/monitorCodeGen/main.go
```diff
@@ -0,0 +1,195 @@
...
+			0644); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func main() {
+	if err := generate(); err != nil {
```
keitwb: If this thing takes more than a second or two to run, let's parallelize it.

.gitignore
jaysfx: I personally consider them build artifacts and not something to be committed (unless it's a library that something else is going to depend on).  I don't really care either way though as long as it's consistent with however the collectd ones are done.

cmd/monitorCodeGen/main.go
jaysfx: I can add a specific `clean-code-gen` target. I just do `ack -g 'generatedMetadata.go|template.go' | xargs rm` right now. Or I can add a `--force` flag. Or both. Have a preference?

cmd/monitorCodeGen/main.go
jaysfx: If you have group named "mem" it'd result in a variable mem which seems weird and potentially conflicting with local variables. Metric tend to be fairly unique so not as much of a concern there.

cmd/monitorCodeGen/main.go
jaysfx: There's not a need for it to be exported currently since it's passed as a reference via Register.

cmd/monitorCodeGen/main.go
jaysfx: nop is 250ms, from clean is 500ms

keitwb:
I'm kind of wary of merging this into master until it actually does something useful with the Output.

cmd/monitorCodeGen/main.go
keitwb: A force flag seems easier to deal with.

cmd/monitorCodeGen/main.go
keitwb: But I mean if `.MonitorType` were, say, `cpu`, then wouldn't the variable name be `cpumonitorMetadata`?

.gitignore
keitwb: Ok, we can leave them out and it will be easy to add them in later if we find some convenient reason to do so.  I think in CI we can generate them fairly easily, but if it greatly complicates CI jobs to omit them from the repo, that would be my main reason to include them.

cmd/monitorCodeGen/main.go
keitwb: What am I missing here?  This is generating map key strings which aren't variables and also shouldn't they be put in quotes?  Or does printf/formatVariable add double quote literals around them?

cmd/monitorCodeGen/main.go
```diff
@@ -0,0 +1,195 @@
...
+const (
+{{- range . }}
+	{{(printf "group.%s" .Name) | formatVariable}} = "{{.Name}}"
+{{- end}}
+)
+{{end}}
+
+var groupSet = map[string]bool {
+{{- range .groupMetricsMap}}
+	{{(printf "group.%s" .Name) | formatVariable}}: true,
```
keitwb: Same thing here; if the variable name for the map is already `groupSet` why does there need to be a `group.` prefix for  all the key names?  Also I don't understand how this generates valid map literals since the keys don't seem to be in double quotes in the output Go src.

cmd/monitorCodeGen/main.go
jaysfx: Yeah it would be wrong in that case. Fixed it by separating the monitor type from monitorMetadata with a period which `formatVariable` will then fixup.

cmd/monitorCodeGen/main.go
jaysfx: Oh I think I misunderstood the original question. So a group "mem" gets a const generated called `groupMem` which is equal to `"mem"`. Then when generating the map it uses the const `groupMem` as the key. Should compile to the equivalent, it's just using our own constants as a safety measure.

jaysfx: Here's what it looks like:

```go
var groupMetricsMap = map[string][]string {
	groupContainers: []string {
		kubernetesContainerReady,
		kubernetesContainerRestartCount,
	},
	groupDaemonSets: []string {
		kubernetesDaemonSetCurrentScheduled,
		kubernetesDaemonSetDesiredScheduled,
		kubernetesDaemonSetMisscheduled,
		kubernetesDaemonSetReady,
	},
	groupDeployments: []string {
		kubernetesDeploymentAvailable,
		kubernetesDeploymentDesired,
	},
	groupNamespaces: []string {
		kubernetesNamespacePhase,
	},
	groupNodes: []string {
		kubernetesNodeReady,
	},
	groupPods: []string {
		kubernetesPodPhase,
	},
	groupReplicaSets: []string {
		kubernetesReplicaSetAvailable,
		kubernetesReplicaSetDesired,
	},
	groupReplicationControllers: []string {
		kubernetesReplicationControllerAvailable,
		kubernetesReplicationControllerDesired,
	},
	groupResourceQuotas: []string {
		kubernetesResourceQuotaHard,
		kubernetesResourceQuotaUsed,
	},
}
```

cmd/monitorCodeGen/main.go
jaysfx: Here's what it looks like:

```go
const (
	groupContainers = "containers"
	groupDaemonSets = "daemon-sets"
	groupDeployments = "deployments"
	groupNamespaces = "namespaces"
	groupNodes = "nodes"
	groupPods = "pods"
	groupReplicaSets = "replica-sets"
	groupReplicationControllers = "replication-controllers"
	groupResourceQuotas = "resource-quotas"
)


var groupSet = map[string]bool {
	groupContainers: true,
	groupDaemonSets: true,
	groupDeployments: true,
	groupNamespaces: true,
	groupNodes: true,
	groupPods: true,
	groupReplicaSets: true,
	groupReplicationControllers: true,
	groupResourceQuotas: true,
}
```

keitwb: oh ok, I see, it just references the constants.

internal/monitors/expvar/metadata.yaml
```diff
@@ -156,6 +157,7 @@
   - name: memstats.heap_sys
     type: gauge
     description: Bytes of heap memory obtained from the OS
+    included: false
```
keitwb: What's up with this?

jaysfx: It was just missing for some reason. Defaults to false but test code assumes they're set explicitly.

PR #681, opened 2019-03-30
adding example for bitlocker

jadeblaquiere:


python/sample/bitlocker.py
```diff
@@ -0,0 +1,31 @@
...
+        for ev in ev_list:
+            drive = ev.DriveLetter
+            bde_enabled = 0
+            bde_locked = 0
+            if ev.IsVolumeInitializedForProtection:
+                bde_enabled = 1
+                bde_locked = 1 if (ev.ProtectionStatus == 2) else 0
+            output.send_gauge("bitlocker_drive_encryption.enabled", bde_enabled, {"volume": drive})
+            output.send_gauge("bitlocker_drive_encryption.locked", bde_locked, {"volume": drive})
+    except Exception:
```
keitwb: We definitely want to log this exception content, probably at the error level like so:

```
# At top of module:
import logging
logger = logging.getLogger(__name__)


# ... here ...
except Exception as e:
    logger.error("Error getting bitlocker state, assuming not encrypted: %s", e)
    ...
```

jadeblaquiere: logger added

PR #688, opened 2019-04-03
bitlocker sample: fall back to minimum COM interface provided by 2008R2

jadeblaquiere:
apparently older hosts do not provide IsVolumeInitializedForProtection... 

also, the way it was previously structure any exception was triggering the exception handler. The only exception we want to specifically catch is when WMI fails to connect to the microsoftvolumeencryption module.


python/sample/bitlocker.py
```diff
@@ -13,19 +12,9 @@ def run(_, output):
 
     # if the bitlocker drive encryption feature is not installed then the attempt
     # to connect to the COM Object will raise an exception
+    mve = None
     try:
-        # BDE is installed
         mve = wmi.WMI(moniker="//./root/cimv2/security/microsoftvolumeencryption")
```
keitwb: The exception is being thrown by the WMI constructor?  It kind of looked like from the output you showed me that it was from the access to `IsVolumeInitializedForProtection`.

jadeblaquiere: That is the issue. The exception handler case should only happen when you fail to connect to the WMI path for micrsoftvolumeencryption. In the log, the exception was being thrown by access to `IsVolumeInitializedForProtection`, so you'll also see `Error connecting to Bitlocker feature, assuming not installed:, which is coming from except portion of the code (in that case it shouldn't be) and it means that it was executing both branches (and reporting incorrect data). 

This PR fixes that by limiting the scope of the try-except.

PR #689, opened 2019-04-03
Add metadata class

jaysfx:
This consolidates the metadata.yaml helper functions into a single class to
make it easier for test functions to query it.

### DIFF
--- a/tests/helpers/metadata.py
+++ b/tests/helpers/metadata.py
@@ -1,0 +1,68 @@
+import os
+
+import yaml
+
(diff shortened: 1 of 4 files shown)


internal/monitors/elasticsearch/metadata.yaml
```diff
@@ -1,3 +1,4 @@
+monitors:
```
keitwb: Doc test on CircleCI is complaining this yaml isn't valid.

PR #690, opened 2019-04-03
Azure pipelines

jchengsfx:


keitwb:
Looks like the Go test job is taking a long time because it is trying to download all the already vendored deps.  It looks like you are using `-mod vendor` on the test commands so I'm not sure why it is.

.azure-pipelines/templates/install-go.yml
```diff
@@ -0,0 +1,19 @@
...
+  displayName: 'Install go'
+- powershell: |
+    $env:GOBIN = "$(GOPATH)\bin"
+    echo "##vso[task.setvariable variable=GOBIN]$env:GOBIN"
+    go env
+    $env:PATH += ";$env:GOBIN"
+    echo "##vso[task.setvariable variable=PATH]$env:PATH"
+    echo $env:PATH
+- ${{ if ne(parameters.packages, '') }}:
+  - powershell: go get -u ${{ parameters.packages }}
```
keitwb: Do this in a different directory from the agent code so that it doesn't try and download everything from the go.mod file again.  Also make sure the GO111MODULE=on envvar is set here.

jaysfx:
This won't do any caching of the dependencies like go, python, etc. right? Would the Windows containers work so that the docker images could be prebuilt with the required dependencies? I haven't used Windows containers before myself. Alternatively, assuming they let you use your own custom VM images perhaps we could create our own with packer. Could possibly use the image for local building as well.

https://docs.microsoft.com/en-us/azure/virtual-machines/windows/build-image-with-packer
https://github.com/Microsoft/azure-pipelines-image-generation

.azure-pipelines/templates/install-go.yml
jchengsfx: `GO111MODULE=on` is already set as a global env var.
Changing to a different directory now installs the packages in seconds.

PR #692, opened 2019-04-04
Add datapoint filter with better semantics

keitwb:


internal/utils/filter/matching.go
```diff
@@ -63,3 +44,39 @@ func stripNegation(value string) (string, bool) {
...
+	negated, ok := m[s]
+	return ok, negated
+}
+
+type regexMatcher struct {
+	re      *regexp.Regexp
+	negated bool
+}
+
+var _ matcher = &regexMatcher{}
```
jaysfx: why is this and the other ones assigned to _ needed?

keitwb: That is an interface assertion so that it fails to compile if it doesn't really implement the interface.

internal/core/dpfilters/overridable.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"errors"
+
+	"github.com/signalfx/golib/datapoint"
+	"github.com/signalfx/signalfx-agent/internal/utils/filter"
+)
+
+type overridableDatapointFilter struct {
+	basicDatapointFilter
```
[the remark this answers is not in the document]
keitwb: No, there isn't much value to it, will remove the nesting.

keitwb:
> I just noticed, do the default agent config files need to be updated for `datapointsToExclude` for things like this in `deployments/fargate/agent.yaml`:
> 
> ```
> metricsToExclude:
>   - {"#from": "env:METRICS_TO_EXCLUDE", default: [], flatten: true}
> ```

That is a top-level filter, which doesn't change in this PR (this is only at the monitor level config).

keitwb:
> Also looks like helm templates need updating. Just thought about it after looking into it for my stuff.

Same thing, since the Helm template doesn't as of yet support monitor-level filtering config options, there is nothing to update.

PR #695, opened 2019-04-05
Add extra metric support

jaysfx:
Additional metrics filters for included metrics as well as any additional
metrics the user has configured.

---
This change is [Reviewable](https://reviewable.io/reviews/signalfx/signalfx-agent/695)


internal/monitors/manager.go
```diff
@@ -105,7 +105,7 @@ func (mm *MonitorManager) Configure(confs []config.MonitorConfig, collectdConf *
 		if err != nil {
 			log.WithFields(log.Fields{
 				"monitorType": conf.Type,
-				"error":       err,
+				"error":       err.Error(),
```
jaysfx: I changed this to `.Error()` because it was including tracebacks from `Wrapf` calls.

internal/monitors/collectd/kafka/kafka.go
```diff
@@ -39,7 +39,7 @@ func init() {
 	}
 	defaultMBeans = defaultMBeans.MergeWith(genericjmx.DefaultMBeans)
 
-	monitors.Register(monitorType, func() interface{} {
```
keitwb: Make sure the `monitorType` const is consistently deleted so it isn't duplicated with the value in `monitorMetadata`.

internal/monitors/diagnostics.go
```diff
@@ -72,19 +72,23 @@ func (mm *MonitorManager) DiagnosticText() string {
...
+Monitored Endpoint ID: %s`,
 				am.config.MonitorConfigCore().DiscoveryRule,
 				am.endpoint.Core().ID)
 		}
 		activeMonText += fmt.Sprintf(
-			"%s. %s\n"+
-				"    Reporting Interval (seconds): %d\n"+
-				"%s"+
-				"    Config:\n%s\n",
+			`%s. %s
```
keitwb: This indentation is pretty hard to read.

internal/monitors/diagnostics.go
```diff
@@ -72,19 +72,23 @@ func (mm *MonitorManager) DiagnosticText() string {
...
+    Reporting Interval (seconds): %d
+    Enabled Metrics:
+%s
+%s    Config:
+%s
+`,
 			am.config.MonitorConfigCore().MonitorID,
 			am.config.MonitorConfigCore().Type,
 			am.config.MonitorConfigCore().IntervalSeconds,
+			utils.IndentLines(config.ToString(am.enabledMetrics), 6),
```
keitwb: Consider making this stringify to a one-line list string instead of multiple lines, if it isn't already.

internal/monitors/manager.go
keitwb: Yeah I don't even use that Wrapf thing anymore, the errors are too verbose.

internal/monitors/manager.go
```diff
@@ -314,26 +314,39 @@ func (mm *MonitorManager) findConfigForMonitorAndRun(endpoint services.Endpoint)
...
 
 	instance := newMonitor(config.MonitorConfigCore().Type, id)
 	if instance == nil {
-		return errors.Errorf("Could not create new monitor of type %s", config.MonitorConfigCore().Type)
+		return errors.Errorf("Could not create new monitor of type %s", monitorType)
+	}
+
+	metadata, ok := MonitorMetadatas[monitorType]
+	if !ok {
+		return errors.Errorf("could not find monitor metadata of type %s", monitorType)
```
keitwb: If this is missing, doesn't it indicate something is majorly wrong with the logic that registers monitors?  If so, it should just panic.

internal/monitors/manager.go
```diff
@@ -314,26 +314,39 @@ func (mm *MonitorManager) findConfigForMonitorAndRun(endpoint services.Endpoint)
...
 	}
 
 	configHash := config.MonitorConfigCore().Hash()
 
+	extraMetricsFilter, err := newMetricsFilter(metadata, coreConfig.ExtraMetrics, coreConfig.ExtraGroups)
+	if err != nil {
+		return errors.Wrapf(err, "unable to construct extraMetrics filter")
```
keitwb: Don't use Wrapf.

internal/monitors/manager.go
jaysfx: Too verbose due to the stack trace or due to it printing the chain of error strings?

internal/monitors/diagnostics.go
jaysfx: I don't really know how to fix it given `go fmt` and its opinions. The indentation was previously weird as well.

internal/monitors/manager.go
keitwb: The stack trace is too long and unnecessary for end users.

internal/monitors/diagnostics.go
keitwb: Yes, except maybe with a `[ ... ]` around it (the YAML compact style).

keitwb: Also, this would include the "included" metrics too right?

internal/monitors/manager.go
```diff
@@ -314,26 +314,49 @@ func (mm *MonitorManager) findConfigForMonitorAndRun(endpoint services.Endpoint)
...
 
 	configHash := config.MonitorConfigCore().Hash()
 
+	var extraMetrics []string
+
+	// Monitors can add additional extra metrics to allow through such as based on config flags.
+	if monitorExtra := config.GetExtraMetrics(); monitorExtra != nil {
+		extraMetrics = append(extraMetrics, coreConfig.ExtraMetrics...)
+		extraMetrics = append(extraMetrics, monitorExtra...)
+	} else {
```
keitwb: Any reason why it couldn't do both.  This would just silently ignore `ExtraMetrics` on the config struct this way?

keitwb: Or maybe just have the default implementation of GetExtraMetrics return ExtraMetrics from the config struct.

internal/monitors/diagnostics.go
jaysfx: Yeah it includes "included" metrics as well. It's going to look messy if it wraps around though. What's the concern with laying it out vertically?

internal/monitors/manager.go
jaysfx: I may be missing your point but I think it's doing both. If `config.GetExtraMetrics()` returns non-nil it merges that and the user-specified `coreConfig.ExtraMetrics`. If the default returns `ExtraMetrics` it puts the onus on the monitor that overrides it (like ES) to do the merging.

internal/monitors/diagnostics.go
keitwb: That it is too long and requires excessive scrolling, esp for monitors with a lot of included metrics.  A lot of the work I just did on the status output was to make it more manageable and understandable.  Something like:

```
Enabled Metrics: metric1, metric2, metric3, metric4,
                             metric5, metric6
```

where it does multiple lines (not just one giant line) up to a certain column width would be nice if feasible.  We just need some formatter utility function to do that.

internal/monitors/manager.go
keitwb: Ok, in that case this can be simplified to do the handling of config.ExtraMetrics once, without an else block.

internal/monitors/manager.go
```diff
@@ -314,26 +314,49 @@ func (mm *MonitorManager) findConfigForMonitorAndRun(endpoint services.Endpoint)
...
 	output := &monitorOutput{
-		monitorType:               config.MonitorConfigCore().Type,
+		monitorType:               monitorType,
 		monitorID:                 id,
-		notHostSpecific:           config.MonitorConfigCore().DisableHostDimensions,
-		disableEndpointDimensions: config.MonitorConfigCore().DisableEndpointDimensions,
-		filter:                    config.MonitorConfigCore().Filter,
+		notHostSpecific:           coreConfig.DisableHostDimensions,
+		disableEndpointDimensions: coreConfig.DisableEndpointDimensions,
+		filter:                    coreConfig.Filter,
```
keitwb: I apologize in advance but I'm going to make a merge conflict with this once I merge https://github.com/signalfx/signalfx-agent/pull/701.

Is it possible to unify the `extraMetricsFilter` with the other filters by making it implement `dpfilters.DatapointFilter` interface?

internal/monitors/metricsfilter.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if strings.ContainsRune(metricName, '*') {
+		// Make sure a metric pattern matches at least one metric.
+		m, err := filter.NewBasicStringFilter([]string{metricName})
+		if err != nil {
+			return err
+		}
+
+		for metric := range metadata.Metrics {
```
keitwb: Would this cover the metrics that have parmaters in their name in the metadata (e.g. the `kubernetes.node_<condition_type>` one from `kubernetes-cluster`)?

internal/monitors/metricsfilter.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	for metric := range add.metadata.IncludedMetrics {
+		metrics = append(metrics, metric)
+	}
+
+	return metrics
+}
+
+func (add *metricsFilter) shouldSend(dp *datapoint.Datapoint) bool {
```
keitwb: If you call this method `Matches`, then this struct would implement `dpfilters.DatapointFilter`.

keitwb: Also `add` is kind of a strange name.

internal/monitors/metricsfilter.go
```diff
@@ -0,0 +1,153 @@
...
+		// It's an included metric so send by default.
+		return true
+	}
+
+	if add.extraMetrics[dp.Metric] {
+		// User has explicitly chosen to send this metric (or a group that this metric belongs to).
+		return true
+	}
+
+	if add.metadata.MetricsExhaustive && !add.metadata.HasMetric(dp.Metric) {
```
keitwb: Not sure about these last two conditionals.

It seems like from this point down it should just return whether the metric matches the `add.stringFilter` and not worry about whether it is in the metadata list or not.

keitwb: Esp this conditional here would prevent somebody from being able to add a metric in if a developer made a typo in the metric name in the metadata.yaml.

internal/monitors/manager.go
jaysfx: Yeah would make sense to unify them. The one thing I like about the name `shouldSend` is that it makes it clear what the boolean return value means. With `Matches` I forget if `true` means it will be sent or won't be sent. Having the interface be `ShouldSend() bool` would make this clear but also seems kind of weird. Thoughts?

internal/monitors/metricsfilter.go
jaysfx: See above comment reply re: Matches.

internal/monitors/metricsfilter.go
jaysfx: Yeah it allows it through because it's not mark as exhaustive. Maybe we can add the snake case as a special case in the future to allow kubernetes-cluster to be marked as exhaustive.

internal/monitors/manager.go
keitwb: As we talked about, the filter variable that is being used to call `Matches` on should be named something descriptive that indicates whether it drops or includes things.

internal/monitors/manager.go
```diff
@@ -349,31 +349,31 @@ func (mm *MonitorManager) createAndConfigureNewMonitor(config config.MonitorCust
...
-		endpoint:                  endpoint,
-		dpChan:                    mm.DPs,
-		eventChan:                 mm.Events,
-		dimPropChan:               mm.DimensionProps,
-		spanChan:                  mm.TraceSpans,
-		extraDims:                 map[string]string{},
+		notHostSpecific:           config.MonitorConfigCore().DisableHostDimensions,
+		disableEndpointDimensions: config.MonitorConfigCore().DisableEndpointDimensions,
+		filterSet: &dpfilters.FilterSet{
+			ExcludeFilters: []dpfilters.DatapointFilter{oldFilter, newFilter},
```
keitwb: I think you could make this line:

```
 			ExcludeFilters: []dpfilters.DatapointFilter{oldFilter, newFilter, dpfilters.Negate(extraMetricsFilter)},
```

And not even need an additional field on the Output instance or the extra logic in the SendDatapoint method.

What do you think?

keitwb: Taking nils into account of course.

internal/core/config/monitor.go
```diff
@@ -129,8 +133,14 @@ func (mc *MonitorConfig) Hash() uint64 {
...
+// GetExtraMetrics returns additional metrics that should be allowed through.
+func (mc *MonitorConfig) GetExtraMetrics() []string {
+	return nil
+}
+
 // MonitorCustomConfig represents monitor-specific configuration that doesn't
 // appear in the MonitorConfig struct.
 type MonitorCustomConfig interface {
 	MonitorConfigCore() *MonitorConfig
+	GetExtraMetrics() []string
```
keitwb: I think this would be a bit cleaner as a separate interface instead of reusing this one.  Then, the monitor's config types can do an interface assertion (the `var _ ... = ` thing) and get some compiler check that they are actually implementing the interface and make it clear to those reading the monitor the intended interface.

Also Go convention for getters is to not have a `Get` prefix.

internal/monitors/metricsfilter.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return metrics, nil
+}
+
+func newMetricsFilter(metadata *Metadata, extraMetrics, extraGroups []string) (*extraMetricsFilter, error) {
+	var filterItems []string
+
+	for _, metric := range extraMetrics {
+		if err := validateMetricName(metadata, metric); err != nil {
+			return nil, err
```
keitwb: This bubbles up to preventing the monitor from starting, when it should really just be a warning as it isn't a critical error that should prevent monitoring.

internal/monitors/metricsfilter.go
```diff
@@ -0,0 +1,144 @@
...
+		// If the user specified a metric that's already included no need to add it.
+		if !metadata.IncludedMetrics[metric] {
+			filterItems = append(filterItems, metric)
+		}
+	}
+
+	for _, group := range extraGroups {
+		metrics, err := validateGroup(metadata, group)
+		if err != nil {
+			return nil, err
```
keitwb: Same thing here.

PR #696, opened 2019-04-08
Simplifying expvar integration

MovieStoreGuy:
I had spent some time making sure the integration was easier to understand for those who would like to contribute.

A lot of the changes have to do with error message lengths as they were rather verbose and would be best as concise, clear statements of what is wrong.


bjsignalfx:
@keitwb it looks good to me too. I'll test the changes locally then merge

PR #698, opened 2019-04-08
Add StatsD monitor

seonsfx:
This PR is including a new StatsD monitor that is capable of receiving statsd metrics, parsing metric name/dimensions and sending the metrics to Ingest.

Here is a set of sample inputs:
```
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_gateway-vn_tcp_8080.membership_change:1|c
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_gateway-vn_tcp_8080.update_attempt:2|c
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_gateway-vn_tcp_8080.update_success:3|c
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_gateway-vn_tcp_8080.upstream_cx_destroy:4|c
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_payment-2-vn_http_5003.upstream_rq_active:5|g
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_payment-2-vn_http_5003.upstream_rq_pending_active:6|g
statsd.appmesh.cluster.cds_ingress_ecommerce-demo-mesh_api-vn_http_5000.circuit_breakers.default.cx_open:7|g
statsd.appmesh.cluster.cds_ingress_ecommerce-demo-mesh_api-vn_http_5000.circuit_breakers.default.rq_pending_open:8|g
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_catalog-vn_http_5004.upstream_cx_connect_ms-average:9|ms
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_catalog-vn_http_5004.upstream_rq_time-average:10|ms
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_checkout-vn_http_5002.external.upstream_rq_time-average:11|ms
statsd.appmesh.cluster.cds_egress_ecommerce-demo-mesh_checkout-vn_http_5002.upstream_cx_connect_ms-average:12|ms
```

When this sample is sent to a open UDP port by the monitor, the monitor will process the input and send these datapoints below:
```
metricName:egress.upstream_rq_active metricType:g value:5 dimensions:map[mesh:ecommerce-demo-mesh service:payment-2 traffic:egress]
metricName:egress.upstream_rq_pending_active metricType:g value:6 dimensions:map[mesh:ecommerce-demo-mesh service:payment-2 traffic:egress]
metricName:ingress.circuit_breakers.default.cx_open metricType:g value:7 dimensions:map[mesh:ecommerce-demo-mesh service:api traffic:ingress]
metricName:ingress.circuit_breakers.default.rq_pending_open metricType:g value:8 dimensions:map[mesh:ecommerce-demo-mesh service:api traffic:ingress]
metricName:egress.upstream_cx_connect_ms-average metricType:ms value:9 dimensions:map[mesh:ecommerce-demo-mesh service:catalog traffic:egress]
metricName:egress.upstream_rq_time-average metricType:ms value:10 dimensions:map[mesh:ecommerce-demo-mesh service:catalog traffic:egress]
metricName:egress.external.upstream_rq_time-average metricType:ms value:11 dimensions:map[mesh:ecommerce-demo-mesh service:checkout traffic:egress]
metricName:egress.upstream_cx_connect_ms-average metricType:ms value:12 dimensions:map[mesh:ecommerce-demo-mesh service:checkout traffic:egress]
```

Below is a list of things that are being worked on currently and will be delivered soon.
- Better docs
- Unit tests

Below is a list of things that need to be done later
- TCP support
- Support for dot-containing dimension values


internal/monitors/statsd/metadata.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+    You can customize a metric name by providing a format string within the converter configuration.
+    
+    ```
+    converters:
+      - pattern: "cluster.cds_{traffic}_{mesh}_{service}-vn_{protocol}_{port}.{action}"
+        metric: "{traffic}.{action}"
+    ```
+    
+    The metrics which match to the given pattern will be reported to SignalFx as `{traffic}.{action}`.
```
keitwb: Are `traffic` and `action` also dimensions in this case or not?

internal/monitors/statsd/metadata.yaml
```diff
@@ -0,0 +1,75 @@
...
+    For instance, metric `cluster.cds_egress_ecommerce-demo-mesh_gateway-vn_tcp_8080.update_success`
+    will be reported as `egress.update_success`.
+
+    The original metric name will be sent without being processed if there's no related configuration.
```
keitwb: Don't allow the original metric name to be sent if there is a pattern specified.  Force the user to specify how to create a metric name in that case.

If no configuration at all (no `pattern` or `metric`), only then should the metric be passed through as is.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+		converters: converters,
+	}
+
+	err := listener.Listen()
+	if err != nil {
+		return err
+	}
+
+	utils.RunOnInterval(ctx, func() {
+		metrics, err := listener.FetchMetrics()
```
keitwb: The listener should be continuously reacting to new UDP datagrams (and TCP streams) such that it pulls data off of the socket immediately, instead of letting it buffer for many seconds and then flushing the whole thing at once.  Otherwise, you are relying on the kernel to buffer things and you have little to no control over how much it will do before dropping things.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			return
+		}
+
+		m.sendDatapoints(metrics)
+	}, conf.ServerTimeout)
```
keitwb: There is already a standard config option for the interval called `conf.IntervalSeconds` that should always be used.

internal/monitors/statsd/metadata.yaml
seonsfx: `dimensions` config wasn't given, so they won't be any dimensions in this case.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,148 @@
...
+				dimSrc = removePrefix(rawMetricName, prefix)
+			} else {
+				dimSrc = rawMetricName
+			}
+
+			metricName, dims = convertMetric(dimSrc, converters)
+		}
+
+		strValue := m[colonIdx+1 : pipeIdx]
+		value, err := strconv.ParseInt(strValue, 10, 64)
```
keitwb: They can be floats as well.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,148 @@
...
+
+		if err == nil {
+			metrics = append(metrics, &statsDMetric{
+				metricName: metricName,
+				metricType: m[pipeIdx+1:],
```
keitwb: There is an optional component after the metric type that specified the rate.  See https://github.com/statsd/statsd/blob/master/docs/metric_types.md#sampling.  We don't have to support this but it should at least parse it right.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,148 @@
...
+
+		rawMetricName := m[0:colonIdx]
+		metricName := ""
+
+		var dims map[string]string
+
+		if converters == nil {
+			metricName = rawMetricName
+		} else {
+			var dimSrc string
```
keitwb: This var name is very confusing.  It is the metric name.

internal/monitors/statsd/parser.go
```diff
@@ -0,0 +1,147 @@
...
+			for _, d := range c.dimensions {
+				if v, exists := fields[d]; exists {
+					dim[d] = v
+				}
+			}
+		}
+
+		// Compose metricName
+		if c.metric == nil {
+			metricName = name
```
keitwb: Don't ever let the metricName be the original metric name if it matches a converter.  Make the user specify it.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (m *Monitor) sendDatapoints(metrics []*statsDMetric) {
+	for _, metric := range metrics {
+		var dp *datapoint.Datapoint
+
+		// StatsD Metric Types https://github.com/statsd/statsd/blob/master/docs/metric_types.md
+		switch metric.metricType {
+		case "g":
+			if !m.conf.DeleteGauges {
```
keitwb: This config option isn't being used properly.  It means whether the gauges should be persisted across a regular interval (for gauges, they should be sent as 0 if not received).

For simplicity, I would just get rid of the `Delete*` config options and not worry about them at all for the first iteration of this monitor.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+			return
+		}
+
+		m.sendDatapoints(metrics)
+	}, conf.ServerTimeout)
+
+	return nil
+}
+
+func (m *Monitor) sendDatapoints(metrics []*statsDMetric) {
```
keitwb: Make this a function that receives a single `statsdDMetric` instance and converts it to a datapoint and do the sending somewhere else.  This function is trying to do too much.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"false" singleInstance:"true"`
+	// The host/address on which to bind the UDP listener that accepts statsd
+	// datagrams
+	ListenAddress string `yaml:"listenAddress" default:"localhost"`
+	// The port on which to listen for statsd messages
+	ListenPort uint16 `yaml:"listenPort" default:"8125"`
+	// The protocol for statsd communications
+	Protocol string `yaml:"protocol" default:"udp"`
```
keitwb: Instead of a free-form string config field, it is easier for users to have a bool field called `listenTCP` that defaults to `false`, but if set to true would listen on a TCP socket.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+
+type converterInput struct {
+	Pattern    string
+	Metric     string
+	Dimensions []string
+}
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"false" singleInstance:"true"`
```
keitwb: There is no reason this should be `singleInstance`.  You can perfectly well have multiple statsd monitors configured in the same agent, as long as you make the ports different.

internal/monitors/statsd/parser.go
```diff
@@ -0,0 +1,147 @@
...
+			isField := c.metric.startWithField
+			for _, substr := range c.metric.substrs {
+				if isField {
+					if v, exists := fields[substr]; exists {
+						metricName = metricName + v
+					}
+				} else {
+					metricName = metricName + substr
+				}
+				isField = !isField
```
keitwb: Does this handle cases where two fields are back-to-back in the string?

seonsfx: Two fields cannot exist without having a proper separator in between I believe.

internal/monitors/statsd/metadata.yaml
seonsfx: That would essentially mean make a metric name for a converter required. I'll do so.

docs/monitors/statsd.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### Adding dimensions to StatsD metrics
+
+The StatsD monitor can parse keywords from a statsd metric name by a set of
+converter that was configured by user.
+
+```
+converters:
+  - pattern: "cluster.cds_{traffic}_{mesh}_{service}-vn_{protocol}_{port}.{action}"
+    dimensions:
```
keitwb: This example is confusing, what is the point of the `protocol` and `port` sections here?  It is probably better to just send everything pulled out as a "section" as a dimension automatically without having to have a list.

docs/monitors/statsd.md
```diff
@@ -0,0 +1,110 @@
...
+
+Metrics will be sent without dimensions if there's no related configuration.
+
+#### Formatting metric name
+
+You can customize a metric name by providing a format string within the converter configuration.
+
+```
+converters:
+  - pattern: "cluster.cds_{traffic}_{mesh}_{service}-vn_{protocol}_{port}.{action}"
```
keitwb: Again, this is confusing.  Are those other four sections turned into dimensions or not?  It seems like they should be if they are bothered to be named.

internal/monitors/statsd/metadata.yaml
keitwb: I think if they named a section in the pattern, it should be a dimension automatically.  If you want to be able to "throw away" sections without adding them as dimensions, just let them do something like `{*}` or something, but I think in most cases they will want the diemnsions anyway.

internal/monitors/statsd/metadata.yaml
keitwb: Docs don't seem to indicate that metricName is required.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,158 @@
...
+
+func convertMetricsToDatapoints(metrics map[string]*statsDMetric) []*datapoint.Datapoint {
+	var dps []*datapoint.Datapoint
+
+	for _, metric := range metrics {
+		var dp *datapoint.Datapoint
+
+		// StatsD Metric Types https://github.com/statsd/statsd/blob/master/docs/metric_types.md
+		switch metricTypeMap[metric.metricType] {
+		case "gauge":
```
keitwb: Use these const enums from golib: https://github.com/signalfx/golib/blob/master/datapoint/datapoint.go#L22.  Note that `Count` is a counter and `Counter` is a cumulative counter (it is kind of confusing).

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,158 @@
...
+	if m.cancel != nil {
+		m.cancel()
+	}
+}
+
+func aggregateMetrics(metrics []*statsDMetric) map[string]*statsDMetric {
+	metricsMap := make(map[string]*statsDMetric)
+
+	for _, metric := range metrics {
+		if _, exists := metricsMap[metric.metricName]; exists {
```
keitwb: It seems like the rest of this for loop can be simplified to just:

```
switch metricTypeMap[metric.metricType] {
case datapoint.Count:
    // Add up
    metricsMap[metric.metricName].value = metricsMap[metric.metricName].value + metric.value
default:
    metricsMap[metric.metricName] = metric
```

internal/monitors/statsd/parser.go
```diff
@@ -0,0 +1,153 @@
...
+
+type converter struct {
+	pattern    *fieldPattern
+	metric     *fieldPattern
+	dimensions []string
+}
+
+func initConverter(input *converterInput) *converter {
+	if input.Pattern == "" {
+		logger.Error("[pattern] is required for a converter")
```
keitwb: It is generally better to do validation of this stuff in the `Config.Validate() error` method and then these functions can be more focused and not have to deal with error checking.

internal/monitors/statsd/parser.go
```diff
@@ -0,0 +1,153 @@
...
+		wc = wa + strings.Index(p[wa:], "}")
+
+		if wa == 0 && wb == 0 {
+			startWithField = true
+		}
+
+		if wa != wb {
+			// edge case : pattern not ending with { and }
+			if wb == -1 {
+				return nil // Invalid pattern, skip.
```
keitwb: Would probably be helpful for this function to also return an error explaining that there was a problem parsing and then have that logged somewhere so the user gets some feedback.

internal/monitors/statsd/parser.go
keitwb: That's true.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,160 @@
...
+func (sl *statsDListener) readTCP(chData chan []byte) {
+}
+
+func (sl *statsDListener) readUDP(chData chan []byte) {
+	// UDP needs to receive data packet by packet. Max packet size is 65535 for now.
+	buf := make([]byte, 65536)
+	for {
+		n, _, err := sl.udpConn.ReadFromUDP(buf)
+
+		if err != nil {
```
keitwb: The only way to shut this thing down is to call `sl.updConn.Close()` (https://golang.org/pkg/net/#UDPConn.Close), which you need to do somewhere through the monitor's `Shutdown` method.  If you just log the errors without considering whether it is because the listener was stopped, this will turn into an infinite loop and make the monitor unstoppable.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (sl *statsDListener) readTCP(chData chan []byte) {
+}
+
+func (sl *statsDListener) readUDP(chData chan []byte) {
+	// UDP needs to receive data packet by packet. Max packet size is 65535 for now.
+	buf := make([]byte, 65536)
+	for {
+		n, _, err := sl.udpConn.ReadFromUDP(buf)
```
keitwb: Edge case here is if `n == len(buf)` then you might be in a situation where part of a line is read and it will fail to parse properly and be dropped unnecessarily.  The simplest way to handle it is to check if `n == len(buf)` and if so, just chop off everything after the last `\n` and keep it in a temporary variable that gets prepended to the next read result.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,160 @@
...
+			sl.metricBuffer = append(sl.metricBuffer, strings.Split(string(data), "\n")...)
+		}
+	}
+}
+
+func (sl *statsDListener) readTCP(chData chan []byte) {
+}
+
+func (sl *statsDListener) readUDP(chData chan []byte) {
+	// UDP needs to receive data packet by packet. Max packet size is 65535 for now.
```
keitwb: IPv4 UDP max packet size is a bit under 64k but for IPv6 it can be much bigger.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,160 @@
...
+		secondPipeIdx := pipeIdx + strings.Index(m[pipeIdx+1:], "|")
+
+		rawMetricName := m[0:colonIdx]
+		metricName := ""
+
+		var metricType string
+		if secondPipeIdx > pipeIdx {
+			metricType = m[pipeIdx+1 : secondPipeIdx]
+		} else {
+			metricType = m[pipeIdx+1:]
```
keitwb: What happens if the metric string ends in a `|`?  Just make sure it won't panic and crash the whole agent due to an out of bounds array index.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,160 @@
...
+			})
+		} else {
+			logger.WithError(err).Errorf("Failed parsing metric value %s", strValue)
+		}
+	}
+
+	return metrics
+}
+
+func removePrefix(original string, prefix string) string {
```
keitwb: Maybe just use `strings.TrimPrefix` instead.

internal/monitors/statsd/statsd.go
seonsfx: @keitwb Did you mean this? https://en.wikipedia.org/wiki/Jumbogram

seonsfx: Looks like pretty much standard https://github.com/prometheus/statsd_exporter/commit/b8aab013c6538118225b031a46dafc344290debe#diff-557eb2a359922e8de5f18397fed0cd99

internal/monitors/statsd/metadata.yaml
seonsfx: I'll allow `{}` for non-collecting section. Concerned if `*` give the user a wrong impression that we support wildcards.

docs/monitors/statsd.md
```diff
@@ -0,0 +1,94 @@
...
+
+The metrics which match to the given pattern will be reported to SignalFx as `{traffic}.{action}`.
+For instance, metric `cluster.cds_egress_ecommerce-demo-mesh_gateway-vn_tcp_8080.update_success`
+will be reported as `egress.update_success`.
+
+`metric` is required for a converter configuration. A converter will be disabled if `metric` is not provided.
+
+
+
+Sample YAML configuration:
```
keitwb: I would just get rid of this section since you already have several samples before this.  It is also outdated since it has the `dimensions` config.

docs/monitors/statsd.md
```diff
@@ -0,0 +1,94 @@
...
+| `listenTCP` | no | `bool` | The protocol for statsd communications (**default:** `false`) |
+| `metricPrefix` | no | `string` |  |
+| `converters` | no | `list of objects (see below)` |  |
+
+
+The **nested** `converters` config object has the following fields:
+
+| Config option | Required | Type | Description |
+| --- | --- | --- | --- |
+| `pattern` | no | `string` |  |
```
keitwb: These need documentation.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,177 @@
...
+
+var logger = utils.NewThrottledLogger(log.WithFields(log.Fields{"monitorType": monitorType}), 30*time.Second)
+var golibLogger = &utils.LogrusGolibShim{FieldLogger: logger.FieldLogger}
+
+func init() {
+	monitors.Register(monitorType, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// converterInput is to receive configs to setup metric converters
+type converterInput struct {
```
keitwb: Make this exported.  It makes wrapping this a lot easier (which you will need for the AppMesh monitor).

tests/monitors/statsd/statsd_test.py
```diff
@@ -19,57 +23,72 @@ def test_statsd_monitor():
...
+      metricName: '{traffic}.{action}'
+"""
+    ) as [backend, _, _]:
+        assert wait_for(p(udp_port_open_locally_netstat, 8127)), "statsd port never opened!"
+        send_udp_message(
+            "localhost", 8127, "cluster.cds_egress_ecommerce-demo-mesh_gateway-vn_tcp_8080.update_success:8|c"
+        )
+
+        assert wait_for(
+            p(has_datapoint_with_metric_name, backend, "derive.egress.update_success")
```
keitwb: Do this: 

`p(has_datapoint, metric_name="derive.egress.update_success", dimensions={"traffic": "egress", "mesh": ...})` to assert that it is getting all those things in a single datapoint.

internal/monitors/statsd/monitor.go
```diff
@@ -0,0 +1,182 @@
...
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"false" singleInstance:"false"`
+	// The host/address on which to bind the UDP listener that accepts statsd
+	// datagrams
+	ListenAddress string `yaml:"listenAddress" default:"localhost"`
+	// The port on which to listen for statsd messages
+	ListenPort uint16 `yaml:"listenPort" default:"8125"`
+	// The protocol for statsd communications
+	ListenTCP bool `yaml:"listenTCP" default:"false"`
```
keitwb: Before we release this, you either need to implement TCP or hide this config option so that you don't tease users with it.

seonsfx: I'll hide it for now. I think that was the whole point of having it as `listenTCP` and default to false.

seonsfx:
@keitwb AppMesh Monitor and a unit test to test it have been added.

internal/monitors/appmesh/metadata.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+    ```yaml
+    monitors:
+     - type: appmesh
+       listenAddress: 0.0.0.0
+       listenPort: 8125
+       metricPrefix: statsd.appmesh
+    ```
+
+  metrics: null
```
keitwb: Fill out the expected metric set so they are documented for users.

internal/monitors/appmesh/metadata.yaml
```diff
@@ -0,0 +1,20 @@
+monitors:
+- dimensions: null
+  doc: |
+
+    This monitor starts a StatsD monitor to listen to StatsD metrics emitted
+    by AWS AppMesh Envoy Proxy.
```
keitwb: This is a good place to explain to users how to configure Envoy to send the metrics to the agent.

internal/monitors/appmesh/metadata.yaml
seonsfx: Done.

internal/monitors/appmesh/metadata.yaml
seonsfx: Done. @keitwb

docs/monitors/appmesh.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+
+## Metrics
+
+The following table lists the metrics available for this monitor. Metrics that are marked as Included are standard metrics and are monitored by default.
+
+| Name | Type | Included | Description |
+| ---  | ---  | ---    | ---         |
+| `gauge.<ingress/egress>.circuit_breakers.<priority>.cx_open` | gauge | ✔ | Whether the connection circuit breaker is closed (0) or open (1) |
```
keitwb: Would be much better if metric names didn't have "gauge"/"derive" in front of them.

keitwb: Also, can ingress/egress just be a dimenison?

seonsfx: Also, can ingress/egress just be a dimenison? -> Sure

internal/monitors/appmesh/metadata.yaml
```diff
@@ -0,0 +1,272 @@
...
+  - description: External origin aggregate HTTP response codes
+    included: true
+    name: derive.<ingress/egress>.external.upstream_rq_<_xx>
+    type: cumulative
+  - description: External origin specific HTTP response codes
+    included: true
+    name: derive.<ingress/egress>.external.upstream_rq_<_>
+    type: cumulative
+  - description: External origin request time milliseconds
+    included: true
```
keitwb: We are going to use all these metrics in the built-in content?

docs/monitors/appmesh.md
seonsfx: "gauge"/"derive" -> this is to match collectd/statsd plugin.

internal/monitors/statsd/parser.go
keitwb: Don't forget about this.

internal/monitors/appmesh/metadata.yaml
seonsfx: Good point. I'll turn off what's not being used for the built-in content

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,168 @@
...
+func (sl *statsDListener) readUDP(chData chan []byte) {
+	// UDP needs to receive data packet by packet. Max packet size is 65535 for now.
+	buf := make([]byte, 65536)
+	for {
+		n, _, err := sl.udpConn.ReadFromUDP(buf)
+
+		if err != nil {
+			logger.WithError(err).Error("Failed reading UDP datagram.")
+			// Exit the loop if the connection is closed
+			if strings.Contains(err.Error(), "use of closed network connection") {
```
keitwb: This is just waiting to be a silent bug when we upgrade Go and they changed the error string and then we get goroutine leaks in the agent.

If you can't figure out a robust way to check the return value from `ReadFromUDP`, then set an atomic flag on the struct and check it to know if should shutdown.

docs/monitors/appmesh.md
keitwb: > Also, can ingress/egress just be a dimenison? -> Sure

If it is a dimension, wouldn't the metric names change now?

keitwb: Oh, docs are out of date.

internal/monitors/statsd/monitor.go
keitwb: Don't forget about this.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (sl *statsDListener) readTCP(chData chan []byte) {
+}
+
+func (sl *statsDListener) readUDP(chData chan []byte) {
+	// UDP needs to receive data packet by packet. Max packet size is 65535 for now.
+	buf := make([]byte, 65536)
+	for {
+		// Exit the loop if the connection is closed
+		if atomic.LoadInt32(&sl.shutdownCalled) > 0 {
```
keitwb: Only downside to doing it up here is that it will log an error on monitor shutdown that isn't really an error.  It should do this check in the `if err != nil` block below before logging out an error.

seonsfx: Thought wouldn't even get to the `if err != nil` but I see it gets there even after turning on the flag. Guess the flag is still not on when the loop checks for it for the first time after the connection is closed. Moved the flag check into the error block.

internal/monitors/statsd/parser.go
seonsfx: Don't know why it's not showing up, but it's been added.
https://github.com/signalfx/signalfx-agent/blob/ae6f1373f7e367934f9015206a2a02208214cecb/internal/monitors/statsd/parser.go#L51

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,162 @@
...
+			metricName = strings.TrimPrefix(rawMetricName, prefix+".")
+		} else {
+			metricName = rawMetricName
+		}
+
+		if converters != nil {
+			metricName, dims = convertMetric(metricName, converters)
+		}
+
+		strValue := m[colonIdx+1 : pipeIdx]
```
keitwb: Guard against `:` as last character in metric string.

internal/monitors/statsd/statsd.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func parseMetrics(raw []string, converters []*converter, prefix string) []*statsDMetric {
+	var metrics []*statsDMetric
+
+	for _, m := range raw {
+		colonIdx := strings.Index(m, ":")
+		pipeIdx := strings.Index(m, "|")
+		if pipeIdx >= len(m)-1 || pipeIdx < 0 || colonIdx+1 > len(m) || colonIdx < 0 {
```
keitwb:
```suggestion
		if pipeIdx >= len(m)-1 || pipeIdx < 0 || colonIdx-1 > len(m) || colonIdx < 0 {
```

keitwb:
```suggestion
		if pipeIdx >= len(m)-1 || pipeIdx < 0 || colonId-1 > len(m) || colonIdx < 0 {
```

PR #701, opened 2019-04-09
Add sql and postgres monitors

keitwb:
The `sql` monitor will run queries against a relational database
(PostgreSQL and MySQL currently supported) and derive metrics from the
results.

The `postgres` monitor is mostly a wrapper around the `sql` monitor that
executes standard queries to pull general information about
PostgreSQL server instances.  It is meant to replace the
`collectd/postgresql` monitor.


internal/monitors/postgresql/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+	m.database, err = dbsql.Open("postgres", conf.connStr()+" dbname=postgres")
+	if err != nil {
+		return err
+	}
+
+	var dbFilter filter.StringFilter
+	if len(conf.Databases) > 0 {
+		dbFilter, err = filter.NewExhaustiveStringFilter(conf.Databases)
+		if err != nil {
+			return fmt.Errorf("problem with databases filter: %v", err)
```
jaysfx: Does `Shutdown()` get called to cleanup stuff if `Configure()` fails?

internal/monitors/postgresql/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/signalfx/signalfx-agent/internal/core/config"
+	"github.com/signalfx/signalfx-agent/internal/monitors"
+	"github.com/signalfx/signalfx-agent/internal/monitors/sql"
+	"github.com/signalfx/signalfx-agent/internal/monitors/types"
+	"github.com/signalfx/signalfx-agent/internal/utils"
+	"github.com/signalfx/signalfx-agent/internal/utils/filter"
+	"github.com/sirupsen/logrus"
+)
+
+const monitorType = "postgresql"
```
jaysfx: Use `monitorMetadata.MonitorType`

internal/monitors/sql/monitor.go
```diff
@@ -0,0 +1,353 @@
...
+	"github.com/signalfx/signalfx-agent/internal/utils"
+	"github.com/sirupsen/logrus"
+
+	// Import to get sql driver registered
+	_ "github.com/go-sql-driver/mysql"
+	// Import to get sql driver registered
+	_ "github.com/lib/pq"
+)
+
+const monitorType = "sql"
```
jaysfx: Use `monitorMetadata.MonitorType`

internal/monitors/sql/monitor.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// Configure the monitor and kick off metric gathering
+func (m *Monitor) Configure(conf *Config) error {
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+
+	// This will "open" a database by verifying that the config is sane but
+	// generally won't try and connect to it.  If it does attempt to connect
+	// here this should be done within RunOnInterval to handle cases where the
+	// database is temporaryily down when the monitor starts.
```
jaysfx: temporarily

internal/monitors/sql/monitor.go
```diff
@@ -0,0 +1,353 @@
...
+	m.database, err = sql.Open(conf.DBDriver, dataSource)
+	if err != nil {
+		return fmt.Errorf("Could not handle %s database config: %v", conf.DBDriver, err)
+	}
+
+	for i := range conf.Queries {
+		querier := newQuerier(&conf.Queries[i], conf.LogQueries)
+
+		utils.RunOnInterval(m.ctx, func() {
+			if err := querier.doQuery(m.ctx, m.database, m.Output); err != nil {
```
jaysfx: Is `Output` thread safe?

internal/monitors/postgresql/monitor.go
keitwb: No, good point.  The Configure method should be responsible for closing the DB in the error case. Will update.

internal/monitors/sql/monitor.go
keitwb: Yes, very.

PR #703, opened 2019-04-10
Fix selfdescribe ast cache

jaysfx:
This brings down selfdescribe.json generation from 6s to 1.2s.

### DIFF
--- a/internal/selfdescribe/sourcedocs.go
+++ b/internal/selfdescribe/sourcedocs.go
@@ -12,10 +12,12 @@ import (
 	"strings"
 )
 
-var astCache = make(map[string]struct {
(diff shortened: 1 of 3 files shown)


internal/selfdescribe/sourcedocs.go
```diff
@@ -34,6 +36,7 @@ func structNodes(packageDir, structName string) (*ast.TypeSpec, *ast.CommentGrou
 		if err != nil {
 			panic(err)
 		}
+		astCache[packageDir] = astCacheEntry{fset: fset, pkgs: pkgs}
```
keitwb: Yeah I guess it would help to actually cache it.

PR #708, opened 2019-04-11
Ensure PS scripts exit on failures

jchengsfx:


jaysfx:
Can `$ErrorActionPreference` [be set](https://www.oreilly.com/library/view/professional-windows-powershell/9780471946939/9780471946939_using_the_dollar_erroractionpreference_v.html) to `Stop` globally instead of per-function? I don't really know anything about PowerShell except what Google tells me though.

jchengsfx:
> Can `$ErrorActionPreference` [be set](https://www.oreilly.com/library/view/professional-windows-powershell/9780471946939/9780471946939_using_the_dollar_erroractionpreference_v.html) to `Stop` globally instead of per-function? I don't really know anything about PowerShell except what Google tells me though.

I don't know much about Powershell either, but I noticed that in some azure-ci builds (e.g. https://dev.azure.com/signalfx/signalfx-agent/_build/results?buildId=182), there were failures in `win_bundle_build` that didn't stop/fail the job, even though the default for `ErrorActionPreference` is `Stop` for powershell steps (see https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#powershell). So I'm hoping that some of these changes will help catch these failures.

jaysfx:
Their examples show the powershell commands embedded in the yaml whereas we're launching powershell from within powershell on our own so I wonder if that setting doesn't get passed through

`powershell: 'powershell -command "& { . ./scripts/windows/make.ps1; vet }"'`

I'm assuming Azure launches a powershell process which then launches a second powershell process and maybe that second process doesn't inherit the variable.

jchengsfx:
> Their examples show the powershell commands embedded in the yaml whereas we're launching powershell from within powershell on our own so I wonder if that setting doesn't get passed through
> 
> `powershell: 'powershell -command "& { . ./scripts/windows/make.ps1; vet }"'`
> 
> I'm assuming Azure launches a powershell process which then launches a second powershell process and maybe that second process doesn't inherit the variable.

Possibly. But when I tried executing just `& { . ./scripts/windows/make.ps1; vet }` in the yaml (without `powershell -command`), I get this error:
`The expression after '.' in a pipeline element produced an object that was not valid. It must result in a command name, a script block, or a CommandInfo object.`

scripts/windows/bundle.ps1
```diff
@@ -13,17 +13,19 @@ function download_collectd([string]$collectdCommit, [string]$outputDir="$BUILD_D
...
-    $requirements = Resolve-Path "$scriptDir\..\get-collectd-plugins-requirements.txt"
-    $script = Resolve-Path "$scriptDir\..\get-collectd-plugins.py"
-    $python = "$buildDir\python\python.exe"
-    $env:PYTHONHOME="$buildDir\python"
-    & $python -m pip install -qq -r $requirements
-    & $python $script $collectdPlugins
-    & $python -m pip list
-    # unset the python home enviornment variable
-    Remove-Item Env:\PYTHONHOME
+    try {
```
keitwb: Wrapping the whole thing in `try` and then just rethrowing whatever it catches seems pointless.  Is this to work around some quirk in Powershell?

jchengsfx:
@jaysfx @keitwb Comments should be addressed with latest push. I removed most of the `try/catch/throw` stuff from the previous commit, and the azure pipeline jobs should now exit/fail correctly on error.

scripts/windows/make.ps1
```diff
@@ -145,28 +135,29 @@ function bundle (
...
     go mod tidy
-    if ($lastexitcode -ne 0){ throw }
     go mod vendor
-    if ($lastexitcode -ne 0){ throw }
 }
 
 function vet() {
     compile_deps
+    $ErrorActionPreference = "Continue"
     go vet -mod vendor ./... 2>&1 | Select-String -Pattern "\.go" | Select-String -NotMatch -Pattern "_test\.go" -outvariable gofiles
```
jaysfx: Can you use `-ErrorAction Ignore` for the command instead of having to change the global flag?

jchengsfx: > Can you use `-ErrorAction Ignore` for the command instead of having to change the global flag?

Tried that, but unfortunately it looks like `ErrorAction` can't override `$ErrorActionPreference = Stop`.  See https://github.com/PowerShell/PowerShell/issues/7623#issuecomment-415581590.

jaysfx: Ah, that's too bad. I guess could also wrap in try/catch? 🤷‍♂️ I'm fine with this though.

PR #709, opened 2019-04-11
Allow arbitrary key/value params to sql/postgresql connectionString o…

keitwb:
…ption

This makes it way easier to use remote config with those monitors


internal/monitors/postgresql/monitor.go
```diff
@@ -75,7 +78,12 @@ func (m *Monitor) Configure(conf *Config) error {
 	m.ctx, m.cancel = context.WithCancel(context.Background())
 
 	var err error
```
jaysfx: `err` variable isn't used anymore in this function. (All usages are now with `:=`).

keitwb:
> btw I just noticed that Close() on database can return an error which isn't handled. Maybe log it? Also results in IDE warnings. Would be a good thing for linting to check as well. Can always ignore the error by assigning to _.

I just ignored it because I don't think anybody cares about errors on closing when the monitor is shutting down  and doesn't need the database anymore anyway.

PR #711, opened 2019-04-12
Add 2 mongodb config options to map to collectd config

mstumpfx:
Maps to:
https://github.com/signalfx/collectd-mongodb/blob/master/mongodb.py#L416
https://github.com/signalfx/collectd-mongodb/blob/master/mongodb.py#L418


mstumpfx:
> Enhance the integration tests to ensure these actually work with the plugin properly.

added

PR #714, opened 2019-04-16
Support Redhat image certification process

keitwb:
- Add a new Dockerfile.rhel that uses a RHEL base image and installs
   the agent via its RPM package.
 - Add a release script to maintain a separate branch called
   `redhat-cert` that can be used to build against using the Redhat build
   service.

### DIFF
--- a/scripts/release-redhat
+++ b/scripts/release-redhat
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+set -euo pipefail
+set -x
+
+# Updates the ${build_branch} branch on github with the newer version of the
+# agent baked into the Dockerfile.rhel.  We can't use build args or run any
+# scripts on the RHEL build service so we have to have the Dockerfile be as
+# self-contained as possible.
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+usage() {
+  echo "$0 <version> <project_id> [redhat_tag]"
+}
+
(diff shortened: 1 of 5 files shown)


scripts/release-redhat
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -euo pipefail
+set -x
+
+# Updates the redhat-cert branch on github with the newer version of the agent
```
jaysfx: Protect the branch in github from deletion if you haven't already.

keitwb: It will do a reset/creation of it (the `-B` flag to `git checkout`) when this script is run so it can be deleted at will as long as it doesn't happen before the build starts in Redhat's servers.

scripts/release-redhat
```diff
@@ -0,0 +1,36 @@
...
+  git checkout -B redhat-cert v$version
+  update_dockerfile
+  if ! git diff --exit-code; then
+    git commit -m "Update RHEL Dockerfile version to $version"
+  fi
+  git push -f origin redhat-cert
+}
+
+trigger_build() {
+  echo "Go to https://connect.redhat.com/project/1090391/build-service and click 'Start New Build'"
```
jchengsfx: Do we have a team account for building?

keitwb: We're not sure yet if the partner portal even has a programmatic API at this point.  Right now I'm build going on the website and clicking a button but hopefully there is something better.

PR #721, opened 2019-04-17
Resolve python security vulnerabilities

jchengsfx:


Dockerfile
```diff
@@ -198,24 +198,31 @@ RUN /opt/collect-libs /opt/deps /usr/sbin/collectd /usr/lib/collectd/
...
 RUN apt update &&\
-    apt install -y git python-pip wget curl &&\
-    pip install --upgrade 'pip==18.0'
+    apt install -y wget build-essential libssl-dev zlib1g-dev libexpat1-dev libffi-dev valgrind libcurl4-openssl-dev
+
+RUN wget -O /tmp/Python-${PYTHON_VERSION}.tgz https://www.python.org/ftp/python/${PYTHON_VERSION}/Python-${PYTHON_VERSION}.tgz &&\
+    cd /tmp &&\
+    tar xzf Python-${PYTHON_VERSION}.tgz && \
+    cd Python-${PYTHON_VERSION} && \
+    ./configure --prefix=/usr --enable-shared --enable-ipv6 --enable-unicode=ucs4 --with-valgrind --with-system-ffi --with-system-expat && \
```
keitwb: We don't need valgrind support.

scripts/get-collectd-plugins-security-requirements.txt
```diff
@@ -0,0 +1,3 @@
+PyYAML==5.1
```
keitwb: This file feels very hacky, can't we just upgrade the original plugin's requirements to match this?  You are forcing the plugins to run with these versions anyway so you might as well just change the plugin's requirements.txt to match and avoid this hack.

keitwb: You would also get the benefit of any testing in the plugin repos would be forced to use the newer versions to get deeper testing of them with these newer version.

Dockerfile
jchengsfx: Will remove.

Dockerfile
```diff
@@ -292,16 +301,16 @@ RUN mkdir -p /opt/root/bin &&\
...
-COPY --from=python-plugins /usr/lib/python2.7/ /opt/root/lib/python2.7
-COPY --from=python-plugins /usr/local/lib/python2.7/ /opt/root/lib/python2.7
-COPY --from=python-plugins /usr/bin/python /opt/root/bin/python
-
 # Gather compiled collectd plugin libraries
 COPY --from=collectd /usr/sbin/collectd /opt/root/bin/collectd
 COPY --from=collectd /opt/deps/ /opt/root/lib/
 COPY --from=collectd /usr/lib/collectd/*.so /opt/root/lib/collectd/
 
+# Gather Python dependencies
```
keitwb: The reason you had to move this block below the collectd COPYs is because collectd is still compiling with the old Python 2.7 in Ubuntu's repos. You are also pulling in `libpython.so` from the `collectd` stage first and then overridding it with `libpython.so` from the `python-plugins` stage.  You need to download/compile Python in the first stage and use the same thing in both stages (either move the `python-plugins` stage to above `collectd` and then COPY `libpython.so` and the headers into the `collectd` stage, or do the download/compile in the `collectd` stage and copy it into the `python-plugins` stage).

keitwb: Right now you are linking collectd with one version of Python and then running it against a different version, which is risky.

scripts/get-collectd-plugins-security-requirements.txt
jchengsfx: Yeah, wasn't sure how best to address this since we need to make sure that these versions are installed in case someone pins an earlier version in existing or new plugins.

keitwb:
Were you able to find a tool to confirm this fixes the CVEs?

Dockerfile
jchengsfx: Will try to get the collectd stage to build with the new python. Need to be careful not to conflict with all the other python dependencies installed by apt.

scripts/get-collectd-plugins-security-requirements.txt
keitwb: Ideally we just have any plugin that needs yaml has a requirement of `PyYAML >= 2.0` (or whatever the earliest version it works with is) and then when we you do the `pip install` it will just get the latest version.

A convenient way to see all the plugins' requirements.txt files is to just run a docker container in the layer right after the docker build has pulled the plugins from github.  Then you can see where the existing versions are coming from.

Also we probably don't want to pin to an exact version either, maybe let it use anything with the 5.x major release (for `PyYAML` and similar for others).

Dockerfile
keitwb: If you compile python in the `collectd` stage, you can set the `--prefix` to something like `/opt/python` and do a `make && make install` and it will put it there without conflicts with system Python.  Then just set some flags in the `./configure` for collectd to point to the alternate Python installation.

collectd-plugins.yaml
```diff
@@ -38,7 +38,7 @@
   repo: signalfx/collectd-jenkins
 
 - name: kong
-  version: v0.0.2
+  version: 2c5c967d0d7e23d8e82c406ce59f1d29a44f0c3c
```
jchengsfx: Need to update version for next release.

Dockerfile
```diff
@@ -226,10 +244,12 @@ RUN mkdir -p /opt/collectd-python &&\
 COPY python/ /opt/sfxpython/
 RUN cd /opt/sfxpython && pip install .
 
+RUN pip list
+
 # Delete all compiled python to save space
-RUN find /usr/lib/python2.7 /usr/local/lib/python2.7/dist-packages -name "*.pyc" | xargs rm
+RUN find $PYTHONHOME/lib/python2.7 -name "*.pyc" -o -name "*.pyo" | xargs rm
```
keitwb: I think this was an existing but, but this should delete pyc and compile headers from `/opt/python` for it to have any effect.

jchengsfx: Changed `$PYTHONHOME/lib/python2.7` to just `$PYTHONHOME`.

PR #722, opened 2019-04-17
Cleanup monitorType

jaysfx:
Generate monitorType in genmetadata.

### DIFF
--- a/internal/monitors/expvar/expvar.go
+++ b/internal/monitors/expvar/expvar.go
@@ -15,24 +15,19 @@ import (
 	"github.com/signalfx/signalfx-agent/internal/monitors"
 	"github.com/signalfx/signalfx-agent/internal/monitors/types"
 	"github.com/signalfx/signalfx-agent/internal/utils"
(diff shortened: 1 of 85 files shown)


internal/monitors/expvar/expvar.go
```diff
@@ -58,12 +53,12 @@ type Monitor struct {
 	metricPathsParts    map[*MetricConfig][]string
 	dimensionPathsParts map[*DimensionConfig][]string
 	allMetricConfigs    []*MetricConfig
-	logger              logrus.FieldLogger
+	logger              log.FieldLogger
```
keitwb: I'm trying to move away from doing the `logrus` -> `log` alias as it just confuses `goimports` too much.

jaysfx: Ah. Yeah I was debating which was the intended one. The nice thing about the alias is if we ever switch out logrus we could keep its API and not have to change all the references to logrus but I suppose s/logrus/something-else isn't a huge deal itself.

PR #723, opened 2019-04-18
Cleanup path handling

jaysfx:
Not finished but it's all I could manage.

---
This change is [Reviewable](https://reviewable.io/reviews/signalfx/signalfx-agent/723)


tests/helpers/metadata.py
```diff
@@ -17,9 +16,7 @@ def __init__(self, monitor_type, included_metrics, nonincluded_metrics, dims):
     @classmethod
     def from_package(cls, monitor_package_path, mon_type=None):
         with open(
-            os.path.join(REPO_ROOT_DIR, "internal", "monitors", monitor_package_path, "metadata.yaml"),
-            "r",
-            encoding="utf-8",
+            REPO_ROOT_DIR / "internal" / "monitors" / monitor_package_path / "metadata.yaml", "r", encoding="utf-8"
```
keitwb: nice, didn't know Python had such an operator.

tests/monitors/collectd_postgresql/postgresql_test.py
```diff
@@ -16,6 +17,8 @@
 
 pytestmark = [pytest.mark.collectd, pytest.mark.postgresql, pytest.mark.monitor_with_endpoints]
 
+DIR = Path(__file__).parent
```
keitwb: These would be a lot clearer if they were called `SCRIPT_DIR` or something less generic than `DIR`.

PR #727, opened 2019-04-18
fix expvar for extra metrics

jaysfx:
* add verify helper for checking all metrics


internal/monitors/expvar/expvar.go
```diff
@@ -140,7 +141,9 @@ func (m *Monitor) fetchMetrics() (map[string][]*datapoint.Datapoint, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
+	defer func() {
```
bjsignalfx: Why not `if res != nil { defer res.Body.Close() }` instead, unless you want to log the error returned by `res.Body.Close()`

jaysfx: I just changed it to ignore the return value of `Close()` to get rid of a warning about unchecked err. It _should_ be the same behavior as before but let me know if I'm missing something.

docs/config-schema.md
```diff
@@ -56,6 +56,7 @@ if not set.
 | `metricsToInclude` | no | [list of objects (see below)](#metricstoinclude) | A list of metric filters that will whitelist/include metrics.  These filters take priority over the filters specified in `metricsToExclude`. |
 | `metricsToExclude` | no | [list of objects (see below)](#metricstoexclude) | A list of metric filters |
 | `propertiesToExclude` | no | [list of objects (see below)](#propertiestoexclude) | A list of properties filters |
+| `enableExtraMetricsFilter` | no | bool | Whether to enable new-style metrics filter (**default:** `false`) |
```
keitwb: Does it break the tests horribly if it already defaults to `true`?

internal/core/meta/meta.go
```diff
@@ -6,6 +6,7 @@ package meta
 // access.
 // TODO: get rid of this since it's hacky
 type AgentMeta struct {
-	InternalStatusHost string
-	InternalStatusPort uint16
+	InternalStatusHost       string
+	InternalStatusPort       uint16
+	EnableExtraMetricsFilter bool
```
keitwb: As the comment on this indicates, I don't really want to expand the use of this.  If this is a permanent thing and not just a transitional hack, I would rather find some different to way to do this.

The main problem with AgentMeta as it stands now is that changes to it don't necessarily get observed by monitors, nor is there any notification system for it, so the only robust way to ensure monitors are in sync with this config is to restart all of them if something in AgentMeta gets changed, or add some notification callback system to it.

internal/monitors/expvar/expvar.go
keitwb: The only potential difference between the old and new is that I think it evaluates the value of `resp.Body` when registering the defer in the old way and evaluates it after function return in the anonymous func way.  Since `resp` and `resp.Body` aren't getting reassigned later in the function it doesn't make any difference here.

docs/config-schema.md
keitwb: Sorry, this would never default to true since existing configs won't have it.  I should have said can you go ahead and set it to true in the package configs we have scattered throughout this repo?

internal/core/config/config.go
```diff
@@ -86,6 +86,8 @@ type Config struct {
 	MetricsToExclude []MetricFilter `yaml:"metricsToExclude" default:"[]"`
 	// A list of properties filters
 	PropertiesToExclude []PropertyFilterConfig `yaml:"propertiesToExclude" default:"[]"`
+	// Whether to enable new-style metrics filter
+	EnableExtraMetricsFilter bool `yaml:"enableExtraMetricsFilter" default:"false"`
```
keitwb: Other name ideas:
 - `enableStandardMetricFilters`
 - `enableStandardFilters`
 - `enableCustomMetricFilters`
 - `enableWhitelistFilters`

Main thing against "extra" is that it doesn't correspond to anything we document about how host based pricing works.
 - `

tests/helpers/containers.py
```diff
@@ -0,0 +1,15 @@
+__all__ = ["expvar"]
+
+import pytest
+
+from tests.helpers.assertions import tcp_socket_open
+from tests.helpers.util import run_service, container_ip, wait_for
+
+
+@pytest.yield_fixture
+def expvar():
```
keitwb: This is going to be cleaner and more scalable if it just goes in the package with the expvar monitor tests.  If this file actually had all the containers, it would be gigantic.

Also consider this, if you just have a function like this:

```
def run_expvar(...):
    with run_service("expvar") as expvar_container:
        host = container_ip(expvar_container)
        assert wait_for(lambda: tcp_socket_open(host, 8080), 60), "service didn't start"
        yield host
```

Then you can wrap it as a fixture if you really want to by just doing:

```
@pytest.yield_fixture
def expvar():
    with run_expvar() as host:
        yield host
```

It is much clearer though to see in a test `with run_expvar(...) as host` and know "this runs expvar" than it is to see some magic function parameter that seems to come out of nowhere and there wouldn't even be a clear import in that same file so somebody not intimately familiar with the tests has to grep through the test dir to find that the fixture comes from a central python file in a very indirect way.

Doing `with run_expvar(...) as host` also lets you customize the running of the container (this isn't that uncommon in the existing tests), which fixtures don't apparently allow from the test function.  It is one line longer and one indentation level deeper to use a context manager instead of a fixture, but it is clearer and gives you way more control over the start/stop of the service.

At the very least, please don't shove the core logic of this stuff in a fixture function where it is totally unreusable, but it in a separate context manager that is reusable directly from the tests and then make a wrapping fixture if there is a super simple test case that wants to use fixtures to save a line.

jaysfx: You can pass parameters to fixtures using [fixture factories](https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures). I'll remove the central containers.py and move it into `tests/monitors/expvar/conftest.py`. It could go in the test file itself but then you get name shadowing that makes pylint unhappy. So you do know you have to look in the tests own `conftest.py` but it's localized in the tests own package. I think that should alleviate most concerns.

keitwb: If you use fixture factories, it is just a more convoluted way of doing `with run_expvar(...) as host` since you will still have to make it a context manager to clean up the container.  The only thing a fixture factory would gain in that case is not having to have an explicit import, but it also makes it less clear to developers what is going on, which is my biggest concern about all of the indirection.

jaysfx: The fixture factory doesn't require using a `with` in the test because the fixture handles the teardown part itself. But I agree it doesn't buy you much. I think for cases where we don't need anything special from the fixture and just want a container it's pretty straightforward. Perhaps naming the fixture `expvar_container` instead would make it more explicit. See what you think in the next push.

I agree that there is some magic to it given the nature of pytest fixtures but if all tests adhere to the convention then I don't think it'll negatively impact clarity.

keitwb: Ok, if the factory thing works without having to have a `with` block and extra indentation then it might be worth it to use.  Especially if it allows you to control container start/stop in the test function then it would allow for more advanced tests without having to have a distinct pattern for doing containers.

tests/monitors/expvar/expvar_test.py
```diff
@@ -12,26 +12,26 @@
 METADATA = Metadata.from_package("expvar")
 
 
-def test_expvar_included(expvar):
+def test_expvar_included(expvar_container):
```
keitwb: This is the IP of the container right?  If so, this needs to be called something along those lines because most of the existing tests use the name `*_cont[ainer]` to refer to the docker-py container instance, which is going to be really confusing.

tests/monitors/df/df_test.py
```diff
@@ -1,70 +1,67 @@
...
-            - percent_*
-        """
-    ) as agent:
-        _ = (
-            wait_for(lambda: set(agent.fake_services.datapoints_by_metric) == expected_metrics),
-            "timed out waiting for metrics and/or dimensions!",
-        )
-        assert set(agent.fake_services.datapoints_by_metric) == expected_metrics
-        assert not has_log_message(agent.output.lower(), "error"), "error found in agent output!"
+# def test_df_included():
```
keitwb: This is commented out.

bjsignalfx: Yes, I uncommented and implemented these tests as part of the df monitor update

jaysfx:
Ignore this PR for now, see #743

PR #729, opened 2019-04-19
Actually run the bundle tests

keitwb:
I forgot to add it explicitly to the build job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -308,7 +308,7 @@ jobs:
       package_type:
         default: "rpm"
         type: enum
-        enum: ["rpm", "deb", "bundle"]
+        enum: ["rpm", "deb"]
     environment:
       PACKAGE_TYPE: "<< parameters.package_type >>"
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -332,7 +332,7 @@ jobs:
               echo "export TEST_MARKERS='<< parameters.package_type >> and upgrade'" >> $BASH_ENV
           fi
       - run_pytest:
-          pytest_options: "-n auto -m \"$TEST_MARKERS\" --test-bundle-path ./signalfx-agent-latest.tar.gz"
+          pytest_options: "-n auto -m \"$TEST_MARKERS\" --test-bundle-path=\"./signalfx-agent-latest.tar.gz\""
```
jaysfx: Mostly to play with the new feature but there's also more quoting than necessary I think:
```suggestion
          pytest_options: -n auto -m "$TEST_MARKERS" --test-bundle-path=./signalfx-agent-latest.tar.gz
```

PR #730, opened 2019-04-19
Allow use of K8s annotations in discovery rules

keitwb:
Had to rework the Get helper function used by govaulate in the rules to
make it work with any map type as well.

### DIFF
--- a/docs/observers/k8s-api.md
+++ b/docs/observers/k8s-api.md
@@ -26,6 +26,7 @@ Observer Type: `k8s-api`
 | --- | --- | --- | --- |
 | `namespace` | no | `string` | If specified, only pods within the given namespace on the same node as the agent will be discovered. If blank, all pods on the same node as the agent will be discovered. |
 | `kubernetesAPI` | no | `object (see below)` | Configuration for the K8s API client |
+| `additionalPortAnnotations` | no | `list of strings` | A list of annotation names that should be used to infer additional ports to be discovered on a particular pod.  The pod's annotation value should be a port number.  This is useful for annotations like `prometheus.io/port: 9230`.  If you don't already have preexisting annotations like this, we recommend using the [SignalFx-specific annotations](https://docs.signalfx.com/en/latest/integrations/agent/kubernetes-setup.html#config-via-k8s-annotations). |
 
 
 The **nested** `kubernetesAPI` config object has the following fields:
--- a/internal/core/services/endpointcore.go
+++ b/internal/core/services/endpointcore.go
@@ -38,8 +38,9 @@ type EndpointCore struct {
 	// The type of monitor that this endpoint has requested.  This is populated
 	// by observers that pull configuration directly from the platform they are
 	// observing.
-	MonitorType     string            `yaml:"-"`
-	extraDimensions map[string]string `yaml:"-"`
+	MonitorType     string                 `yaml:"-"`
+	extraDimensions map[string]string      `yaml:"-"`
+	extraFields     map[string]interface{} `yaml:"-"`
 }
 
 // NewEndpointCore returns a new initialized endpoint core struct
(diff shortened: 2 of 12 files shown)


test-services/minikube/start-minikube.sh
```diff
@@ -45,9 +45,6 @@ MINIKUBE_OPTIONS="--vm-driver=none --bootstrapper=${BOOTSTRAPPER} --kubernetes-v
 KUBEADM_OPTIONS="--feature-gates=CoreDNS=true \
     --extra-config=kubelet.authorization-mode=AlwaysAllow \
     --extra-config=kubelet.anonymous-auth=true"
-if [[ "$K8S_VERSION" =~ ^v1\.11\. ]]; then
```
keitwb: @jchengsfx do you remember why this was there?  We shouldn't really be using the cadvisor port for stats anymore anyway.

jchengsfx: In minikube, cadvisor was enabled by default for k8s versions <1.11, disabled by default in 1.11, and then removed in 1.12.  We have a cadvisor test for it (now skipped if testing in k8s >=1.12). If this is now unnecessary, we can delete this and disable/remove the test.

Makefile
```diff
@@ -243,9 +243,42 @@ run-chef-tests:
...
+
+	echo "Minikube is now running. Push up an agent image to minikube-test:5000/signalfx-agent:latest or run 'make push-minikube-agent'"
+
+.PHONY: setup-minikube-host
+.SILENT:
+setup-minikube-host:
+	sudo perl -ni -e 'print unless /minikube-test/' /etc/hosts
+	sudo bash -c "echo '$(shell docker inspect --format '{{.NetworkSettings.IPAddress}}' minikube) minikube-test' >> /etc/hosts"
+	sudo mkdir -p /etc/docker/certs.d/minikube-test\:5000/
+	sudo cp test-services/minikube/minikube-test.crt /etc/docker/certs.d/minikube-test\:5000/ca.crt
```
jchengsfx: Need to include `sudo` package in dev-image.

Makefile
```diff
@@ -243,9 +243,42 @@ run-chef-tests:
...
+		--build-arg 'MINIKUBE_VERSION=$(MINIKUBE_VERSION)' \
+		test-services/minikube
+
+	docker rm -fv minikube 2>/dev/null || true
+	docker run -d \
+		--name minikube \
+		--privileged \
+		-e K8S_VERSION=$(K8S_VERSION) \
+		-e TIMEOUT=$(MINIKUBE_TIMEOUT) \
+		minikube:$(MINIKUBE_VERSION)
```
jchengsfx: Need to bind port 5000:5000 for the minikube registry?

tests/conftest.py
```diff
@@ -86,111 +97,36 @@ def pytest_generate_tests(metafunc):
...
+        ), f"You must start the minikube container ({minikube_container_name}) before running pytest by running `make minikube`."
+
+        # Pull the kubeconfig out of the minikube container and make a cluster instance based on it.
+        minikube_cont = dclient.containers.get(minikube_container_name)
+        _, kubeconfig_bytes = minikube_cont.exec_run("cat /kubeconfig")
+
+        _, kube_config_path = tempfile.mkstemp()
+        request.addfinalizer(lambda: os.remove(kube_config_path))
+
+        with open(kube_config_path, "wb") as fd:
```
jaysfx: `Path("path").write_bytes(kubeconfig_bytes)` is a nice one-liner for this. It also has `read_text()` for reading and `write_text()` for writing.

tests/helpers/kubernetes/cluster.py
```diff
@@ -0,0 +1,164 @@
...
+                for pod in utils.get_all_pods(namespace):
+                    print(self.exec_kubectl(f"describe pod {pod.metadata.name}", namespace=namespace))
+
+    @contextmanager
+    def create_resources(self, yamls, timeout=utils.K8S_CREATE_TIMEOUT):
+        resources = []
+        for yaml_file in yamls:
+            assert os.path.isfile(yaml_file), '"%s" not found!' % yaml_file
+            with open(yaml_file, "r") as fd:
+                for doc in yaml.load_all(fd.read()):
```
jchengsfx: Should use `yaml.safe_load_all()`.

PR #736, opened 2019-04-23
OpenShift cluster monitor

jaysfx:
Implements cluster quota metrics for openshift as a superset of the Kubernetes
cluster monitor.

Adds ability to write generated metadata to seprate package which also exports
the variables for use outside of the local package.

Other quality of life improvements:

* Broke genmetadata.go template out into file so that line numbers in error
  messages were easier to use.

* Added `docker-compose.yml` file to use with `docker-sync`. It's meant to be
  used directly with `docker-compose` instead of through the Makefile. The
  advantage of it is that it uses a volume for the container `HOME`. This means
  when the container exits you don't lose your go build and mod caches. This makes
  it quick to rebase against a new `dev-image` and restart in general.

  See development.md update for more details.

* Added new Kubernetes AuthType that uses credentials and endpoints from
  `~/.kube/config`. This makes it easy to run the agent against a local
  minishift, minikube, or any other cluster you have already configured to use
  with `kubectl`.


.gitignore
```diff
@@ -1,31 +1,33 @@
...
-local-etc/
-/agent
-/.docker-sync
-internal/monitors/**/template.go
-internal/monitors/**/genmetadata.go
-internal/core/common/constants/versions.go
-internal/monitors/collectd/collectd.conf.go
-/signalfx-agent
-/monitor-code-gen
+**boxcutter/
```
jaysfx: This is jumbled due to sorting it.

internal/monitors/kubernetes/cluster/metadata.yaml
```diff
@@ -141,3 +143,124 @@ monitors:
...
     name: <pod label>
+- <<: *kubernetes
+  monitorType: openshift-cluster
+  sendAll: true
+  dimensions:
+  - description: This is always set to `openshift`
+    name: metric_source
+# NOTE: Do not manually update list below. Update gen-metrics.py and rerun.
+  metrics:
+  - description: Hard limit for number of cpu across all namespaces
```
jaysfx: Right now this results in only the openshift metrics instead of the combination of kubernetes and openshift metrics. Apparently [it's not possible](https://github.com/yaml/yaml/issues/35) to merge yaml arrays with anchors and pointers best I can tell. Maybe just link to the kubernetes monitor? Otherwise I think we will have to do some templating or something. Or the gen script can read the kubernetes metrics and print those out with the generated ones. I dunno.

internal/monitors/kubernetes/cluster/monitor.go
```diff
@@ -7,25 +7,19 @@
 // watch API is used to efficiently maintain the state between read intervals
 // (see `clusterstate.go`).
 //
-// This plugin should only be run at one place in the cluster, or else metrics
```
jaysfx: Looks like half the documentation was removed and it's in the actual docs now so I removed the rest.

internal/monitors/kubernetes/cluster/metadata.yaml
keitwb: Or just make the openshift monitor be specific to OpenShift components and not try and overlap with the kubernetes-cluster monitor?

internal/monitors/kubernetes/gen-metrics.py
```diff
@@ -0,0 +1,36 @@
...
+                "description": description,
+                "included": False,
+                "name": f"{metricPrefix}.{resource}.{category}",
+                "type": "gauge",
+            }
+
+METRICS += build_metrics("openshift.clusterquota", "across all namespaces")
+METRICS += build_metrics("openshift.appliedclusterquota", "by namespace")
+
+yaml.safe_dump({"metrics": METRICS}, sys.stdout, encoding='utf-8', default_flow_style=False)
```
keitwb: Not really liking this.  If you really want to map the openshift monitor a superset of kubernetes-cluster, can't you just add another field to the metadata.yaml like `extendMetricsFrom: kubernetes-cluster` or something and have the monitor-code-gen thing read that and do the right thing?

internal/monitors/kubernetes/cluster/metadata.yaml
jaysfx: Seems a little janky. It means if you want to use both openshift-cluster and kubernetes-cluster you have to configuration both separately. Could reuse same configuration with yaml anchors but feels clunky.

internal/monitors/kubernetes/gen-metrics.py
jaysfx: This script just generates the openshift-specific metrics. Yeah I think the only thing to do is do extendMetricsFrom. The downside being any consumer of the yaml files (like the tests) has to know about it and do populate the extended metrics.

jaysfx: Hmm, would it be possible to use the `#from` thing? I'm not entirely sure how to reference the other metrics in the value. But one idea...

```yaml
  - {"#from": "Monitors['kubernetes-cluster']['metrics']", default: [], flatten: true}
  - description: Consumed number of pods across all namespaces
    included: false
    name: openshift.clusterquota.pods.used
    type: gauge
...
```

internal/monitors/kubernetes/cluster/metadata.yaml
keitwb: I wouldn't be too concerned about the config because 99% of the time people rely on standard k8s mechanisms (e.g. a service account) to do the authentication, which is the default.

internal/monitors/kubernetes/gen-metrics.py
keitwb: not very easily, that is a pretty complicated mechanism and it's tied in pretty closely to the agent config loader.

internal/monitors/kubernetes/cluster/metrics/openshiftquotas.go
```diff
@@ -0,0 +1,72 @@
...
+	v1.ResourceServices,
+	v1.ResourcePersistentVolumeClaims,
+	v1.ResourceServicesNodePorts,
+	v1.ResourceServicesLoadBalancers,
+}
+
+func datapointsForClusterQuotas(quota *quota.ClusterResourceQuota) []*datapoint.Datapoint {
+	dimensions := map[string]string{
+		"metric_source": "openshift",
+		"uid":           string(quota.UID),
```
mstumpfx: I think probably should be `kubernetes_uid` but will defer to @keitwb 
We use `uid` on daemonsets and replication controllers, but `kubernetes_uid` on deployments and replicasets. And pods have their own `kubernetes_pod_uid`.

jaysfx: Should any of them use uid instead of kubernetes_uid? Is there a difference between them?

keitwb: Yes, let's stick with `kubernetes_uid`.

docs/monitors/openshift-cluster.md
```diff
@@ -0,0 +1,141 @@
...
+configured will decide amongst themselves which should send metrics for this
+monitor, and the rest will stand by ready to activate if the leader agent
+dies.  You can override leader election by setting the config option
+`alwaysClusterReporter` to true, which will make the monitor always report
+metrics.
+
+This monitor is similar to
+[kube-state-metrics](https://github.com/kubernetes/kube-state-metrics), and
+sends many of the same metrics, but in a way that is less verbose and better
+fitted for the SignalFx backend.
```
keitwb: Can you add a note about how it is related to the `kubernetes-cluster` monitor and how they *don't* need to run both?  I can imagine people seeing both and thinking they complement each other instead of being exclusive.

PR #738, opened 2019-04-24
Filtering docker/ecs

seonsfx:
WIP until `expvar` PR gets merged.

### DIFF
--- a/cmd/monitorcodegen/main.go
+++ b/cmd/monitorcodegen/main.go
@@ -75,6 +75,7 @@ func generate(templateFile string, force bool) error {
 				return "", fmt.Errorf("unknown metric type %s", metricType)
 			}
 		},
+		"deref": func(p *string) string { return *p },
 	}).ParseFiles(templateFile)
 
 	if err != nil {
--- a/internal/monitors/docker/docker.go
+++ b/internal/monitors/docker/docker.go
@@ -86,7 +86,7 @@ type dockerContainer struct {
 func (m *Monitor) Configure(conf *Config) error {
 	m.logger = logrus.WithFields(logrus.Fields{"monitorType": monitorType})
 
-	enhancedMetricsConfig := EnableExtraGroups(conf.EnhancedMetricsConfig, conf.ExtraGroups, conf.ExtraMetrics)
+	enhancedMetricsConfig := EnableExtraGroups(conf.EnhancedMetricsConfig, conf.EnabledMetrics)
 
 	defaultHeaders := map[string]string{"User-Agent": "signalfx-agent"}
 
@@ -224,20 +224,17 @@ func parseContainerEnvSlice(env []string) map[string]string {
 
 // EnableExtraGroups enables extra metrics that were individually turned on
 // by ExtraMetrics/ExtraGroups configuration
-func EnableExtraGroups(initConf EnhancedMetricsConfig, extraGroups, extraMetrics []string) EnhancedMetricsConfig {
+func EnableExtraGroups(initConf EnhancedMetricsConfig, enabledMetrics []string) EnhancedMetricsConfig {
 	groupEnableMap := map[string]bool{
-		groupBlkio: initConf.EnableExtraBlockIOMetrics,
-		groupCPU: initConf.EnableExtraCPUMetrics,
-		groupMemory: initConf.EnableExtraMemoryMetrics,
+		groupBlkio:   initConf.EnableExtraBlockIOMetrics,
+		groupCPU:     initConf.EnableExtraCPUMetrics,
+		groupMemory:  initConf.EnableExtraMemoryMetrics,
 		groupNetwork: initConf.EnableExtraNetworkMetrics,
 	}
 
-	for _, group := range extraGroups {
-		groupEnableMap[group] = true
-	}
-	for _, metric := range extraMetrics {
-		if metricSet[metric].Group != "" {
-			groupEnableMap[metricSet[metric].Group] = true
+	for _, metric := range enabledMetrics {
+		if metricInfo, ok := metricSet[metric]; ok {
+			groupEnableMap[metricInfo.Group] = true
 		}
 	}
 
--- a/internal/monitors/docker/docker.go
+++ b/internal/monitors/docker/docker.go
@@ -86,6 +86,8 @@ type dockerContainer struct {
 func (m *Monitor) Configure(conf *Config) error {
 	m.logger = logrus.WithFields(logrus.Fields{"monitorType": monitorType})
 
+	enhancedMetricsConfig := EnableExtraGroups(conf.EnhancedMetricsConfig, conf.ExtraGroups, conf.ExtraMetrics)
+
 	defaultHeaders := map[string]string{"User-Agent": "signalfx-agent"}
 
 	var err error
@@ -150,12 +152,7 @@ func (m *Monitor) Configure(conf *Config) error {
 		// only the map that holds them.
 		lock.Lock()
 		for id := range containers {
-			go m.fetchStats(containers[id], conf.LabelsToDimensions, conf.EnvToDimensions, EnhancedMetricsConfig{
-				EnableExtraBlockIOMetrics: conf.EnableExtraBlockIOMetrics,
-				EnableExtraCPUMetrics:     conf.EnableExtraCPUMetrics,
-				EnableExtraMemoryMetrics:  conf.EnableExtraMemoryMetrics,
-				EnableExtraNetworkMetrics: conf.EnableExtraNetworkMetrics,
-			})
+			go m.fetchStats(containers[id], conf.LabelsToDimensions, conf.EnvToDimensions, enhancedMetricsConfig)
 		}
 		lock.Unlock()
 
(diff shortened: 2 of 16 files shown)


internal/monitors/docker/docker.go
```diff
@@ -225,9 +227,66 @@ func parseContainerEnvSlice(env []string) map[string]string {
...
+	groupNames := []string{groupBlkio, groupCPU, groupMemory, groupNetwork}
+
+	for _, group := range extraGroups {
+		for _, gn := range groupNames {
+			if group == gn {
+				groupEnableMap[gn] = true
+			}
+		}
+	}
+	for _, metric := range extraMetrics {
```
keitwb: Ideally this would be a helper method somewhere (either directly in generated code or common package that takes generated variables as argumnets) and also ideally we could do the check in O(1) for each metric, esp for monitors with lots of groups like elasticsearch. @jaysfx

internal/monitors/docker/docker.go
```diff
@@ -225,9 +227,66 @@ func parseContainerEnvSlice(env []string) map[string]string {
...
 
+// EnableExtraGroups enables extra metrics that were individually turned on
+// by ExtraMetrics/ExtraGroups configuration
+func EnableExtraGroups(groupEnableMap map[string]bool, extraMetrics, extraGroups []string) EnhancedMetricsConfig {
+	groupNames := []string{groupBlkio, groupCPU, groupMemory, groupNetwork}
+
+	for _, group := range extraGroups {
+		for _, gn := range groupNames {
+			if group == gn {
+				groupEnableMap[gn] = true
```
keitwb: Modifying the input param is pretty tricky and not really recommended unless you are going for some super optimization here.

keitwb: Also, why can't this just be:
```
for _, group := range extraGroups {
    groupEnableMap[group] = true
}
```

If it is a non-existent group, that should be handled somewhere else (e.g. some kind of validation).

seonsfx: was trying not to rely on user provided input for the keys, but it looks better to keep validation somewhere like you suggested.

seonsfx: Please correct me if I'm wrong, but here I'm not modifying input params but just taking those as a default set of values. See https://github.com/signalfx/signalfx-agent/blob/aa3d3acbe4c7b207477ce01568f2fd86e7914cc5/internal/monitors/docker/docker.go#L90

internal/monitors/docker/docker.go
seonsfx: What do we think about having auto-generated `metricGroupMap` as well as `groupMetricsMap`?

seonsfx: What we really should do, IMO, is to add `MetricInfo.Group` so we can get a group name for a metric in O(1)

internal/monitors/docker/docker.go
keitwb: Well, you are modifying `groupEnableMap` which is pass in as an arg.  Since this is specific to this monitor, I would just pass in conf instead and then derive the "default set of values" from conf in this function so it is more self-contained.

internal/monitors/docker/docker.go
keitwb: > What we really should do, IMO, is to add `MetricInfo.Group` so we can get a group name for a metric in O(1). `MetricInfo` looks like a place we would have this kind of info. @keitwb @jaysfx

yeah that seems like the easiest option.

keitwb: If we want to be exhaustive about this, we also need to handle the case where `extraMetrics` has a glob/regex in it by iterating every metric that might match it.  @jaysfx

internal/monitors/filesystems/filesystems.go
```diff
@@ -243,6 +243,20 @@ func (m *Monitor) Shutdown() {
 	}
 }
 
+// GetExtraMetrics returns additional metrics that should be allowed through.
+func (c *Config) GetExtraMetrics() []string {
+	if c.ReportInodes {
+		return []string{
+			"df_inodes.free",
```
keitwb: This seems like it overlaps with #741 .

internal/monitors/docker/docker.go
seonsfx: Got it. Thought you meant `input param` of the monitor not the function.

cmd/monitorcodegen/main.go
```diff
@@ -58,7 +58,11 @@ const (
 
 var metricSet = map[string]monitors.MetricInfo {
 {{- range .metrics}}
-	{{.Name | formatVariable}}: {Type: {{.Type | convertMetricType}}},
+	{{- if .Group}}
+		{{.Name | formatVariable}}: {Type: {{.Type | convertMetricType}}, Group: {{.Group | deref | printf "group.%s" | formatVariable}}},
```
seonsfx: Had to defef `.Group` here, but couldn't really figure out why `.Group` is shown as a pointer in printf statement.

internal/monitors/docker/docker.go
jaysfx: Yeah adding `Group` to `MetricInfo` makes sense.

I improved the enabled metrics list in the [this PR](https://github.com/signalfx/signalfx-agent/pull/743/files#diff-71a16ae4b4cab4ccd7b59767ef22b69bR358) (previously called expvar). It runs all the known metrics across the filterset (which includes both extra metrics and excluded datapoints). This would be ideal to use I think since in addition to handling globs if datapoints have been excluded there's no need to collect those metrics/enable relevant groups.

So maybe this computed `enabledMetrics` can be available to `Configure` (set it on the `MonitorConfig` struct or monitor itself?)

internal/monitors/filesystems/filesystems.go
seonsfx: @jaysfx can you confirm this?

jaysfx: This is filesystems plugin whereas #741 is df plugin but I guess they have same/similar metrics. Not sure if there's an easy way to share this though.

seonsfx: df/filesystem should not be dependent on each other IMO

jaysfx: Yeah I think that makes sense but the groupings should be the same since the metrics are basically duplicated. Maybe punt on filesystems plugin until the df one is merged so that they're both done the same way.

internal/monitors/docker/docker.go
```diff
@@ -67,6 +68,17 @@ type Config struct {
 	ExcludedImages []string `yaml:"excludedImages"`
 }
 
+// Validate docker-container-stats monitor config
+func (c *Config) Validate() error {
+	for _, group := range c.ExtraGroups {
```
jaysfx: This is validated already by the extra metrics filter (actually we don't error if the group name is invalid, we just warn per @keitwb) so I don't think we need to do it here. But it does mean that we can't assume that the groups in `c.ExtraGroups` are valid anywhere else.

internal/monitors/docker/docker.go
```diff
@@ -225,9 +234,63 @@ func parseContainerEnvSlice(env []string) map[string]string {
 	return out
 }
 
+// EnableExtraGroups enables extra metrics that were individually turned on
+// by ExtraMetrics/ExtraGroups configuration
+func EnableExtraGroups(initConf EnhancedMetricsConfig, extraGroups, extraMetrics []string) EnhancedMetricsConfig {
```
jaysfx: Looks like this can be unexported?

internal/monitors/docker/docker.go
```diff
@@ -225,9 +234,63 @@ func parseContainerEnvSlice(env []string) map[string]string {
...
+		groupMemory: initConf.EnableExtraMemoryMetrics,
+		groupNetwork: initConf.EnableExtraNetworkMetrics,
+	}
+
+	for _, group := range extraGroups {
+		groupEnableMap[group] = true
+	}
+	for _, metric := range extraMetrics {
+		if metricSet[metric].Group != "" {
+			groupEnableMap[metricSet[metric].Group] = true
```
jaysfx: We need to do the thing [I referenced above](https://github.com/signalfx/signalfx-agent/pull/738/files#r278372837) so that it handles globs and (and excluded metrics for free). @keitwb thoughts on the implementation details of it?

internal/monitors/docker/docker.go
```diff
@@ -225,9 +234,63 @@ func parseContainerEnvSlice(env []string) map[string]string {
...
+	if c.EnableExtraMemoryMetrics {
+		extraMetrics = append(extraMetrics, groupMetricsMap[groupMemory]...)
+	}
+
+	if c.EnableExtraNetworkMetrics {
+		extraMetrics = append(extraMetrics, groupMetricsMap[groupNetwork]...)
+	}
+	
+	logger := logrus.WithFields(logrus.Fields{"monitorType": "GetExtraMetrics"})
+	logger.WithFields(logrus.Fields{"extraMetrics":extraMetrics}).Debug("Extra metrics")
```
jaysfx: Not sure it should be logged per-monitor like this, I'll let @keitwb chime in.

internal/monitors/docker/metadata.yaml
```diff
@@ -19,287 +19,351 @@ monitors:
     included: false
     name: blkio.io_service_bytes_recursive.async
     type: cumulative
+    group: blkio
```
jaysfx: Also need `groups` section to document the groups.

internal/monitors/ecs/metadata.yaml
```diff
@@ -10,134 +10,167 @@ monitors:
     included: false
     name: blkio.io_service_bytes_recursive.async
     type: cumulative
+    group: blkio
```
jaysfx: `groups` section docs

internal/monitors/docker/docker.go
seonsfx: oops this is just a blip. was for debugging. getting rid of this

internal/monitors/docker/docker.go
seonsfx: This need to be used by ecs monitor.

tests/monitors/docker_container_stats/docker_container_stats_test.py
```diff
@@ -5,10 +5,15 @@
...
 from tests.helpers.agent import Agent
 from tests.helpers.assertions import has_datapoint_with_dim, has_datapoint_with_metric_name
+from tests.helpers.metadata import Metadata
 from tests.helpers.util import ensure_always, run_service, wait_for
+from tests.helpers.verify import verify_included_metrics, verify_all_metrics
 
 pytestmark = [pytest.mark.docker_container_stats, pytest.mark.monitor_without_endpoints]
 
+METADATA = Metadata.from_package("docker")
+METADATA.all_metrics = METADATA.all_metrics - {"memory.stats.swap"}
```
jaysfx: I don't think it's a good idea to change `METADATA.all_metrics` as there's also `METADATA.nonincluded_metrics` which won't be correct (even if it might not matter for this particular test). I'd just set it to a different variable and instead of calling `verify_all_metrics` call `verify` and pass that variable to verify.

internal/monitors/docker/metadata.yaml
```diff
@@ -14,292 +14,365 @@ monitors:
...
       description: Total amount of inactive file cache memory. Cache memory = active_file
         + inactive_file + tmpfs
       included: false
       type: gauge
+      group: memory
     memory.stats.total_mapped_file:
-      description: Total amount of memory mapped by the processes in the control group.
-        It doesn’t give you information about how much memory is used; it rather tells
-        you how it is used.
+      description: "Total amount of memory mapped by the processes in the control group.\
```
jaysfx: Looks like a number of these got mangled. Shouldn't need escape sequences for apostrophe. Should probably convert them to using `>` if they're long enough.

tests/monitors/docker_container_stats/docker_container_stats_test.py
```diff
@@ -153,3 +155,33 @@ def test_docker_stops_watching_destroyed_containers():
...
+        """,
+        METADATA,
+    )
+
+
+ENHANCED_METRICS = METADATA.all_metrics - {"memory.stats.swap"}
+
+
+def test_docker_enhanced():
+    with Agent.run(
```
jaysfx: Can you add a new function to verify module?

```python
def verify_custom(config, metrics, timeout=util.DEFAULT_TIMEOUT):
    with Agent.run(config) as agent:
        verify(agent, metrics, timeout=timeout)
```

PR #741, opened 2019-04-24
Monitor df metric filtering

bjsignalfx:


docker-sync.yml
```diff
@@ -38,3 +38,11 @@ syncs:
     #
     # for more see http://www.cis.upenn.edu/~bcpierce/unison/download/releases/stable/unison-manual.html#pathspec
     sync_excludes_type: none
+
```
keitwb: This overlaps with another PR.

docs/config-schema.md
```diff
@@ -57,6 +57,7 @@ if not set.
 | `metricsToInclude` | no | [list of objects (see below)](#metricstoinclude) | A list of metric filters that will whitelist/include metrics.  These filters take priority over the filters specified in `metricsToExclude`. |
 | `metricsToExclude` | no | [list of objects (see below)](#metricstoexclude) | A list of metric filters |
 | `propertiesToExclude` | no | [list of objects (see below)](#propertiestoexclude) | A list of properties filters |
+| `enableExtraMetricsFilter` | no | bool | Whether to enable new-style metrics filter (**default:** `false`) |
```
keitwb: That commit should be added to filtering first before this PR.

internal/monitors/collectd/df/df.go
```diff
@@ -49,6 +49,18 @@ type Monitor struct {
 	collectd.MonitorCore
 }
 
+// GetExtraMetrics returns additional metrics to allow through.
+func (c *Config) GetExtraMetrics() []string {
+	var extraMetrics []string
+	if c.ReportInodes {
+		extraMetrics = append(extraMetrics, groupMetricsMap[groupReportInodes]...)
+	}
+	if c.ValuesPercentage {
```
jaysfx: I believe there's a a set of metrics that are only emitted when ReportInodes and ValuesPercentage are both true (percent_inodes.*). If that's true then I think we'll just have to hardcode some metrics here (use the generated constants in genmetadata.go though).

internal/monitors/collectd/df/metadata.yaml
```diff
@@ -20,37 +25,46 @@ monitors:
...
   - description: Free inodes on the file system, expressed as a percentage.
     included: false
     name: percent_inodes.free
     type: gauge
+    group: ValuesPercentage
   - description: Reserved inodes on the file system, expressed as a percentage.
     included: false
     name: percent_inodes.reserved
     type: gauge
+    group: ValuesPercentage
```
jaysfx: Ben and I discussed this before and I think we decided that for metrics thats appear to fall into two groups we would choose the primary one. In this case it seems like inodes is the primary one.

internal/monitors/collectd/df/metadata.yaml
```diff
@@ -20,26 +25,32 @@ monitors:
     included: false
     name: df_inodes.free
     type: gauge
+    group: ReportInodes
```
jaysfx: Let's make these camel case so `reportInodes` just for consistency with some other configs I had done.

jaysfx:
I pushed some changes to the branch. See what you think. Also update all the tests to use `verify_*` functions like I did for some of them.

jaysfx:
@keitwb I think we can probably come up with a more generic way to handle flags implying metrics and metrics implying flags so we don't have to repeat all this on every monitor. But I haven't given much thought to it yet. Gets messy when you have metrics that are emitted when both flags are set. Or the reverse where if you enable a certain metric both flags have to get turned on.

internal/monitors/collectd/df/df.go
```diff
@@ -49,7 +49,36 @@ type Monitor struct {
...
+	conf := *config
+	// Setting group flags in conf for enable extra metrics
+	for _, metric := range conf.EnabledMetrics {
+		switch metricSet[metric].Group {
+		case groupReportInodes:
+			conf.ReportInodes = true
+		case groupValuesPercentage:
+			conf.ValuesPercentage = true
+		default:
+			conf.ReportInodes = true
```
jaysfx: `df_complex.free` and others don't have groups so they'd end up enabling both config flags I think.

tests/monitors/df/df_test.py
```diff
@@ -1,67 +1,105 @@
...
+def test_df_percent_prefixed_metrics():
+    expected_metrics = METADATA.included_metrics | {
+        "percent_bytes.free",
+        "percent_bytes.reserved",
+        "percent_bytes.used",
+        "percent_inodes.free",
+        "percent_inodes.reserved",
+        "percent_inodes.used",
+    }
+    percent_prefixed_wildcard = "percent_*"
```
jaysfx: Doesn't need to be a variable since it's just being included in the config.

tests/monitors/df/df_test.py
```diff
@@ -1,67 +1,105 @@
...
+            extraMetrics:
+            - {percent_prefixed_wildcard}
+        """,
+        expected_metrics,
+    )
+
+
+def test_df_invalid_extra_metric():
+    expected_metrics = METADATA.included_metrics
+    invalid_extra_metric = "Y2W8OBrdZZ"
```
jaysfx: Doesn't need to be a variable since it's just being included in the config. I don't think this needs to be a test case here since it's not df monitor specific.

tests/monitors/df/df_test.py
```diff
@@ -1,67 +1,105 @@
...
+        """,
+        expected_metrics,
+    )
 
 
 def test_df_extra_metrics_all():
     expected_metrics = METADATA.all_metrics
-    with Agent.run(
-        """
+    any_wildcard_in_quotes = "'*'"
```
jaysfx: Doesn't need to be a variable since it's just being included in the config.

tests/monitors/df/df_test.py
```diff
@@ -1,67 +1,105 @@
...
+        """,
+        expected_metrics,
+    )
+
+
+# TODO: Fix failing test
+@pytest.mark.skip(reason="failing due to bug")
+def test_df_blank_extra_metric():
+    expected_metrics = METADATA.included_metrics
+    blank_extra_metric = " "
```
jaysfx: Doesn't need to be a test case for df monitor.

internal/core/config/monitor.go
```diff
@@ -73,11 +73,12 @@ type MonitorConfig struct {
...
-	BundleDir       string          `yaml:"-" json:"-"`
-	ValidationError string          `yaml:"-" json:"-" hash:"ignore"`
-	MonitorID       types.MonitorID `yaml:"-" hash:"ignore"`
-	EnabledMetrics  []string        `yaml:"-" json:"-" hash:"ignore"`
+	Hostname          string          `yaml:"-" json:"-"`
+	BundleDir         string          `yaml:"-" json:"-"`
+	ValidationError   string          `yaml:"-" json:"-" hash:"ignore"`
+	MonitorID         types.MonitorID `yaml:"-" hash:"ignore"`
+	EnabledMetrics    []string        `yaml:"-" json:"-" hash:"ignore"`
+	EnabledMetricsSet map[string]bool `yaml:"-" json:"-" hash:"ignore"`
```
jaysfx: Can remove this if you don't end up using it.

tests/monitors/df/df_test.py
bjsignalfx: I think it's more readable that way

tests/monitors/df/df_test.py
bjsignalfx: I think it's more readable this way

tests/monitors/df/df_test.py
bjsignalfx: More readable this way

internal/core/config/monitor.go
bjsignalfx: Yes, I'll remove it

tests/monitors/df/df_test.py
bjsignalfx: The monitor is supposed to behave as for invalid extra metric. In that case a warning is issued and included metrics are sent according to @keitwb. But no metrics are sent in this case.

internal/core/config/monitor.go
keitwb: Please remove this for sure, optimizations shouldn't really exist on the config structs, do that somewhere else if it really matters.

internal/monitors/collectd/df/metadata.yaml
```diff
@@ -20,26 +20,32 @@ monitors:
       description: Number of inodes that are free.
       included: false
       type: gauge
+      group: reportInodes
```
jaysfx: I just remembered, we decided on kebab case so these should be report-inodes and values-percentage.

tests/monitors/df/df_test.py
```diff
@@ -1,67 +1,81 @@
...
         monitors:
           - type: collectd/df
             hostFSPath: /
             valuesPercentage: true
+        """
+    verify_custom(agent_config, expected_metrics)
+
+
+def test_df_report_inodes_and_values_percentage_flags():
+    expected_metrics = set(METADATA.all_metrics)
```
jaysfx: `expected_metrics = METDATA.all_metrics - {"df_complex.reserved"}`

tests/monitors/df/df_test.py
```diff
@@ -1,67 +1,81 @@
...
-        assert set(agent.fake_services.datapoints_by_metric) == expected_metrics
-        assert not has_log_message(agent.output.lower(), "error"), "error found in agent output!"
+    verify_custom(agent_config, expected_metrics)
 
 
-def test_df_extra_metrics_all():
-    expected_metrics = METADATA.all_metrics
-    with Agent.run(
+def test_df_report_inodes_flag():
+    expected_metrics = METADATA.included_metrics | {"df_inodes.free", "df_inodes.reserved", "df_inodes.used"}
```
jaysfx: Can use `METADATA.metrics_by_group["report-inodes"]` (though there's something to be said for testing that the groups are correct in the metadata by explicitly listing the metrics). @keitwb what do you think?

internal/monitors/collectd/df/metadata.yaml
```diff
@@ -20,26 +20,32 @@ monitors:
...
     df_inodes.reserved:
       description: Number of inodes reserved for the super user.
       included: false
       type: gauge
+      group: reportInodes
```
keitwb: Just call these groups `inodes` and `percentage`.  There is no need to perpetuate the collectd config field names.

tests/monitors/df/df_test.py
keitwb: Yeah nothing wrong with enumerating them explicitly as an extra thorough check.

keitwb:
> @keitwb I think we can probably come up with a more generic way to handle flags implying metrics and metrics implying flags so we don't have to repeat all this on every monitor. But I haven't given much thought to it yet. Gets messy when you have metrics that are emitted when both flags are set. Or the reverse where if you enable a certain metric both flags have to get turned on.

Yeah nothing really stands out to me as a generic enough solution.  Ideally new monitors going forward won't have any of these special flags and will just rely on extraMetrics/Groups to alter the internal behavior of the monitor.

PR #743, opened 2019-04-24
Extra metrics fixes

jaysfx:
* Add helpers for verifying metrics
* Fix GetExtraMetrics()

### DIFF
--- a/tests/monitors/expvar/expvar_test.py
+++ b/tests/monitors/expvar/expvar_test.py
@@ -1,38 +1,40 @@
 """
 Tests for the expvar monitor
 """
-from functools import partial as p
-from textwrap import dedent
 
 import pytest
 
-from tests.helpers.agent import Agent
-from tests.helpers.assertions import has_datapoint, tcp_socket_open
 from tests.helpers.metadata import Metadata
-from tests.helpers.util import container_ip, run_service, wait_for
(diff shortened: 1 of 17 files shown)


tests/monitors/expvar/expvar_test.py
```diff
@@ -1,38 +1,40 @@
...
-             port: 8080
-         """
-            )
-        ) as agent:
-            for metric in METADATA.included_metrics:
-                print("Waiting for %s" % metric)
-                assert wait_for(
-                    p(has_datapoint, agent.fake_services, metric_name=metric)
-                ), "Didn't get included datapoints"
+def test_expvar_included(expvar_container_ip):
```
jaysfx: I renamed this to `expvar_container_ip` for now based on previous PR comments. My original thinking was to originally have a `Container` class to wrap common docker stuff and get rid of that `container_ip` helper function.

keitwb:
This does seem to have broken some basic integration tests (the appmesh one should pass now with latest master).

jaysfx:
> In case you aren't aware, you don't have to use `conftest.py` for fixtures -- you can put them in any file.

Yeah only reason I did was because pylint is unhappy that the variable name `expvar_container_ip` gets shadowed.

Issue #744, opened 2019-04-24
Questions about python monitor


rklabs:
Hello,

I had a couple of questions about python monitor in signalfx. 

1. `pip` is not installed as part of python dist which is installed along with `signalfx-agent` which is understandable since installing new packages or upgrading existing packages may have unintended side effects. I wanted to understand the motivation for not including `pip`. `pip` would actually help us in installing packages which are required for our custom checks.

2. Why are the modules in https://github.com/signalfx/signalfx-agent/tree/v4.6.0/python not packaged and published to pypi? We could build python wheel and distribute via private artifactory or clone the repo and add the package to `sys.path` but that is an ugly hack. If we could install signalfx packages mentioned above via `pip` that would be ideal.

Thanks,
Raju Kadam

keitwb:
Hi @rklabs,

1. The `pip` binary isn't there but the pip module is there.  You can run `PYTHONHOME=/ python -m pip` in the Docker image _OR_ `PYTHONHOME=/usr/lib/signalfx-agent python -m pip` if running it straight from the deb/rpm package installation.  The main limitation of doing pip installs is that any installs that require C-extension compilation probably won't work, but that is mitigated by the prevalence of wheels.  

If you have dependencies that require extension compilation or you want to use a specific version of Python, you can install Python separately on your host and use the `pythonBinary` config option to the [python-monitor](https://github.com/signalfx/signalfx-agent/blob/v4.6.0/docs/monitors/python-monitor.md) (if you are installing the agent on a host and not running the container image).  There is no such config for the `collectd/python` monitor because Python is linked directly in the collectd binary but you should really consider using the `python-monitor` monitor to extend the agent anyway since it is much more natural to our datapoint format.

2. We can definitely publish that as a package to Pypi if it makes things easier for you.  I'm assuming the main reason you want it there is for development/IDE support?  Or is it something else?

Let me know any other feedback/issues on the Python stuff -- this is still pretty new so it has a lot of room for maturation.

rklabs:
Thanks for the quick reply!

1. Thanks for the tip! I was looking for `pip` executable in `bin` directory of `signalfx-agent` installation. But this works as well and currently, we don't really need C-extension for python.

[...]

PR #745, opened 2019-04-25
Prepare relasing the python extension code to Pypi

keitwb:
As requested in #744.


rklabs:
Thanks for doing this! Is `sfxpython` pushed to pypi? `pip search` gives no results.

keitwb:
Not pushed yet, I'm still waiting to get credentials to Pypi for our company org there but should have it soon and will let you know.

PR #746, opened 2019-04-25
Allow use of existing cluster in pytest k8s tests

keitwb:


tests/monitors/mongodb/mongodb_test.py
```diff
@@ -63,28 +55,3 @@ def test_mongo_enhanced_metrics():
             assert wait_for(
                 p(has_datapoint, agent.fake_services, metric_name="counter.collection.commandsTime"), 60
             ), "Did not get datapoint from SendCollectionTopMetrics config"
-
-
-@pytest.mark.kubernetes
-def test_mongodb_in_k8s(agent_image, minikube, k8s_observer, k8s_test_timeout, k8s_namespace):
-    yaml = SCRIPT_DIR / "mongodb-k8s.yaml"
```
jaysfx: I noticed all the run_k8s_monitors_test tests were deleted. Are they tested some other way in kubernetes?

keitwb: I converted the existing kubernetes-specific monitor tests to not use that helper and removed the old monitor tests that were just deploying an app on k8s and testing monitor code that had nothing to do with kubernetes.  We weren't actually running them on CircleCI for a few months now anyway.

PR #748, opened 2019-04-25
Fix statsd integration tests by dynamic port assignment

seonsfx:


tests/monitors/statsd/statsd_test.py
```diff
@@ -70,16 +87,20 @@ def test_statsd_monitor_conversion():
...
-    listenAddress: localhost
-    listenPort: 8127
+    listenPort: 0
     converters:
     - pattern: 'cluster.cds_{traffic}_{mesh}_{service}-vn_{}.{action}'
       metricName: '{traffic}.{action}'
 """
     ) as agent:
-        assert wait_for(p(udp_port_open_locally_netstat, 8127)), "statsd port never opened!"
+        assert wait_for(p(regex_search_matches_output, agent.get_output, STATSD_RE.search))
```
keitwb: Make a helper that gets the port from the agent instance and cut out all the duplication between test cases.

seonsfx: Done

PR #749, opened 2019-04-25
Keep major version floating tag up to date on releases

keitwb:


scripts/release
```diff
@@ -262,10 +267,15 @@ build_docker_image() {
...
-push_docker_image() {
-  local stage="$1"
-  local new_version="$2"
-  docker push $(docker_repo_from_stage $stage):$new_version
+update_floating_docker_tags() {
+  local new_version="$1"
+  local new_image="$2"
+  local image_base="$3"
+
+  major_version=$(echo -n "$new_version" | head -c1)
```
seonsfx: Are we planning to change this script when we start having double digit major versions?

jaysfx: To @seonsfx's point, maybe do `cut -d . -f 1` instead since it's future-proof?

keitwb:
```suggestion
  major_version=$(echo -n "$new_version" | cut -d. -f1)
```

Issue #750, opened 2019-04-26
Emitting a metrics for failing plugins


MovieStoreGuy:
I wanted to open a conversation about potentially reducing the amount of logs been written to switch them to a counter / gauge being sent to SignalFx instead.

For example, the collected plugins can be rather noisey if they are interrupted or if the service they are monitoring restarts, doesn't respond in time, etc. 
My thought would be to emit an event that we could track or alert on in the UI to follow up rather than searching through all of our logs as that can get rather expensive storing it.

keitwb:
Sure, having metrics about the agent and its monitors themselves is a good idea; I've been thinking about doing it for a while.  Events would provide an exact error message, but metrics are a lot more scaleable in our backend, and events aren't really intended to be used for potentially high-frequency error logs anyway in our current product offering.

Therefore, having metrics for monitors that track failures to connect/read from a service being monitored is the best thing to do.  Which monitors/collectd plugins specifically are you dealing the most with?  I can try adding some metrics for a few of them and see how it works for you.

I think maintaining the current logging output though is pretty essential for basic debugging of the agent and to know more context about the errors directly from the agent.  Not sure how much control you have over log retention but I can't imagine there is much value in keeping agent logs for more than a few days anyway unless you are trying to do some kind of analytics on them.  I know you were talking about filtering in #719 and that is probably your best bet for this kind of thing.  You can filter out the most common stuff (e.g. a connection refused/timeout error) but still let through rarer error conditions (e.g. a 500 response from a diagnostic endpoint on a service) so that they show up in logs.

MovieStoreGuy:
Yeah, I want use this as reducing the amount of things ending up in our logging system.
I was having a discussion else where about having detectors for events but I gathered that is a bit of time away.

So the monitors I have mainly used that have been creating a lot of noise are:
- expvar
- collectd/genericjmx 
We also get a decent amount of _EOF_ or _timed out awaiting headers_ errors so we also don't really don't care about that per say and we definitely don't want service owners to freak out that there is an issue when it is just a NOOP.

I was thinking about the dimensions that would be useful and the ones I have are:
- monitor
- monitor_version
- agent_version 
So that way we can see if bumping agent version increases / decreases issues as we run the smart agent across all of instances.

We log an insane amount a day so anything we can do to reduce is always appreciated for those who maintain our logging system. Also, while we are on the topic of logging. How hard of a change would it be to switch logging format as a config option?

keitwb:
[...]

PR #751, opened 2019-04-26
Add JSON log output option

keitwb:


seonsfx:
Is it for https://github.com/signalfx/signalfx-agent/issues/750 ?

keitwb:
It could be if they want JSON output.

internal/core/config/validation/validator.go
```diff
@@ -28,14 +31,19 @@ func ValidateCustomConfig(conf interface{}) error {
 
 // ValidateStruct uses the `validate` struct tags to do standard validation
 func ValidateStruct(confStruct interface{}) error {
+	en := en.New()
+	uni := ut.New(en, en)
+	trans, _ := uni.GetTranslator("en")
```
jaysfx: What's this stuff for?

keitwb: It is used in the error message generated below to make it a lot more user friendly.  You have to do this convolution to make it English since they support tons of languages.

MovieStoreGuy: Also, why are you ignoring the error being returned here?

internal/core/config/validation/validator.go
```diff
@@ -28,14 +31,19 @@ func ValidateCustomConfig(conf interface{}) error {
...
+
 	validate := validator.New()
+	validate.RegisterTagNameFunc(utils.YAMLNameOfField)
+	_ = en_translations.RegisterDefaultTranslations(validate, trans)
```
MovieStoreGuy: What is being ignored here?

MovieStoreGuy:
Is it really necessary to commit the vendor files?
I know there is some level of optimisation for the build work but does it really make it that much faster?

It feels like a very bad mix of Pre modules and Post modules.

keitwb:
> Is it really necessary to commit the vendor files?
> I know there is some level of optimisation for the build work but does it really make it that much faster?
> 
> It feels like a very bad mix of Pre modules and Post modules.

We will probably move off vendoring at some point but I don't really want to be that dependent on Github to do builds (part of getting off vendoring would probaly involve finding a fast, reliable package mirror repo service).  We just recently switched to modules so getting off vendoring is probably the next move once we are fully comfortable with modules.

internal/core/config/validation/validator.go
keitwb: Because English is always there in our locked dependency version.

internal/core/config/validation/validator.go
keitwb: An error that will never occur with the standard english translation.

MovieStoreGuy: Ah, sounds good :D

PR #753, opened 2019-04-29
k8s test updates

jchengsfx:


.circleci/config.yml
```diff
@@ -137,15 +137,15 @@ commands:
...
             cd /tmp
             curl -LO https://storage.googleapis.com/kubernetes-release/release/<< parameters.version >>/bin/linux/amd64/kubectl
             chmod +x ./kubectl
-            sudo mv ./kubectl /usr/local/bin/kubectl
+            sudo mv ./kubectl /home/circleci/bin/kubectl
       - save_cache:
           key: "v1-kubectl-binary-<< parameters.version >>"
           paths:
-            - /usr/local/bin/kubectl
+            - /home/circleci/bin/kubectl
```
keitwb: Existing cache from prior runs will contain the old path, you have to bust it somehow.

jchengsfx: > Existing cache from prior runs will contain the old path, you have to bust it somehow.

Updated kubectl cache to v2.

tests/observers/k8s_api_test.py
```diff
@@ -3,21 +3,22 @@
...
-from tests.helpers.kubernetes.utils import add_pod_spec_annotations
+from tests.helpers.kubernetes.utils import add_pod_spec_annotations, get_discovery_rule
 from tests.helpers.util import wait_for
 from tests.paths import TEST_SERVICES_DIR
 
 
 @pytest.mark.kubernetes
 def test_k8s_api_observer_basic(k8s_cluster):
     nginx_yaml = TEST_SERVICES_DIR / "nginx/nginx-k8s.yaml"
+    discovery_rule = get_discovery_rule(nginx_yaml, "k8s-api", namespace=k8s_cluster.test_namespace)
```
keitwb: Not really a big fan of the `get_discovery_rule` helper because it obscures what it's actually testing in terms of discovery rules.  Does the helper do something needed for tests to consistently pass?

tests/helpers/kubernetes/tunnel.py
```diff
@@ -20,7 +21,10 @@ def deploy_fake_backend_proxy_pod(namespace):
     """
     corev1 = kube_client.CoreV1Api()
     pod_yaml = Path(SCRIPT_DIR / "tunnel/pod.yaml").read_bytes()
-    pod = corev1.create_namespaced_pod(body=yaml.safe_load(pod_yaml), namespace=namespace)
+    pod = retry(
```
keitwb: Do you know why this was failing?  It seems more like a bug in Kubernetes if the pod fails to be created.

tests/observers/k8s_api_test.py
jchengsfx: > Not really a big fan of the `get_discovery_rule` helper because it obscures what it's actually testing in terms of discovery rules. Does the helper do something needed for tests to consistently pass?

I added `get_discovery_rule` just to include more endpoint variables and the `Get` and `Contains` functions in the test.

keitwb: It would be better to have tests that explicitly test those things, e.g. `test_pod_labels_in_discovery_rule` or something so that it was explicit and obvious.

tests/deployments/helm/helm_test.py
```diff
@@ -63,6 +63,12 @@ def helm_command_prefix(k8s_cluster):
     )
 
 
+def exec_helm(args):
+    proc = subprocess.run(args, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
jaysfx: If you set `encoding='utf-8'` the encoding is done for stdout/stderr automatically. In 3.7 you can set `text=True` I think but we're on 3.6 afaik.

keitwb:
@jchengsfx I just noticed that windows tests didn't run on this PR, is that some kind of setting in the Azure Pipelines config that it only runs tests against master PRs?

PR #755, opened 2019-04-29
adding support for timerCount statsd config

dr-strangecode:
Looks like timerCount config for statsd was missed.

you can see it as an option for collectd here:
https://github.com/collectd/collectd/blob/d1c056161220ce78dedefccf90c876cbea64e9d4/src/statsd.c#L86

I *think* I found all the relevant places that needed updating


keitwb:
@dr-strangecode Thanks for the contribution.  I approved and added a small tweak from the master branch that I did dealing with the newline at the end of the selfdescribe.json doc so it should pass CI now.

If you make further contributions, know that you can run `make signalfx-agent && make docs` and it will add the stuff to `selfdescribe.json` and the monitor docs under `docs/monitors` for you if you don't want to hand edit it.

PR #757, opened 2019-04-30
Making puppet module more robust on Debian

phardy:
Hey folks. I'm working on migrating to SignalFx across our infrastructure, but ran in to a couple of problems with the Puppet module:

* The module was attempting to create `/etc/signalfx/agent.yaml` before the package was installed, therefore `/etc/signalfx` did not yet exist and provisioning errored out.
* The `Package['wget']` resource was duplicating our attempt to install wget elsewhere in our Puppet manifests.

This PR addresses the first problem by simply adding some chaining arrows.

To address the second problem, I've chosen to introduce a dependency on the `puppetlabs/apt` module to manage the apt repository. Happy to discuss alternatives for that, though.

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -406,8 +406,9 @@ jobs:
                 ;;
             puppet)
                 docker run --rm \
+                    -e "CI_SPEC_OPTIONS=--format RspecJunitFormatter -o ~/testresults/puppetspec.xml" \
                     signalfx-agent-puppet-dev \
-                    rspec spec --format RspecJunitFormatter | tee ~/testresults/puppetspec.xml
+                    rake spec
                 docker run --rm \
                     signalfx-agent-puppet-dev \
                     puppet-lint .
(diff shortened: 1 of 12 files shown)


phardy:
Sorry, despite using puppet for years I'm a relative newcomer to actually writing modules. I'm still missing something getting the tests to pass on this. I've added the new dependency in `metadata.json` and `.fixtures.yml` as per https://github.com/puppetlabs/puppetlabs_spec_helper#using-fixtures and https://puppet.com/docs/puppet/5.1/modules_metadata.html but the new fixture still doesn't seem to be included when I build the signalfx-agent-puppet-dev image. Am I doing something stupid?

keitwb:
Hi @phardy, thanks for the contribution.  It looks good what you did.  I pushed a commit to try and resolve the fixture install issues (I think it wasn't ever setup properly from the beginning since we didn't have any non-built-in dependencies).  Also I moved the apt dependency in the metadata.json to the `dependencies` section instead of `requirements`, as everything I saw seemed to indicate modules should go there.

keitwb:
@phardy this was released as version 0.2.0 in the Puppet Forge.

PR #760, opened 2019-05-02
Monitor elasticsearch metric filtering

bjsignalfx:


internal/monitors/elasticsearch/client/metricnames.go
```diff
@@ -35,44 +35,44 @@ const (
 	// Flush stats
 	indicesFlushTotal     = "elasticsearch.indices.flush.total"
 	indicesFlushPeriodic  = "elasticsearch.indices.flush.periodic"
-	indicesFlushTotalTime = "elasticsearch.indices.flush.total-time"
+	IndicesFlushTotalTime = "elasticsearch.indices.flush.total-time"
```
jaysfx: Remove all these where possible and use the generated constants from genmetadata.go. If there are metrics missing from metadata add them.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -319,6 +319,139 @@ func (m *Monitor) sendDatapoints(dps []*datapoint.Datapoint) {
...
+	}
+	if c.EnableEnhancedTransportStats {
+		extraMetrics = append(extraMetrics, groupMetricsMap[groupNodeTransport]...)
+	}
+	enhancedStatsForIndexGroups := utils.StringSliceToMap(append(c.EnableEnhancedNodeStatsForIndexGroups, c.EnableEnhancedIndexStatsForIndexGroups...))
+	if enhancedStatsForIndexGroups[client.StoreStatsGroup] {
+		extraMetrics = append(extraMetrics, elasticsearchIndicesStoreThrottleTime)
+	}
+	if enhancedStatsForIndexGroups[client.IndexingStatsGroup] {
+		extraMetrics = append(extraMetrics, elasticsearchIndicesIndexingIndexCurrent)
```
jaysfx: Do these not relate to the groups somehow? We ideally wouldn't have to hardcode so many metrics here.

internal/monitors/elasticsearch/elasticsearch.go
```diff
@@ -319,6 +319,139 @@ func (m *Monitor) sendDatapoints(dps []*datapoint.Datapoint) {
...
+		extraMetrics = append(extraMetrics, elasticsearchIndicesGetExistsTime)
+		extraMetrics = append(extraMetrics, elasticsearchIndicesGetMissingTotal)
+		extraMetrics = append(extraMetrics, elasticsearchIndicesGetMissingTime)
+	}
+	if enhancedStatsForIndexGroups[client.SearchStatsGroup] {
+		extraMetrics = append(extraMetrics, elasticsearchIndicesSearchQueryCurrent)
+		extraMetrics = append(extraMetrics, elasticsearchIndicesSearchFetchCurrent)
+		extraMetrics = append(extraMetrics, elasticsearchIndicesSearchScrollCurrent)
+		extraMetrics = append(extraMetrics, elasticsearchIndicesSearchSuggestCurrent)
+		extraMetrics = append(extraMetrics, client.IndicesSearchOpenContexts)
```
jaysfx: Looks like these metrics from client should all be in the metadata.

jaysfx:
Also see seon's docker change that just got merged. It uses .EnabledMetrics to determine which groups should be implicitly turned on as well.

internal/monitors/elasticsearch/client/metricnames.go
jaysfx: Is this possible?

bjsignalfx: No, because the metadata constants are not exported so can't use them in package client.

PR #764, opened 2019-05-03
add an example usage of helm chart

jadeblaquiere:
simple example showing usage to clarify usage for the helm-unitiated

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/README.md
+++ b/deployments/k8s/helm/signalfx-agent/README.md
@@ -32,8 +32,13 @@ See the [values.yaml](./values.yaml) file for more information on how to
 configure releases.
 
 There are two **required** config options to run this chart: `signalFxAccessToken`
-and `clusterName` (if not overridding the agent config template and providing your own
-cluster name).
+and `clusterName` (if not overridding the agent config template and providing your own cluster name).
+
+It is also **recommended** that you explicitly specify `agentVersion` when deploying a release so that the agent will not be unintentionally updated based on updates of the helm chart from the repo.
+
+For example a basic command line install setting these three values would be:
+
+`$ helm install --set signalFxAccessToken=<YOUR_ACCESS_TOKEN> --set clusterName=<YOUR_CLUSTER_NAME> --set agentVersion=<VERSION_NUMBER> signalfx/signalfx-agent`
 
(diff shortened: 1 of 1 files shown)


deployments/k8s/helm/signalfx-agent/README.md
```diff
@@ -32,8 +32,13 @@ See the [values.yaml](./values.yaml) file for more information on how to
 configure releases.
 
 There are two **required** config options to run this chart: `signalFxAccessToken`
-and `clusterName` (if not overridding the agent config template and providing your own
-cluster name).
+and `clusterName` (if not overridding the agent config template and providing your own cluster name).
+
+It is also **recommended** that you explicitly specify `agentVersion` when deploying a release so that the agent will not be unintentionally updated based on updates of the helm chart from the repo.
```
jaysfx: FYI right now you have to set `agentVersion` or else you'll get an old version of the agent (4.5 I believe). We currently don't update the helm chart with new agent versions when we push new agents. I have it on my list of things to do but haven't gotten around to it.

jadeblaquiere: I discussed with @keitwb and while it's not a bad idea for demo scenarios, his guidance was the above... recommend setting version explicitly so that you are deliberately setting the version and upgrading would be a choice rather than something that mysteriously happened in the middle of the night because somebody pulled an update of the helm chart.

Still good to set it to "latest" by default for POC cases and such where latest is generally right.

jaysfx: @jadeblaquiere  I'm not a helm power user but as I understand it just pulling an updated helm chart wouldn't result in an upgrade. Somebody would have to actually do a `helm upgrade` to deploy the new chart.

I was just letting you know that right now if somebody does `helm install` they get version 4.5 of the agent, not the newest agent. Right now we don't bump the default agent version as part of release inside the helm chart.

mstumpfx:
I duplicated this in #778 and added a little more content. If the other merges we can just close this one.

keitwb:
> I duplicated this in #778 and added a little more content. If the other merges we can just close this one.

@mstumpfx I don't see it duplicated in that PR.  Do you still need to push something?

PR #767, opened 2019-05-05
Filtering Test Fixes and Refactorings

keitwb:
This should have been several PRs but it got too interwoven to split up after I got too far into it.


deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
```diff
@@ -117,6 +117,8 @@ data:
       timeout: {{ .Values.timeout | default 40 }}
       logLevel: {{ .Values.logLevel | default "notice" }}
 
+    enableBuiltInFiltering: true
```
jaysfx: I wonder if this should be put into values.yaml and conditionalized to false for those upgrading from a previous helm release? ie only turn it on by deafult if it's a non-upgrade and for existing installs they can override it to true when they're ready to transition. This could be checked with `Release.IsUpgrade`

deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -131,7 +131,6 @@ gatherDockerMetrics: true
 # cAdvisor metrics that cause a large increase in DPM for little value for most
 # customers.
 metricsToExclude:
```
jaysfx: Similar to above.

deployments/puppet/README.md
```diff
@@ -14,6 +14,7 @@ class accepts the following parameters:
     ```ruby
     $config = {
       signalFxAccessToken: "MY_TOKEN",
+	  enableBuiltInFiltering: true,
```
jaysfx: Indentation off here

deployments/salt/pillar.example
```diff
@@ -4,6 +4,7 @@ signalfx-agent:
   conf_file_path: '/etc/signalfx/agent.yaml'
   conf:
     signalFxAccessToken: 'My_Token'
+	enableBuiltInFiltering: true
```
jaysfx: Indentation off here

tests/monitors/docker_container_stats/docker_container_stats_test.py
```diff
@@ -161,27 +160,33 @@ def test_docker_stops_watching_destroyed_containers():
...
-    verify_included_metrics(
-        f"""
-        monitors:
-        - type: docker-container-stats
-        """,
-        METADATA,
-    )
+    with run_service(
+        "elasticsearch/6.6.1"
+    ) as _:  # just get a container that does some block io running so we have some stats
```
jaysfx: Can drop `as _` here and below.

whitelist.json
```diff
@@ -1,1014 +1 @@
...
-            "gauge.openstack.nova.server.tx_drop",
-            "gauge.openstack.nova.server.vda_errors",
-            "gauge.openstack.nova.server.vda_read",
-            "gauge.openstack.nova.server.vda_read_req",
-            "gauge.openstack.nova.server.vda_write",
-            "gauge.openstack.nova.server.vda_write_req"
-        ]
-    }
-]
+[]
```
jaysfx: What happens to people that upgrade the agent but still reference this file using their existing configs? Won't their filtering break?

deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
keitwb: Fortunately in the Helm configmap we never let people pass through `metricsToInclude` anywhere (they would have to provide their own config in its entirety instead of relying on templating in this file).  Wasn't the whole reason we went with the flag because of uncertainties around `metricsToInclude` backwards compatibility?  Or was there something else I am forgetting?

deployments/k8s/helm/signalfx-agent/values.yaml
keitwb: Since I took the whitelist.json reference out of the metricsToExclude default value in `values.yaml`, it shouldn't use it at all unless people explicitly set `metricsToExclude` to something non-default.

whitelist.json
keitwb: Yeah that's true, I guess we will need to leave it.

deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
jaysfx: Yep that should be safe then.

whitelist.json
jaysfx: Sorry, I know how badly you wanted to burn it.

tests/monitors/collectd_postgresql/postgresql_test.py
```diff
@@ -51,3 +53,6 @@ def test_postgresql():
                 p(has_datapoint_with_dim, agent.fake_services, "plugin", "postgresql")
             ), "Didn't get postgresql datapoints"
             assert wait_for(p(has_datapoint_with_metric_name, agent.fake_services, "pg_blks.toast_hit"))
+            import time
```
jaysfx: ?

keitwb: Sorry, left overs from test debugging.

PR #768, opened 2019-05-06
Add GetExtraMetrics RabbitMQ

seonsfx:


internal/monitors/collectd/etcd/etcd.go
```diff
@@ -70,3 +70,16 @@ func (m *Monitor) Configure(conf *Config) error {
 
 	return m.PyMonitor.Configure(conf)
 }
+
+// GetExtraMetrics returns additional metrics that should be allowed through.
+func (c *Config) GetExtraMetrics() []string {
+	if c.EnhancedMetrics != nil && *c.EnhancedMetrics {
+		return []string{
```
keitwb: Use `monitorMetadata.NonIncludedMetrics()` instead of listing them out.  It will be less prone to bugs if we add more non-included metrics in the future (or realize we didn't include some in metadata.yaml).

seonsfx:
just did an experimental commit to check a integration test failure on CI

PR #775, opened 2019-05-07
More expvar rework and generic extraMetrics integration test

keitwb:


internal/monitors/expvar/expvar.go
```diff
@@ -56,6 +56,10 @@ type Monitor struct {
 func (m *Monitor) Configure(conf *Config) error {
 	m.logger = log.WithFields(log.Fields{"monitorType": monitorType})
 
+	if m.Output.HasEnabledMetricInGroup(groupEnhanced) {
```
jaysfx: I don't think we should use enhanced as the group. There are probably better groups to assign the metrics to. Just off the top of my head it looks like maybe a memory group and a gc group. But regardless it seems what this is really saying is `m.Output.IsAnyCustomMetricEnabled()`.

tests/conftest.py
```diff
@@ -21,6 +21,8 @@ def pytest_collection_modifyitems(items):
                 item.add_marker("integration")
 
 
+pytest.register_assert_rewrite("tests.helpers.verify")
```
jaysfx: Duplicate

tests/basic/extra_metrics_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_extra_metrics_passthrough():
+    """
+    The specified extraMetrics should be allowed through even though they are
+    not included by default.
+    """
+    metadata = Metadata.from_package("expvar")
+
+    with run_expvar() as expvar_container_ip:
```
jaysfx: btw you can have multiple context managers with one with statement that gets rid of multiple levels of indenting. Something like:

```python
with run_expvar() as expvar_container_ip, Agent.run(..):
   pass
```

tests/basic/extra_metrics_test.py
```diff
@@ -0,0 +1,31 @@
...
+        with Agent.run(
+            dedent(
```
jaysfx: I've been removing `dedent` usage with seemingly no ill effects. Was also thinking of just adding it to `Agent.run` itself.

tests/monitors/expvar/expvar_test.py
```diff
@@ -2,52 +2,67 @@
...
-          port: 8080
-          enhancedMetrics: true
-        """,
-        METADATA,
-    )
-
-
-def test_expvar_custom_metric(expvar_container_ip):
+@contextmanager
+def run_expvar():
```
jaysfx: Looks like you can remove `expvar/conftest.py`.

internal/monitors/expvar/config.go
```diff
@@ -30,6 +30,18 @@ type Config struct {
 	MetricConfigs []MetricConfig `yaml:"metrics"`
 }
 
+// GetExtraMetrics handles the legacy enhancedMetrics option.
+func (c *Config) GetExtraMetrics() []string {
+	if c.EnhancedMetrics {
+		// Returning everything is more future-proof than trying to enumerate
+		// each one.
+		return []string{"*"}
```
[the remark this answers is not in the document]
keitwb: Yeah, I mean if the monitor gets refactored somehow later such that the MetricConfig thing doesn't correspond 1:1 to metric names just returning `*` is the most robust thing that would avoid regressions.

For the monitors with a single generic "enhanced metrics" flag, it usually just means https://i.kym-cdn.com/photos/images/original/000/832/085/578.gif

internal/monitors/expvar/expvar.go
keitwb: Yeah, the method to test for any non-default metric seems cleaner.

tests/basic/extra_metrics_test.py
keitwb: Yeah I think as long as the YAML is otherwise indented properly the parser in the agent is tolerant of consistent leading whitespace.

tests/basic/extra_metrics_test.py
keitwb: That would work if the agent config didn't depend on the container ip.

jaysfx: It still works in that case:

```python
    with run_expvar() as expvar_container_ip, Agent.run(
        f"""
        monitors:
          - type: expvar
            host: {expvar_container_ip}
            port: 8080
            intervalSeconds: 1
            extraMetrics:
             - memstats.by_size.mallocs
        """
    ) as agent:
        assert "memstats.by_size.mallocs" in metadata.nonincluded_metrics
        verify(agent, metadata.included_metrics | {"memstats.by_size.mallocs"})
```

jaysfx: It doesn't really bother me either way, just an fyi thing.

PR #779, opened 2019-05-07
Filtering tests for cadvisor, collectd/processes, and collectd/apache

jaysfx:


docs/monitors/collectd-processes.md
```diff
@@ -49,6 +49,78 @@ Monitor Type: `collectd/processes`
 
 
 
+## Metrics
+
+The following table lists the metrics available for this monitor. Metrics that are marked as Included are standard metrics and are monitored by default.
+
+| Name | Type | Included | Description |
```
keitwb: This monitor needs to have `sendAll: true` to match existing behavior (it isn't in the whitelist).

PR #780, opened 2019-05-07
 Get platform-specific metadata metrics

jchengsfx:


tests/helpers/metadata.py
```diff
@@ -11,8 +12,8 @@ class Metadata:
     def __init__(self, monitor_type, included_metrics, nonincluded_metrics, metric_by_group, dims):
         self.monitor_type = monitor_type
         self.included_metrics = included_metrics
-        self.nonincluded_mertics = nonincluded_metrics
-        self.all_metrics = self.included_metrics | self.nonincluded_mertics
+        self.nonincluded_metrics = nonincluded_metrics
```
keitwb: I think you are like the fourth person to correct this.  Expect merge conflicts.

tests/helpers/metadata.py
```diff
@@ -56,7 +57,10 @@ def _find_monitor(monitors, mon_type):
 def _get_monitor_metrics(monitor, included):
     for metric, info in (monitor.get("metrics") or {}).items():
         if info["included"] == included:
-            yield metric
+            if sys.platform == "win32" and "linux only" not in info["description"].lower():
```
keitwb: 👎 This needs to be a dedicated field in the metadata.yaml that gets passed through.

jchengsfx:
> Make platform a first-class field in metadata.yaml. There are several things we could do with it if it were such a thing and would be way more robust.

Agreed @keitwb. Is it already planned or in development by someone?

keitwb:
Nothing formal has been planned for it yet.  If you have some time feel free to tackle it.

PR #781, opened 2019-05-08
Consul and HAProxy + many more filtering work

keitwb:


tests/monitors/solr/solr_test.py
```diff
@@ -28,10 +26,6 @@ def test_solr_monitor():
...
-            assert wait_for(
-                p(has_datapoint_with_dim, agent.fake_services, "plugin", "solr")
-            ), "Didn't get solr datapoints"
-            assert ensure_always(
-                lambda: has_datapoint_with_metric_name(agent.fake_services, "counter.solr.http_5xx_responses")
-            )
-            assert not has_log_message(agent.output.lower(), "error"), "error found in agent output!"
+            # We don't get all default metrics but this ensures we don't get
+            # any non-default metrics through with default config.
+            verify_expected_is_superset(agent, METADATA.included_metrics)
```
jaysfx: Seems like this would have the problem that the condition is immediately satisfied on the first execution regardless if there have been any datapoints received as the empty set is a subset of all sets.

keitwb: Ah yeah, that's a problem..

keitwb: Ok, added a manual sleep and assertion that at least some datapoints have been received.  Open to a better solution but this seems the simplest to me.

PR #782, opened 2019-05-08
Couchbase filtering updates

jaysfx:
* Also changes `HasAnyNonDefaultMetricEnabled` to `HasAnyExtraMetrics` because the
  former didn't account for the fact that not all metrics are documented.


internal/monitors/filtering.go
```diff
@@ -63,15 +65,10 @@ func (mf *monitorFiltering) HasEnabledMetricInGroup(group string) bool {
...
 // enabled for this output instance.
-func (mf *monitorFiltering) HasAnyNonDefaultMetricEnabled() bool {
-	for _, m := range mf.EnabledMetrics() {
-		if !mf.metadata.IncludedMetrics[m] {
-			return true
-		}
-	}
-	return false
+func (mf *monitorFiltering) HasAnyExtraMetrics() bool {
+	return mf.hasExtraMetrics
```
keitwb: `mf` can be `nil` if `enableBuiltInFiltering: false`.

keitwb: See https://github.com/signalfx/signalfx-agent/pull/781/files#diff-246334eb435b0d0c6d4e4357d843a959R39

PR #783, opened 2019-05-08
collectd/cpu and cpu monitor filtering/tests

jaysfx:
* Removed `cpu.utilization` from `collectd/cpu` because it's actually generatd
  by the metadata plugin.

* Renamed audit-monitor to audit_monitor to remove the Python path issues.
  Easiest way to run it now is:

     python -m scripts.audit_monitor


internal/monitors/collectd/cpu/metadata.yaml
```diff
@@ -8,10 +8,48 @@ monitors:
     (the `cpu.utilization` metric).
 
     See https://collectd.org/wiki/index.php/Plugin:CPU
+  # There are no included metrics but shouldn't sendAll by default.
+  sendAll: false
```
keitwb: Why not send all?  It is a potential source of confusion to have a monitor that emits absolutely nothing when turned on.

PR #785, opened 2019-05-09
Add more generic filtering tests

keitwb:


tests/basic/built_in_filtering_test.py
```diff
@@ -0,0 +1,91 @@
...
+               metricsToExclude:
+                - {{"#from": "{REPO_ROOT_DIR}/whitelist.json", flatten: true}}
+               """
+        ) as agent:
+            verify(agent, metadata.all_metrics)
+
+
+def test_built_in_filtering_disabled_whitelisted_monitor():
+    """
+    Test a monitor that doesn't have any entries in whitelist.json
```
jaysfx: Doesn't appear to match the test

keitwb:
```suggestion
    Test a monitor that is in whitelist.json.
```

PR #793, opened 2019-05-10
Migrate activemq docs from integrations repo

keitwb:
Also a few tweaks to doc gen and metric generation for integrations
repo


alstory:
No comment from me.

PR #796, opened 2019-05-10
Add docs for built-in filtering

keitwb:


docs/monitors/collectd-cassandra.md
```diff
@@ -60,7 +60,7 @@ The **nested** `values` config object has the following fields:
 
 The following table lists the metrics available for this monitor. Metrics that are marked as Included are standard metrics and are monitored by default.
```
barbara-sfx: If you're going to change the table header, then you have to change the text from:

Metrics that are marked as Included

to 

Metrics that are marked as "Default (non-custom)" 

throughout

I added quotes because now it's two words.

docs/monitors/appmesh.md
```diff
@@ -131,80 +131,12 @@ The following table lists the metrics available for this monitor. Metrics that a
 | `upstream_rq_tx_reset` | cumulative |  | Total requests that were reset locally |
```
barbara-sfx: Where is all the info about metricsToInclude going?

jaysfx:
I added some changes in https://github.com/signalfx/signalfx-agent/tree/filtering-docs-jay for you to consider.

Mainly I split the old filtering stuff into its own document to try and make the filtering page look less daunting by just having the new way.

Even with that change it's still a lot to grok at once. I wonder if it might help to have separate pages for extra metrics and filtering. I think they're fairly independent in that you can learn and use one without understanding the other. From an implementation perspective we think about extra metrics in terms of filtering but as a concept I don't think it has that much to do with filtering.

barbara-sfx:
> I added some changes in https://github.com/signalfx/signalfx-agent/tree/filtering-docs-jay for you to consider.
> 
> Mainly I split the old filtering stuff into its own document to try and make the filtering page look less daunting by just having the new way.
> 

@jaysfx  Without a PR, it's hard (impossible?) to know which files you changed in that branch. (Unless there is a way that you can teach me.) Can you either make a note here or create a PR for pulling that branch into this one, and then request reviews on that? Thanks.

jaysfx:
@barbara-sfx If you want you can wait until Ben gets back on Monday and he decides if he agrees with my changes or not before you do your review. However if you'd like to see the changes I'm proposing you can compare branches on Github.  You can do it by switching to my branch on the main page, clicking the button compare (may have to do find, it's hard to see), then choose filtering-docs as the base branch. Or just use this link:

https://github.com/signalfx/signalfx-agent/compare/filtering-docs...filtering-docs-jay

You can plug in whatever branches you want there in the future for easy comparison. Hope that helps.

barbara-sfx:
Thanks, that does help. (And I think I even knew that at one time.) I have set aside time for PR reviews Monday and Thursday afternoons, so it probably does make sense for Ben to look at it first.

docs/monitors/appmesh.md
keitwb: It is going away and replaced by the added text below that links to the filtering doc that would have pointers to the legacy stuff but generally deemphasize it.

docs/monitors/collectd-cassandra.md
keitwb: I changed it.

scripts/docs/templates/monitor-page.md.tmpl
```diff
@@ -41,40 +42,35 @@ This monitor has no configuration options.
...
-{{if not $info.included}}
-    - {{ $metric }}
-{{- end -}}
-{{end}}
-    monitorType: {{.monitorType}}
-```
-{{end}}
+### Built in filtering
+This monitor will perform built-in filtering if you are using agent version
+4.7.0+ and have the `enableBuiltInFiltering: true` option set at the top-level
```
jrcamp: This goes along with my comment about having separate topics for filtering and extra metrics. I think it's confusing here that to get more metrics you have to go to a page called Filtering. Filtering implies you want less of something when you're trying to actually get more of something.

It also means the option `enableBuiltInFiltering` is confusing if you're thinking in the context of configuring additional metrics. (Even though in implementation details it makes perfect sense to us).

keitwb: Ok, I moved all of the relevant parts about extra/custom metrics to the monitor docs and left the "filtering" doc to just contain stuff about excluding datapoints.  That does seem to be clearer.

docs/config-schema.md
```diff
@@ -54,7 +54,7 @@ if not set.
...
+| `enableBuiltInFiltering` | no | bool | If true, the agent will inherently filter out [custom metrics](https://docs.signalfx.com/en/latest/admin-guide/usage.html#about-custom-bundled-and-high-resolution-metrics) without having to rely on the `whitelist.json` filter that was previously configured under `metricsToExclude`.  Whether a metric is custom or not is documented in each monitor's documentation.  Now every monitor's default configuration (i.e. the minimum amount of configuration to make it work) will only send non-custom metrics, but if you enable certain flags or config options on the monitor, custom metrics will also be sent without having to modify the whitelist via `metricsToInclude`.  If this option is set to `true`, the `whitelist.json` entry under `metricToExclude` should be removed, if it was present. (**default:** `false`) |
```
barbara-sfx: I can't quite figure out what this is saying. Either it's saying the same thing different ways, or it's mushing together a few different ideas, or something.. Especially the sentence starting "Now every monitor's.."  You mention metricsToInclude and metricsToExclude and it's just a bit unclear. Seems like it needs to be broken out into a couple paragraphs or bullet list, maybe. 

And if I'm reading it right, it sounds like this option replaces the need for whitelist.json and metricsToExclude. If I'm reading it right, you should say it in that way. Also, in the docs for whitelist.json and metricsToExclude, are you going to mention that that usage is deprecated or superseded in favor of this usage?  (Or just delete it altogether and refer to this option instead?) 

Also, please make the last sentence active: "If you set this option to true, delete the whitelist.json entry..."  BTW, what happens if they don't delete it?

barbara-sfx: Remove "inherently"

docs/legacy-filtering.md
```diff
@@ -0,0 +1,151 @@
...
+`metricsToExclude` section of the standard distributed config file.  This
+method is now deprecated but will continue to work if already configured as
+such.  If you upgrade to 4.7.0 or later, you will continue to need the
+`whitelist.json` custom-metric filtering until you set `enableBuiltInFiltering:
+true` in your config.  You should **remove** the reference to `whitelist.json` in
+your config at the same time you set that flag.
+
+Essentially the effect of setting `enableBuiltInFiltering: true` and removing
+the reference to `whitelist.json` is to make the monitors inherently emit the
+metrics that were in `whitelist.json` by default, which making it significantly
```
barbara-sfx: which making ==> which makes

docs/legacy-filtering.md
```diff
@@ -0,0 +1,151 @@
...
+easier to configure monitors to emit additional metrics in a single step.
+
+We recommend going ahead and upgrading the agent to at least 4.7.0 and setting
```
barbara-sfx: remove "going ahead and"

docs/monitor-config.md
```diff
@@ -36,7 +36,7 @@ The following config options are common to all monitors:
...
 | `dimensionTransformations` |  | no | `map of string` | A map from dimension names emitted by the monitor to the desired dimension name that will be emitted in the datapoint that goes to SignalFx.  This can be useful if you have custom metrics from your applications and want to make the dimensions from a monitor match those. Also can be useful when scraping free-form metrics, say with the `prometheus-exporter` monitor.  Right now, only static key/value transformations are supported.  Note that filtering by dimensions will be done on the *original* dimension name and not the new name. |
-| `extraMetrics` |  | no | `list of string` | Extra metrics to enable besides the default included ones. |
+| `extraMetrics` |  | no | `list of string` | Extra metrics to enable besides the default included ones.  This is an [overridable filter](https://docs.signalfx.com/en/latest/integrations/agent/filtering.html#overridable-filtering). |
```
barbara-sfx: list of string => list of strings (throughout)

docs/monitor-config.md
```diff
@@ -36,7 +36,7 @@ The following config options are common to all monitors:
...
 | `extraGroups` |  | no | `list of string` | Extra metric groups to enable besides the metrics that are included by default. |
```
barbara-sfx: Is "metric group" defined somewhere?

docs/monitors/appmesh.md
```diff
@@ -48,20 +48,24 @@ Monitor Type: `appmesh`
...
 | `metricPrefix` | no | `string` | A prefix in metric names that needs to be removed before metric name conversion |
 
 
-
-
 ## Metrics
 
-The following table lists the metrics available for this monitor. Metrics that are marked as Included are standard metrics and are monitored by default.
+The following table lists the metrics available for this monitor.
+Metrics that are categorized as [container/host/bundled](https://docs.signalfx.com/en/latest/admin-guide/usage.html#about-custom-bundled-and-high-resolution-metrics)
```
barbara-sfx: Bundled does not apply to these metrics. The language to be used above these tables was developed and approved a few weeks ago in discussions with Joe D. Also, it doesn't make sense to link to that doc section (here or in the table header) since Default means they are NOT custom.

docs/monitors/appmesh.md
```diff
@@ -131,80 +135,41 @@ The following table lists the metrics available for this monitor. Metrics that a
...
-    - upstream_rq_pending_total
-    - upstream_rq_per_try_timeout
-    - upstream_rq_retry_overflow
-    - upstream_rq_rx_reset
-    - upstream_rq_timeout
-    - upstream_rq_total
-    - upstream_rq_tx_reset
-    monitorType: appmesh
-```
+**The following information applies to the agent version 4.7.0+ and has
```
barbara-sfx: Run-on; need to break out into a bullet list of phrases separated by semi=colons.  like " The following information assumes: You are using agent version 4.7.0 or higher; you have set... to true (link to where that's documented); and there are no references to .... in metricsToExclude..."

docs/monitors/aspdotnet.md
```diff
@@ -33,37 +33,42 @@ Monitor Type: `aspdotnet`
...
-| `asp_net_applications.pipeline_instance_count` | gauge |  | Number of instances in the ASP.NET application pipeline. |
-| `asp_net_applications.requests_failed` | gauge |  | Count of failed requests in the ASP.NET application |
-| `asp_net_applications.requests_sec` | gauge |  | Rate of requests in the ASP.NET application per second. |
-| `asp_net_applications.session_sql_server_connections_total` | gauge |  | Number of connections to microsoft sql server by an ASP.NET application. |
-| `asp_net_applications.sessions_active` | gauge |  | Number of active sessions in the ASP.NET application. |
-
-
-
+The following table lists the metrics available for this monitor.
+This monitor emits all metrics by default; however, **none are categorized as [container/host/bundled](https://docs.signalfx.com/en/latest/admin-guide/usage.html#about-custom-bundled-and-high-resolution-metrics)**.
```
barbara-sfx: IOW they are all custom? Then say that. And I wouldn't use a different format for the table. Just use the same format and leave Default column blank.

docs/monitors/aspdotnet.md
```diff
@@ -33,37 +33,42 @@ Monitor Type: `aspdotnet`
...
+| `asp_net_applications.errors_total_sec` | gauge | Error rate per second for the given ASP.NET application. |
+| `asp_net_applications.errors_unhandled_during_execution_sec` | gauge | Unhandled error rate per second countered while an ASP.NET application is running. |
+| `asp_net_applications.pipeline_instance_count` | gauge | Number of instances in the ASP.NET application pipeline. |
+| `asp_net_applications.requests_failed` | gauge | Count of failed requests in the ASP.NET application |
+| `asp_net_applications.requests_sec` | gauge | Rate of requests in the ASP.NET application per second. |
+| `asp_net_applications.session_sql_server_connections_total` | gauge | Number of connections to microsoft sql server by an ASP.NET application. |
+| `asp_net_applications.sessions_active` | gauge | Number of active sessions in the ASP.NET application. |
+
+
+The agent does not do any built-in filtering of metrics coming out of this
```
barbara-sfx: Is that another way or repeating what you said above the table? If so, does it need to be here twice?

docs/config-schema.md
keitwb: Ok, I tried to clarify it a bit.  See if it makes more sense to you now.

docs/monitor-config.md
keitwb: No, but I added a description of what it is here in this field doc.

docs/monitors/appmesh.md
keitwb: Removed the word "bundled" from this but I think we should link to those docs somewhere in here to make it very clear why the difference matters.

docs/monitors/aspdotnet.md
keitwb: The reason we omitted the column in the case of monitors like this is because they actually emit _all_ metrics by default (as the doc says) but _none_ of the metrics are actually _default_/_non-custom_ in our backend, so is it potentially confusing to have the column and have the meaning overloaded like that.  So we just went with having the _Default_ column only when there was a 1:1 correspondence between the agent's default and our categorization so that the _Default_ column has a consistent meaning.

docs/monitors/aspdotnet.md
keitwb: Removed.

keitwb:
@barbara-sfx continue to do a review on this PR if you have additional comments.  We are going to merge it to do a release but can make incremental improvements as needed.

docs/monitors/aspdotnet.md
barbara-sfx: Well I just ran across this, which in fact has an empty default column. https://docs.signalfx.com/en/latest/integrations/agent/monitors/internal-metrics.html  I think it is confusing to users to have what looks like the same table format in a bunch of places, but in some cases it's clear whether something is default or not, and in the other it isn't; you have to read the text to understand what the table is showing you. First rule of tech writing: consistency.

keitwb: That is going to be fixed as Mark just recategorized those to be bundled, so they will now show check boxes for all metrics.

barbara-sfx: I don't want to be the word police, but I really suggest folks stop using the word "bundled," since it doesn't match how we document bundled and it could confuse, say, a new person coming aboard and trying to understand how this thing works. I liked the term "included" as the column header, but "default" is also fine. Just sayin' ....

PR #798, opened 2019-05-10
Add agent image digest to release notes

jchengsfx:
Agent image digest will be appended to the release message and will look like:

 - Added JSON log output option
 - Added statsd timerCount config option
 - Added OpenShift cluster monitor

Breaking Changes: none

> Docker Image: `quay.io/signalfx/signalfx-agent:4.6.2` (digest: `sha256:4fa05ff59b2ee2c4ab8464fc7304ffad18d6640cdd9dcd655f4c002b341c9449`)

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -33,6 +33,10 @@ parse_args_and_release() {
   local allow_uncommitted=no
   local github_username="${GITHUB_USERNAME-}"
   local github_token="${GITHUB_TOKEN-}"
+  local github_release=no
+  local new_image=
+  local image_base=
+  local image_digest=
 
   while [ -n "${1-}" ]; do
     case $1 in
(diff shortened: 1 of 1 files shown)


scripts/release
```diff
@@ -394,4 +395,48 @@ release_python_to_pypi() {
   popd
 }
 
+add_digest_to_release() {
+  local tag=$1
+  local new_image=$2
+  local image_base=$3
+  local username=$4
+  local token=$5
+  local digest=$(docker inspect --format='{{.RepoDigests}}' $new_image | sed "s|\[${image_base}@\(sha256:.*\)\]|\1|")
```
keitwb: Split this out to a dedicated function that echos the sha256 sum for a given image.

scripts/release
```diff
@@ -102,6 +107,40 @@ parse_args_and_release() {
...
+    fi
+    if [[ "$stage" != "final" ]]; then
+      echo "Release stage must be 'final' in order to create a Github release." >&2
+      exit 1
+    fi
+    if ! docker pull $new_image; then
+      echo "$new_image must already be built and pushed in order to create a Github release." >&2
+      exit 1
+    fi
+    image_digest=$(get_image_digest $new_image)
```
keitwb: What is the advantage of doing this here and not just always waiting until right before creating the release?

jchengsfx: > What is the advantage of doing this here and not just always waiting until right before creating the release?

Leftover from testing. Will remove.

PR #812, opened 2019-05-14
Add tests for collectd-docker, collectd-mysql

seonsfx:


tests/monitors/collectd_mysql/mysql_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def create_db_activity(host):
+    conn = get_mysql_connection(host)
+    create_alter_table(conn)
+    insert_rows(conn)
+    update_rows(conn)
+    select_rows(conn)
+    delete_rows(conn)
+
```
jaysfx: Should it close conn here?

tests/monitors/collectd_mysql/mysql_test.py
```diff
@@ -0,0 +1,97 @@
...
+        create_db_activity(host_ip)
+
+        run_agent_verify_included_metrics(
+            f"""
+            monitors:
+            - type: collectd/mysql
+              host: {host_ip}
+              port: 3306
+              username: {ENV.get("MYSQL_USER")}
```
jaysfx: I think `ENV["MYSQL_USER"]` would be better to prevent silently putting `None` in there if you always expect those values to be set in `ENV`.

jaysfx: Another option would be to do
```
"""
username: {MYSQL_USER}
...
""".format(**ENV)
```

tests/monitors/collectd_mysql/mysql_test.py
```diff
@@ -0,0 +1,97 @@
...
+    create_alter_table(conn)
+    insert_rows(conn)
+    update_rows(conn)
+    select_rows(conn)
+    delete_rows(conn)
+
+
+def get_mysql_connection(host):
+    return mysql.connector.connect(
+        host=host, user="root", passwd=ENV.get("MYSQL_ROOT_PASSWORD"), database=ENV.get("MYSQL_DATABASE")
```
jaysfx: Same here as well.

tests/monitors/collectd_mysql/mysql_test.py
```diff
@@ -0,0 +1,98 @@
...
+from tests.helpers.metadata import Metadata
+from tests.helpers.util import container_ip, ensure_always, run_container, wait_for
+from tests.helpers.verify import (
+    verify,
+    verify_expected_is_subset,
+    run_agent_verify_included_metrics,
+    run_agent_verify_all_metrics,
+)
+
+pytestmark = [pytest.mark.docker_container_stats, pytest.mark.monitor_without_endpoints]
```
jaysfx: Sorry I didn't notice this before but looks like the docker mark got copied over here. Otherwise looks good.

seonsfx: Oops! my bad. Just fixed it. Thanks for catching!

PR #813, opened 2019-05-14
Link checker for markdown docs

jchengsfx:


test-services/check-links/run.sh
```diff
@@ -0,0 +1,23 @@
...
+
+nfailed=0
+for f in $(find "$REPO_ROOT_DIR" -name "*.md" -not -path "$REPO_ROOT_DIR/vendor/*" | sort); do
+    markdown-link-check -c "$REPO_ROOT_DIR/test-services/check-links/config.json" -qv $f
+    (( nfailed += $? ))
+done
+
+[ $nfailed -eq 0 ] || echo -e "\nERROR: dead links found! See output above for details." >&2
+
+exit $nfailed
```
jaysfx: Since the max exit code is 255 if there were 256 errors it'd wrap to exit code 0. Unlikely I suppose.

jchengsfx: > Since the max exit code is 255 if there were 256 errors it'd wrap to exit code 0. Unlikely I suppose.

Good point. I'll just `exit 1` just in case.

test-services/check-links/config.json
```diff
@@ -1,7 +1,7 @@
 {
     "ignorePatterns": [
         {
-            "pattern": "^https://(dl|ingest|api)\\.signalfx\\.com"
+            "pattern": "^https://.*\\.signalfx\\.com"
```
jaysfx: What about links to docs.signalfx.com that might be invalid?

jchengsfx: > What about links to docs.signalfx.com that might be invalid?

Ah, thanks, forgot about that. Will change.

test-services/check-links/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:8.16.0-alpine
```
keitwb: This would be more appropriate under the `scripts/docs` directory as it isn't really used by the integration tests.

docs/monitors/elasticsearch.md
```diff
@@ -122,7 +122,7 @@ monitors:
 
 For more information on the built-in content we have for Elasticsearch,
 [see
-here](https://github.com/signalfx/integrations/tree/master/elasticsearch)
+here](https://github.com/signalfx/integrations/tree/master/collectd-elasticsearch)
```
keitwb: This is a reversion back to the old directory.  See https://github.com/signalfx/integrations/pull/686

jchengsfx: > This is a reversion back to the old directory. See [signalfx/integrations#686](https://github.com/signalfx/integrations/pull/686)

Will revert when signalfx/integrations#686 is merged.

test-services/check-links/Dockerfile
jchengsfx: > This would be more appropriate under the `scripts/docs` directory as it isn't really used by the integration tests.

Will move to `scripts/docs` in a new PR.

PR #814, opened 2019-05-15
Kafka testing and metadata fixes

jaysfx:
* collectd/kafka metrics include status was synced against categorizer
* fixed unclean elections rate collection
* added missing log flush mbean collection
* bunch of metric names fixed


tests/monitors/kafka/kafka_test.py
```diff
@@ -1,112 +1,163 @@
...
+    random_hex,
+    container_hostname,
+)
+from tests.helpers.verify import verify
 
 pytestmark = [pytest.mark.collectd, pytest.mark.kafka, pytest.mark.monitor_with_endpoints]
+KAFKA_METADATA = Metadata.from_package("collectd/kafka")
+PRODUCER_METADATA = Metadata.from_package("collectd/kafkaproducer")
+CONSUMER_METADATA = Metadata.from_package("collectd/kafkaconsumer")
+VERSIONS = ["0.9.0.0", "0.10.0.0", "0.11.0.0", "1.0.0", "1.0.1", "1.1.1", "2.0.0"]
```
seonsfx: Any basis for all these versions? 3 `1.x`s feel like redundant to me if there's no significant diffs.

jaysfx: I have no idea, I just kept what was there.

jaysfx: @asuresh4 any thoughts?

seonsfx: Got no comment if that's the case. Thanks

asuresh4: @seonsfx @jaysfx - can't remember what the rationale was when I added these tests, maybe there was none. Looking at the release notes for kafka, I think we should be ok without the test for `1.0.1`.

PR #815, opened 2019-05-15
Update puppet repos in tests

jchengsfx:


tests/deployments/puppet/images/Dockerfile.amazonlinux1
```diff
@@ -1,6 +1,6 @@
 FROM amazonlinux:1
 
-RUN rpm -Uvh https://yum.puppet.com/puppet/puppet-release-el-6.noarch.rpm
+RUN rpm -Uvh https://yum.puppetlabs.com/puppet-release-el-6.noarch.rpm
```
jaysfx: Their docs say `https://yum.puppet.com/puppet6-release-el-6.noarch.rpm`. Looks they just changed the path (they removed `/puppet/` but the domain should probably stay the same since just Puppet is their new name.

jchengsfx: Changed to `https://yum/puppet.com`

PR #833, opened 2019-05-28
Update GenericJMX to support multiple Attributes

MovieStoreGuy:
There are cases were the underlying collectd agent would allow for multiple attributes being defined but currently isn't expressible with the current config.

Taking this collectd conf for tomcat: https://gist.github.com/pmauduit/c1ea230b81302001f6bc
You can see some values have attributes twice.


internal/monitors/expvar/expvar.go
```diff
@@ -18,6 +19,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const monitorType = "expvar"
```
keitwb: This is perhaps leftover code in your clone from when you did #696? Should be removed here.

keitwb:
See #834

PR #836, opened 2019-05-29
Add cluster property sync to host dimensions at startup

keitwb:
If the user specifies the `cluster` config option, the agent will add
this property to the set of properties on the host-id dimensions
identified by the agent.  If there is a platform-specific dimension
(e.g. AWSUniqueId), then the property will only be set on that dimension
and *not* on `host`.

Right now there is no retry on the dimension update so if it fails the
cluster property won't be attempted to be set until the next agent
restart.


docs/config-schema.md
```diff
@@ -380,6 +381,7 @@ where applicable:
   intervalSeconds: 10
   globalDimensions: 
   sendMachineID: false
+  cluster:
```
mpetazzoni: The indent on some of the lines of this example seem off?

internal/core/agent.go
```diff
@@ -204,3 +210,26 @@ func StreamDatapoints(configPath string, metric string, dims string) (io.ReadClo
...
+func startSyncClusterProperty(propertyChan chan *types.DimProperties, cluster string, hostDims map[string]string) {
+	for dimName, dimValue := range hostDims {
+		if len(hostDims) > 1 && dimName == "host" {
+			// If we also have a platform-specific host-id dimension that isn't
+			// the generic 'host' dimension, then skip setting the property on
+			// 'host' since it tends to get reused frequently. The property
+			// will still show up on all MTSs that come out of this agent.
+			continue
+		}
+		log.Infof("Setting cluster property on '%s' dimension", dimName)
```
mpetazzoni: Might be worth logging `dimValue` and `cluster` as well:

```go
log.Infof("Setting cluster:%s property on %s:%s dimension...", cluster, dimName, dimValue)
```

internal/core/writer/properties/client.go
```diff
@@ -195,6 +225,41 @@ func (dpc *DimensionPropertyClient) isDuplicate(dimProps *types.DimProperties) b
...
+		return nil, nil, err
+	}
+
+	resp, err := dpc.client.Get(url.String())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if resp.StatusCode != 200 {
+		// No existing dimension value, so return blank for both.
```
mpetazzoni: Checking `!= 200` for this is dangerous. No dimension object would be a 404; you may get other status codes when throttled, or even a 502 or 503 if SignalFx as a partial outage. In which case you don't want to end up sending a PUT with none of the original metadata.

docs/config-schema.md
keitwb: It seems ok looking at it here in Github (and in my text editor).  This is basically a dump of the "default" config.

internal/core/writer/properties/client.go
keitwb: Ok, made it so that only a 404 returns the empty property/tag set.  Anything >=500 returns error, which now causes an indefinite retry with exponential backoff (had to refactor some things to make it not block other non-merging dim updates).  Am now basically assuming that anything not a 404 or 5xx response is "ok", although subsequent json parse errors will also trigger an error and thus the retry.

PR #837, opened 2019-05-29
Do not sync AWS properties to host dimension

asuresh4:


asuresh4:
@keitwb - yes, that's what we agreed on, not to sync AWS properties to host dimension. Is it possible to sync these properties to AWSUniqueId instead of host dimension from the agent too?

PR #841, opened 2019-06-03
Handle PS command line args for target

jchengsfx:


scripts/windows/make.ps1
```diff
@@ -40,12 +40,20 @@ function signalfx-agent([string]$AGENT_VERSION="", [string]$AGENT_BIN=".\signalf
     compile_deps
 
     go build -mod vendor -o "$AGENT_BIN" github.com/signalfx/signalfx-agent/cmd/agent
+
+    if (!(Test-Path -Path "$AGENT_BIN")) {
```
keitwb: Why is this needed?  If the above command fails, it should exit the script anyway right?  If not, is there some equivalent of the `set -e` flag from bash in powershell?

jchengsfx: `$ErrorActionPreference = "Stop"` is set at the top of the script, but it was still continuing even though there were build issues when i was testing.

keitwb: Ok, can you investigate further and try and understand why not?  Mysterious behavior like that might be hiding a bug or something.

jchengsfx: https://stackoverflow.com/questions/47032005/why-does-a-powershell-script-not-end-when-there-is-a-non-zero-exit-code-using-th
`ErrorActionPreference` only affects internal powershell commands and ignores exit codes from external commands (e.g. `go build`).  So we'll have handle these errors individually since there doesn't seem to be a "global" way to do this in powershell like `set -e` in bash.

PR #845, opened 2019-06-06
infra nav metrics

bjsignalfx:


mstumpfx:
Rules file is in place, this is good to go.

PR #846, opened 2019-06-07
Fix typing issue with dimension filters with list values

keitwb:


tests/basic/new_dp_filter_test.py
```diff
@@ -33,3 +35,24 @@ def no_filtered_metrics():
         assert "sfxagent.go_frees" in metrics_received
         assert "sfxagent.go_heap_inuse" in metrics_received
         assert "sfxagent.go_heap_released" in metrics_received
+
+
+def test_filtering_by_dimensions():
+    with run_service("nginx") as nginx_cont:
+        with Agent.run(
+            dedent(
```
jaysfx: We can get rid of dedents.

internal/core/config/filters.go
```diff
@@ -40,8 +40,12 @@ func (mf *MetricFilter) Normalize() (map[string][]string, error) {
 	dimSet := map[string][]string{}
 	for k, v := range mf.Dimensions {
 		switch s := v.(type) {
-		case []string:
-			dimSet[k] = s
+		case []interface{}:
+			newSet := []string{}
+			for _, iv := range s {
+				newSet = append(newSet, fmt.Sprintf("%v", iv))
```
jaysfx: Can you not check of s.(type) is string? Or is it actually a non-string value?

keitwb: If somebody gives an integer, for example, I'll just take it and convert to a string rather than trying to worry about validating it.

Issue #847, opened 2019-06-10
Interface monitor does not collect metrics from other network namespaces


MovieStoreGuy:
I currently have the agent running inside a docker container, in privilege mode but not running within the host network. The network statistics produced is only for the network that the container is current running in instead of the host.

I was wondering if there is some magic that could be done either in code or config to have it collect from other namespaces?

keitwb:
You could run it with the docker run flag `--net host` to run the agent in the host network, but you still would only get stats for the root network namespace and none specific to the containers.  However, on the most standard container network setup, the container's net namespace contains a veth device that is paired with a specific veth device in the host/root network namespace.   So, if you collect interface stats at the host level you should get a pretty full view of things, assuming all traffic flows through the root namespace.

The simplest way to collect everything is to use this monitor config:

```
  - type: collectd/custom
    template: |
        LoadPlugin interface
        <Plugin "interface">
        </Plugin>
```

The template that gets used for the `collectd/interface` monitor filters out a bunch of stuff by default so I would avoid it.

It collects the metrics via the `/proc` filesystem and there is currently no way to override the path of the procfs filesystem to look at, so you are pretty much limited to monitoring one network namespace with that collectd plugin.

MovieStoreGuy:
Unfortunately within our infrastructure, I am not able to run it within the host network. It's a shame because it is an easy fix.

Do you know if there is any way of running a pre startup step without hacking the docker entry?
I believe you can switch namespaces within the container using some shell commands, not sure if the smart agent is expressive enough to support that in config.

keitwb:
If you aren't able to run in the host network, I doubt you would be able to do any of the more sophisticated things like access other container's namespaces from a given container (it would kind of defeat the point of the restriction your sys admins have put in place of not being able to run in the host namespace).

If you can manage to mount in the host's `/proc` directory to somewhere in your agent container, you can write a [python-monitor](https://github.com/signalfx/signalfx-agent/blob/master/docs/monitors/python-monitor.md#python-monitor) that scrapes the interface data from the init process's net namespace.  If you mount the host procfs to `/hostproc` in the agent container, you should be able to parse the file `/hostproc/1/net/dev`, which is a nice table of interfaces with a bunch of stats about them.  This is basically all the collectd interface plugin does on Linux, except that it is hardcoded to only read the file `/proc/net/dev`.  We would also welcome any PRs to patch up the interface plugin in collectd to accept a configurable path to the net/dev table :)

[...]

PR #848, opened 2019-06-10
Enhance config for collectd/interface monitor

keitwb:
Allow configuring the list of interfaces


keitwb:
Might be helpful for #847

PR #851, opened 2019-06-10
[INT-1076] Exclude http and https links from md -> html conversion

jaysfx:
Only relative links should be converted.

### DIFF
--- a/scripts/docs/to-product-docs
+++ b/scripts/docs/to-product-docs
@@ -15,7 +15,7 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 AGENT_REPO_BASE="$SCRIPT_DIR/../.."
 PRODUCT_DOCS_REPO="${PRODUCT_DOCS_REPO:-$AGENT_REPO_BASE/../product-docs}"
 AGENT_PRODUCT_DOCS="$PRODUCT_DOCS_REPO/integrations/agent"
-MANAGE_GIT="${MANAGE_GIT-'yes'}"
+MANAGE_GIT="${MANAGE_GIT-yes}"
 
 check_pandoc_version() {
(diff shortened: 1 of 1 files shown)


scripts/docs/to-product-docs
```diff
@@ -39,12 +39,12 @@ convert_relative_links() {
...
 # Changes .md extensions to .html
 convert_doc_extensions() {
-  sed -E -e '/(development|signalfx-agent.1|MANUAL).md/! s/\.md(#.*)?[[:>:]]/.html\1/g' -e'/(development|signalfx-agent.1).md/! s@\./docs/@./@g'
+  sed -E -e '/http:|https:|(development|signalfx-agent.1|MANUAL).md/! s/\.md(#.*)?[[:>:]]/.html\1/g' -e'/(development|signalfx-agent.1).md/! s@\./docs/@./@g'
 }
 
 reset_repo() {
   if ! git -C $PRODUCT_DOCS_REPO diff --exit-code; then
-    echo "Product docs repo has uncommitted changes, please deal with them first" >&2 
+    echo "Product docs repo has uncommitted changes, please deal with them first" >&2
```
barbara-sfx: Hi - what does this mean? Is this a script to push directly to master? Or does it mean the specific branch you are trying to push to has uncommitted changes?

jaysfx: I just trimmed the whitespace so shouldn't have a functional change. Apparently it's a piece of broken code though, pushed a change to fix it. @keitwb is reset_repo still valid?

keitwb: You have to set `MANAGE_GIT=yes` in the environment, but yes it should theoretically work.  It just wipes out any other changes in product-docs locally so you have a fresh copy.  We could also use the git worktree thing I was telling you about.

PR #853, opened 2019-06-11
Truncate dimension values to 256 chars

keitwb:
That appears to be the limit on the backend


barbara-sfx:
Could this lead to duplicate dimension names?

keitwb:
Two excessively long original dimension values could be truncated to the same thing, but the alternative is that the whole datapoint is dropped.

PR #856, opened 2019-06-14
Force persmissions on gpg key to be default after downloading

mstumpfx:
We assume that the default unmask on user's systems will be 644. If they have modified it (e.g. 640 not world readable) then the install script fails. This is an assertion makes sure the permissions on the key will be default 644.

### DIFF
--- a/deployments/installer/install.sh
+++ b/deployments/installer/install.sh
@@ -209,6 +209,7 @@ download_debian_key() {
     echo "Could not get the SignalFx Debian GPG signing key" >&2
     exit 1
   fi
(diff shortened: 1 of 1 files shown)


deployments/installer/install.sh
```diff
@@ -337,6 +337,7 @@ install() {
     ubuntu|debian)
       if [ "$stage" != "test" ]; then
         download_debian_key
+        chmod 644 /etc/apt/trusted.gpg.d/signalfx.gpg
```
keitwb: This conceptually goes with `download_debian_key`.

PR #857, opened 2019-06-17
Make metric docs a list instead of table

keitwb:
This makes formatting easier with long metric names.

Also metrics are grouped in the docs by their metric group, if present.

Also expand cassandra docs from integrations repo.


jaysfx:
I'm not sure what it looks like in product docs but the rendered markdown in github looks pretty jarring with the mixed plain and bold/itatlics. Could maybe group into two separate lists instead for custom/default? Although with the groups grouping that's a lot of sections.

I also think having all the metric descriptions all start at different places makes it look really messy as well.

One thought is to keep the table and put the description with in the metric cell but only have the columns default and maybe type but I think type looks ok here as well.

keitwb:
Another problem this solves is that descriptions with special formatting (e.g. lists, multiple paragraphs) are very ugly in tables and require special html markup (`<br>`) to not break table parsing.  Tables in markdown are quite inflexible.

keitwb:
Changed it so the description starts on a new line.  Take a look again.

keitwb:
Open to suggestions on how to better render the default/non-default distinction without a table.

jaysfx:
If it's on a separate line I'd indent the description as well. What are your thoughts on separate sections for default and custom? I don't see any great option.

jaysfx:
I'd also remove the - if it's on a separate line.

keitwb:
I took away the dash but left them unindented (the indent threw off alignment along the left and looked awkward).

The problem with sections for custom/default is that you can't quickly identify what a metric is without establishing context within the page, and I want this to be as quick and convenient a reference as possible.

jaysfx:
I think we could use a definition list. I just tested it and it rendered in product docs:

```html
<dl>
  <dt><strong>container_spec_cpu_period</strong></dt>
  <dd>In CPU quota for the CFS process scheduler. In K8s this is equal to the containers's CPU limit as a fraction of 1 core and multiplied by the `container_spec_cpu_period`.  So if the CPU limit is `500m` (500 millicores) for a container and the `container_spec_cpu_period` is set to 100,000, this value will be 50,000.</dd>
</dl>
````

Screen Shot 2019-06-17 at 11 30 26 AM

jaysfx:
Well, any markdown inside it would need to be rendered to html somehow, like ` to code. So guess that's an issue.

keitwb:
Tried definition lists and confirmed that at least Github won't render special syntax (e.g. ```, `*`) within HTML tags, so that is super limiting.

jaysfx:
I wouldn't think we would need ``` within metric descriptions but definitely code, b, and italics. Would have to render the descriptions to html during doc generation which would probably not be easy with the current build process.

keitwb:
> I wouldn't think we would need ``` within metric descriptions but definitely code, b, and italics. Would have to render the descriptions to html during doc generation which would probably not be easy with the current build process.

Yeah, sorry that was meant to be a literal single backtick, not the triple backticks.

PR #859, opened 2019-06-18
Enhance Consul docs

keitwb:
Also add script to pull metric descriptions from integrations repo


scripts/docs/metric-doc-to-agent
```diff
@@ -1,33 +1,55 @@
...
-  echo
-done
+# INTEGRATIONS_REPO=~/work/integrations ./metric-doc-to-agent collectd-kafka internal/monitors/collectd/kafka
+
+SCRIPT_DIR = Path(__file__).parent.resolve()
+
+INTEGRATION = sys.argv[1]
+METADATA_DIR = Path(sys.argv[2]).resolve()
+
+INTEGRATIONS_REPO = os.environ.get("INTEGRATIONS_REPO", SCRIPT_DIR / "../../../integrations")
```
jaysfx: Need to wrap it so that if from env var it's a Path since it'll be a Path in the default case:

```suggestion
INTEGRATIONS_REPO = Path(os.environ.get("INTEGRATIONS_REPO", SCRIPT_DIR / "../../../integrations"))
```

scripts/docs/metric-doc-to-agent
```diff
@@ -1,33 +1,55 @@
...
+
+if not os.path.isdir(INTEGRATIONS_REPO):
```
jaysfx:
```suggestion
if not INTEGRATIONS_REPO.is_dir():
```

scripts/docs/metric-doc-to-agent
```diff
@@ -1,33 +1,55 @@
...
+metadata_path = METADATA_DIR / "metadata.yaml"
+metadata = yaml.round_trip_load(metadata_path.read_bytes())
+if len(metadata) != 1:
+    print("This script not meant for metadata.yaml with more than one monitor", file=sys.stderr)
+    sys.exit(2)
+
+metrics = metadata["monitors"][0]["metrics"]
+
+for f in (INTEGRATIONS_REPO / INTEGRATION).glob("docs/*.md"):
+    if "readme.md" in str(f):
```
jaysfx: Does it have to be in or can it be ==?

```suggestion
    if f.name == "readme.md"
```

PR #862, opened 2019-06-20
Pull in better collectd/df metric docs from integrations repo

keitwb:


docs/monitors/collectd-df.md
```diff
@@ -38,28 +44,28 @@ Metrics that are categorized as
 (*default*) are ***in bold and italics*** in the list below.
 
 
- - ***`df_complex.free`*** (*gauge*)<br>    Free disk space in bytes
- - `df_complex.reserved` (*gauge*)<br>    Reserved disk space in bytes
- - ***`df_complex.used`*** (*gauge*)<br>    Used disk space in bytes
- - `percent_inodes.free` (*gauge*)<br>    Free inodes on the file system, expressed as a percentage.
- - `percent_inodes.reserved` (*gauge*)<br>    Reserved inodes on the file system, expressed as a percentage.
- - `percent_inodes.used` (*gauge*)<br>    Used inodes on the file system, expressed as a percentage.
+ - ***`df_complex.free`*** (*gauge*)<br>    This metric measures free disk space in bytes on this file system.
```
jaysfx: I realize you just took these from the integrations repo but I just want to point out it's redundant to say "This metric measures..." but every single metric description. Maybe whoever wrote it chose to do it so it was a complete sentence but I think it adds a lot of noise.

keitwb: Stripped off the leading "This metric ".

keitwb: It kind of needs to be a sentence if we are going to put a period at the end of it, esp in the cases where there are multiple sentences.

jaysfx: "Measures free inodes as a percentage of total inodes in the file system." isn't an actual sentence though right? I think it sounds fine for these though since it reads as  "df_complex.free measures free inodes as a percentage of total inodes in the file system"

keitwb: Yes, it is a sentence with an implicit subject of "This".

PR #863, opened 2019-06-24
Add k8s v1.15 to integration tests

jchengsfx:


.circleci/config.yml
```diff
@@ -308,7 +308,7 @@ jobs:
     environment:
       K8S_VERSION: "<< parameters.k8s_version >>"
       K8S_MIN_VERSION: "v1.8.0"
```
keitwb: We should drop 1.8 support if 1.15 is out.

PR #865, opened 2019-06-27
Add agent image security scan to circleci

jchengsfx:


keitwb:
If you run the scanner with all levels enabled, does it even output anything?  I would be surprised if it did since our agent image is based on a scratch image, which doesn't have any package manager in it, and also doesn't really use standard paths for stuff.  Whatever scanner quay.io uses just blatantly doesn't work on our image because of that: https://quay.io/repository/signalfx/signalfx-agent?tag=latest&tab=tags.

Probably the best thing to do for security scanning is to scan the repo source directly somehow.

.circleci/config.yml
```diff
@@ -537,6 +542,56 @@ jobs:
           fi
           make check-links
 
+  owasp_image_scan:
```
keitwb: I would omit this unless you can actually confirm that it does anything useful with the scratch image that we use.

scripts/dependency-check/scan_image.sh
```diff
@@ -0,0 +1,38 @@
...
+
+docker build -t dependency-check "$SCRIPT_DIR"
+
+set +e
+
+docker run --rm \
+    -v "$BUNDLE_DIR":/bundle \
+    -v "$CACHE_DIR":/usr/share/dependency-check/data \
+    -v "$REPORTS_DIR":/reports \
+    dependency-check \
```
keitwb: Make this the thing in the Makefile.  There is a lot of information encoded in these command line flags that should be more visible.

scripts/dependency-check/scan_source.sh
```diff
@@ -0,0 +1,25 @@
...
+
+mkdir -p "$REPORTS_DIR"
+
+docker build -t dependency-check "$SCRIPT_DIR"
+
+docker run --rm \
+    -v "$REPO_DIR":/src \
+    -v "$CACHE_DIR":/usr/share/dependency-check/data \
+    -v "$REPORTS_DIR":/reports \
+    dependency-check \
```
keitwb: Same thing here.

.circleci/config.yml
jchengsfx: > I would omit this unless you can actually confirm that it does anything useful with the scratch image that we use.

This does seem to catch some issues if I don't use the suppression.xml below. I need to investigate some more to see if the issues are actually valid or false positives.

jchengsfx: https://circleci.com/gh/signalfx/signalfx-agent/153822#tests/containers/ and https://153822-80042799-gh.circle-artifacts.com/0/home/circleci/project/test_output/dependency-check-report.html show some vulnerabilities found (the job currently only fails if critical vulns are found).

PR #868, opened 2019-07-02
Update couchbase monitor doc

mstumpfx:
The link to the `metric_info.py` is misleading because the "default" vs "detailed" metrics in that list do *not* match our version of "standard" vs "custom".

Removing to link to that list of metrics will prevent confusion and encourage users to use the list of metrics below instead.

Also, fixed a broken conviva link.


mstumpfx:
@keitwb ok i reverted my commit to fix that link, but the build test is going to fail on this PR from that.

PR #874, opened 2019-07-08
Handle unique ids from Azure Scale Set VM

asuresh4:
- Use newer version of instance metadata endpoint `2018-10-01`
- Make `azure_resource_id` correctly for Azure Scale Set VMs

Followed: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service

Tested manually by starting up an agent with the changes both in a standalone VM and also a VM in a Scale Set.

### DIFF
--- a/internal/core/hostid/azure.go
+++ b/internal/core/hostid/azure.go
@@ -5,16 +5,19 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 )
 
 // AzureUniqueID constructs the unique ID of the underlying Azure VM.  If
 // not running on Azure VM, returns the empty string.
+// Details about Azure Instance Metadata ednpoint:
+// https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service
 func AzureUniqueID() string {
 	c := http.Client{
 		Timeout: 1 * time.Second,
 	}
-	req, err := http.NewRequest("GET", "http://169.254.169.254/metadata/instance?api-version=2017-08-01", nil)
+	req, err := http.NewRequest("GET", "http://169.254.169.254/metadata/instance?api-version=2018-10-01", nil)
 	if err != nil {
 		return ""
 	}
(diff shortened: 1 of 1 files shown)


internal/core/hostid/azure.go
```diff
@@ -49,5 +53,19 @@ func AzureUniqueID() string {
...
+	parsedName := strings.Split(compute.Doc.Name, "_")
+
+	// names of VM's in VMScalesets seem to follow the of `<scale-set-name>_<instance-id>`
+	// where scale-set-name is alphanumeric (and is the same as compute.vmScaleSetName
+	// field from the metadata endpoint)
+	if len(parsedName) != 2 {
+		return ""
+	}
+
+	return fmt.Sprintf("%s/%s/microsoft.compute/virtualmachinescalesets/%s/virtualmachines/%s", compute.Doc.SubscriptionID, compute.Doc.ResourceGroupName, compute.Doc.VMScaleSetName, parsedName[1])
```
keitwb: Is there logic already in our backend that also generates this for the Azure metric API integration?

internal/core/hostid/azure.go
```diff
@@ -49,5 +53,19 @@ func AzureUniqueID() string {
...
+		return fmt.Sprintf("%s/%s/microsoft.compute/virtualmachines/%s", compute.Doc.SubscriptionID, compute.Doc.ResourceGroupName, compute.Doc.Name)
+	}
+
+	parsedName := strings.Split(compute.Doc.Name, "_")
+
+	// names of VM's in VMScalesets seem to follow the of `<scale-set-name>_<instance-id>`
+	// where scale-set-name is alphanumeric (and is the same as compute.vmScaleSetName
+	// field from the metadata endpoint)
+	if len(parsedName) != 2 {
+		return ""
```
keitwb: Can you just do `strings.TrimLeft(compute.Doc.VMScaleSetName + "_")` instead?  It seems a bit flaky to fail out with no azure_resource_id if the name isn't exactly as you expect it with a Scale Set.

asuresh4: Updated to use `TrimLeft`

internal/core/hostid/azure.go
asuresh4: Marking as resolved as per our conversation

PR #876, opened 2019-07-09
Suse support

jchengsfx:
- Installer and chef tests will use rpm in test stage for opensuse15 until the changes are merged and released.
- Installer and chef changes were manually tested on SLES 12 and 15 VMs.

### DIFF
--- a/deployments/chef/metadata.rb
+++ b/deployments/chef/metadata.rb
@@ -3,13 +3,15 @@ maintainer 'SignalFx, Inc.'
 maintainer_email 'support@signalfx.com'
 license 'Apache-2.0'
 description 'Installs/Configures the SignalFx Agent'
-version '0.2.6'
+version '0.2.7'
 chef_version '>= 12.1' if respond_to?(:chef_version)
 
 supports 'amazon'
 supports 'centos'
 supports 'debian'
+supports 'opensuse'
 supports 'redhat'
+supports 'suse'
 supports 'ubuntu'
 supports 'windows'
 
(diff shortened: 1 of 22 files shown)


deployments/chef/metadata.rb
```diff
@@ -3,13 +3,15 @@
 maintainer_email 'support@signalfx.com'
 license 'Apache-2.0'
 description 'Installs/Configures the SignalFx Agent'
-version '0.2.6'
+version '0.2.7'
```
keitwb: I would go to 0.3.0, this is a pretty big addition.

keitwb: Also don't do the release increment until right before release.

deployments/chef/recipes/default.rb
```diff
@@ -24,13 +24,34 @@
...
+  tmpdir = Dir.mktmpdir
+  package_name = if node['signalfx_agent']['package_version'].nil?
+                   'signalfx-agent'
+                 else
+                   "signalfx-agent-#{node['signalfx_agent']['package_version']}"
+                 end
+  execute 'download-agent' do
+    command "zypper --pkg-cache-dir=#{tmpdir} download #{package_name}"
+  end
+  execute 'install-agent' do
```
keitwb: This needs conditional logic so it doesn't do this every time chef runs.  Same thing with the download command above.

deployments/chef/metadata.rb
jchengsfx: Reverted back to 0.2.6. Will increment to 0.3.0 when releasing to chef supermarket.

deployments/chef/recipes/default.rb
jchengsfx: Added logic to detect installed and latest versions, and only download/install when there is a mismatch.

PR #879, opened 2019-07-16
Update values.yaml

jonbarker-signalfx:
Two changes: make realm required (default is us0, trials are now not going to us0).  Change agent from collectd to smart agent.

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/values.yaml
+++ b/deployments/k8s/helm/signalfx-agent/values.yaml
@@ -61,7 +61,7 @@ agentConfig:
 # values of `ingestUrl`, `traceEndpointUrl` and `apiUrl` will be automatically
 # set based on this realm value.  If unset, it defaults to the original us0
 # realm in the agent.
-signalFxRealm:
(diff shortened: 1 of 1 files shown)


keitwb:
@jonbarker-signalfx I'm not sure we can make realm required since it would break any existing deployments of the chart that were relying on the default and have it "just work".  I think having it clear in the values.yaml file that it defaults to us0 (or even just setting the value to `us0`) and then telling people they need to override it with a comment is the best thing to maintain backwards-compatibility.

deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -110,7 +110,7 @@ kubeletAPI:
 
 # Any values put in this object correspond to the 'collectd' config block of
 # the agent config
-collectd: {}
+collectd: {}  #we should be setting up the smart agent not the collectd agent.
```
keitwb: The Smart Agent runs collectd as a sub-process, which is what this config is about.  Normally the defaults for it are fine.

PR #883, opened 2019-07-17
Puppet module for windows

asuresh4:


deployments/puppet/README.md
```diff
@@ -42,11 +42,15 @@ class accepts the following parameters:
  - `$config_file_path`: The path of the config file that will be rendered by the
    module (**default:** '/etc/signalfx/agent.yaml')
 
- - `$version`: The agent package version.  This is of the form `<agent
-	 version>-<package revision>` (e.g. package version `3.0.1-1` is the first
-	 package revision that contains the agent version `3.0.1`).  Releases with
-	 package revision > 1 contain changes to some aspect of the packaging
-	 scripts (e.g. init scripts) but contain the same agent bundle.
+ - `$agent_version`: The agent release version, in the form `1.1.1`.  This
```
keitwb: Clarify which one of these takes priority if both get set.

deployments/puppet/manifests/init.pp
```diff
@@ -1,15 +1,52 @@
...
+) {
+
+  $service_name = 'signalfx-agent'
 
   if !$config['signalFxAccessToken'] {
     fail("The \$config parameter must contain a 'signalFxAccessToken'")
   }
 
+  if $::osfamily == 'windows' {
+    $agent_location = "C:\\Program Files\\SignalFx\\"
```
keitwb: This could be changed by a config option (i.e. if somebody were trying to install it to a different drive), so it shouldn't be hardcoded here.

deployments/puppet/manifests/init.pp
```diff
@@ -1,15 +1,52 @@
...
+      ensure          => 'installed',
+      source          => "${agent_location}\\SignalFxAgent\\bin\\signalfx-agent.exe",
+      install_options => [{ '-service' => '"install"' }, '-logEvents', { '-config' => $config_file_path }],
+    }
+  }
+  else {
+    $split_config_file_path = $config_file_path.split("/")
+    $config_parent_directory_path = $split_config_file_path[0, - 2].join("/")
+
+    unless $package_version == 'latest' {
```
keitwb: What about if `agent_version` is set to something and `package_version` is left at the default of `latest`?  Where does that come into play?

asuresh4: Updated this so that `package_version` will take precedence if both are set.

deployments/puppet/manifests/init.pp
asuresh4: Added `installation_directory` config option

deployments/puppet/README.md
asuresh4: Added a note under `package_version` option definition

deployments/puppet/metadata.json
```diff
@@ -54,12 +62,19 @@
...
+        "2012 R2",
+        "2016"
+      ]
     }
   ],
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.7.0 < 6.0.0"
+      "version_requirement": ">= 4.7.0 <= 6.6.0"
```
keitwb: Will it not work with puppet newer than 6.6.0?

deployments/puppet/manifests/init.pp
```diff
@@ -1,15 +1,60 @@
...
 
+  if $::osfamily == 'windows' {
+    if $agent_version == '' {
+      fail("The \$agent_version parameter must be set to a valid SignalFx Agent version")
+    }
+
+    $split_config_file_path = $config_file_path.split("\\\\")
+    $config_parent_directory_path = $split_config_file_path[0, - 2].join("\\")
+
+    package { $service_name:
```
keitwb: Is this ok to be before the `win_repo` class below?

deployments/puppet/metadata.json
asuresh4: Our tests use 6.6.0. @jchengsfx thoughts?

jchengsfx: I've tested manually with 6.2.0, and integration tests for 5.0.0 (oldest from chocolatey) and 6.6.0 (current latest from chocolatey) have been successful.

deployments/puppet/manifests/init.pp
asuresh4: I think it is ok, because in line 88 we have  `-> Package[$service_name]` that indicates the ordering. So that `win_repo` is applied before `-> Package[$service_name]`

asuresh4: Screen Shot 2019-07-19 at 10 27 37 AM

@keitwb The above graph shows the relationship of resources. It's not very legible but you can see `win_repo` being a applied before the package resource.

deployments/puppet/metadata.json
asuresh4: Updated to ">= 5.0.0"

deployments/puppet/metadata.json
```diff
@@ -9,6 +9,14 @@
     {
       "name": "puppetlabs/apt",
       "version_requirement": ">= 4.1.0 <= 7.0.0"
+    },
+    {
+      "name": "puppet/archive",
+      "version_requirement": ">= 5.5.8 < 7.0.0"
```
jchengsfx: @asuresh4 are these versions correct? I only see 0.4.4 to 4.1.0 at https://forge.puppet.com/puppet/archive.

deployments/puppet/metadata.json
```diff
@@ -9,6 +9,14 @@
...
+    },
+    {
+      "name": "puppetlabs/powershell",
+      "version_requirement": ">= 4.7.0 < 7.0.0"
```
jchengsfx: https://forge.puppet.com/puppetlabs/powershell has 1.0.1 to 2.3.0 as available versions.

deployments/puppet/metadata.json
asuresh4: Oh I put the puppet versions in there. Let me update that

deployments/puppet/metadata.json
asuresh4: Same as above

deployments/puppet/metadata.json
asuresh4: corrected

deployments/puppet/metadata.json
asuresh4: corrected

PR #887, opened 2019-07-18
Ignore everything under /var/lib/docker

jaysfx:
There can be other directories under /var/lib/docker besides containers. A
directory named overlay2 was observed.


keitwb:
Docs have to be regenerated since default changed.

PR #894, opened 2019-07-25
K8s apiserver monitor

bjsignalfx:


internal/monitors/kubernetes/apiserver/metadata.yaml
```diff
@@ -0,0 +1,1443 @@
+monitors:
+- monitorType: kubernetes-apiserver
+  doc: |
+      ```
```
keitwb: Needs a bit more description on what this monitor is useful for.  Maybe include a link to that article you mentioned.

internal/monitors/kubernetes/apiserver/config.go
```diff
@@ -0,0 +1,80 @@
...
+			var body io.ReadCloser
+			defer func() {
+				if body != nil {
+					body.Close()
+				}
+			}()
+			if body, err = k8sClient.CoreV1().RESTClient().Get().RequestURI(c.MetricPath).Stream(); err != nil {
+				return
+			}
+			decoder := expfmt.NewDecoder(body, expfmt.FmtText)
```
keitwb: If you're going to abstract the client like this, at least don't duplicate the code that converts from the text body of the response to the metrics.  Have another method that focuses entirely on getting the text, and then share the same code that converts it to metrics.

bjsignalfx: ok

deployments/k8s/helm/signalfx-agent/templates/clusterrole.yaml
```diff
@@ -71,6 +71,12 @@ rules:
   - get
   - list
   - watch
+- nonResourceURLs:
+  - '*'
```
jaysfx: We should limit this to just `/metrics`.

bjsignalfx: I was under the impression that, that path is configurable. Isn't it? 🤔

jaysfx: Configurable in Kubernetes or in our monitor?

bjsignalfx: I made the path configurable in the monitor thinking that it's configurable in k8s.

jaysfx: It might be but I think /metrics is a pretty standard endpoint for prometheus metrics. I think it's better to limit it to /metrics and if it's wrong for their environment they can change the role.

bjsignalfx: ok

PR #895, opened 2019-07-27
Add exclude dimensions monitor configuration parameter

jcmackie:
Added a new configuration parameter to monitor configuration, giving the ability to explicitly drop dimensions by name from a datapoint before it gets sent. This is to deal with monitors that add potentially high cardinality tags with no way to disable them otherwise.

My editor also removed some whitespace from `config-schema.md`.

I also removed two lines from cadvisor plugin because if you read the code the lines actually do nothing. The high cardinality dimensions are `container-id` and `container-name` respectively.

### DIFF
--- a/docs/config-schema.md
+++ b/docs/config-schema.md
@@ -34,7 +34,7 @@ if not set.
 
 ## Config Schema
 
-  
+
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
 | `signalFxAccessToken` | no | string | The access token for the org that should receive the metrics emitted by the agent. |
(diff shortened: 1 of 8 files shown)


jcmackie:
I will fix the failing checks.

jcmackie:
Okay I'm pretty sure `docs_test` and `check_links` failures aren't due to my change.
Check_links seems to be because some markdown file in the Integrations repo is gone now, and `make docs` fails on master so... I'm not sure if I should be investigating or fixing them, I'll take a look however.

keitwb:
Also to be perfectly on the safe side we should probably do the `delete()` call in the loop that processes `dimensionTransformations` instead of relying on the implicit behavior of dimensions being dropped if they have a 0-length value.

jcmackie:
Thanks for the comments @keitwb.
If `dimensionTransformations` is the prefered method for doing this I will update the documentation, and make a `delete()` call inside that function to make the behaviour explicit.

Would you prefer a new PR, or should I just update this one?

jcmackie:
Closing this PR because it's superseded by my new PR: https://github.com/signalfx/signalfx-agent/pull/896

PR #897, opened 2019-07-29
Windows installer test and CI improvements

jchengsfx:


.azure-pipelines/templates/changes-include.yml
```diff
@@ -0,0 +1,21 @@
+parameters:
+  paths: ''
+
+steps:
+- bash: |
+    echo '##vso[task.setvariable variable=hasChanges]true'
```
keitwb: What is this magic?

keitwb: Definitely put a comment on it since most of us are not going to have any idea what this does.

jchengsfx: That's how azure pipelines passes variables between steps.

tests/deployments/chef/chef_test.py
```diff
@@ -168,30 +180,55 @@ def run_win_chef_client(backend, agent_version=None, stage=STAGE):
...
     installed_version = get_win_agent_version()
-    if agent_version:
-        assert installed_version == agent_version, "installed agent version is '%s', expected '%s'" % (
-            installed_version,
-            agent_version,
+    assert installed_version == agent_version, "installed agent version is '%s', expected '%s'" % (
+        installed_version,
+        agent_version,
+    )
+    backend.reset_datapoints()
```
keitwb: Is this necessary?

tests/deployments/puppet/puppet_test.py
```diff
@@ -87,52 +97,57 @@ def run_puppet_agent(cont, backend, monitors, agent_version=None, stage=STAGE):
...
-            installed_version,
-            agent_version,
-        )
-    return installed_version
+    assert installed_version == agent_version, "installed agent version is '%s', expected '%s'" % (
+        installed_version,
+        agent_version,
+    )
+    assert is_agent_running_as_non_root(cont), "Agent is not running as non-root user"
+    backend.reset_datapoints()
```
keitwb: Same thing here.

keitwb: Ok, I see what you are trying to do with it.  Do the reset in the test code instead of here so it doesn't mix concerns so much.

tests/deployments/puppet/puppet_test.py
```diff
@@ -141,87 +156,67 @@ def test_puppet(base_image, init_system):
...
-    run_win_command(cmd, [0, 2])
+def run_win_puppet_agent(backend, monitors, agent_version, stage):
+    with tempfile.TemporaryDirectory() as tmpdir:
+        manifest_path = os.path.join(tmpdir, "agent.pp")
+        config = get_config(backend, monitors, agent_version, stage)
+        print(config)
+        with open(manifest_path, "w+") as fd:
+            fd.write(config)
+        cmd = f"puppet apply {manifest_path}"
+        run_win_command(cmd, [0, 2])
```
keitwb: This is pretty cryptic what `[0, 2]` means.  At least comment it.

tests/deployments/puppet/puppet_test.py
jchengsfx: > Ok, I see what you are trying to do with it. Do the reset in the test code instead of here so it doesn't mix concerns so much.

Will move out of the function and into the test code.

PR #902, opened 2019-08-05
Fix windows pathing issue

kaypeter87:
The escaped '\\' in the init.pp file is not required at the end of $installation_directory. This causes windows to read certain parts of the paths as ```C:\Program Files\SignalFx\\etc...```

Also added commenting on the $config parameter as it would give more clarity on how to set this. For example: Hiera lookup functions.


asuresh4:
Thanks for submitting a PR @kaypeter87! Updates LGTM. Will let you know once the update is released.

PR #904, opened 2019-08-06
add option to internal-metrics to override host dim with agent's host…

charliesignalfx:
… value


charliesignalfx:
Yep, just confirmed that the recent pr implemented this functionality.

PR #321, opened 2018-06-04
Restructured K8S tests

jchengsfx:

PR #905, opened 2019-08-08
Updated PSP for SFX k8s

cpate3:


keitwb:
Hi @cpate3, thanks for submitting this.  This would be good to integrate with the existing Helm chart as a configurable flag in the main chart's values.yaml (e.g. `pspMode: restricted/privileged`) instead of being separated off in another directory.  Can you merge it into the main chart?  It should default to not using PSP mode if no values are provided by the user.

cpate3:
@keitwb merge it into the main chart is done. Also have updated common README so please take a look if that make sense.

deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -23,6 +23,9 @@ rollingUpdateMaxUnavailable: 1
 # Namespace to deploy agent in (Optional: Will default to release namespace)
 namespace:
 
+# Specify PSP (Pod Security Policy) (Optional: default to not using PSP mode if no values are provided by the user)
+pspMode: restricted/privileged
```
keitwb: This value is used as a default so it should be blank to indicate no security mode.

deployments/k8s/helm/signalfx-agent/restricted-sfx-agent.yaml
```diff
@@ -0,0 +1,81 @@
+# This Pod Security Policy is setup to allow the signalfx-agent daemonset to
```
keitwb: These files have to be wrapped in a condition to prevent them from being deployed when the `pspMode` value is not set to `restricted`.

deployments/k8s/helm/signalfx-agent/README.md
```diff
@@ -54,3 +54,71 @@ If you are using OpenShift set `kubernetesDistro` to `openshift` to get
 OpenShift-specific functionality:
 
 `$ helm install --set signalFxAccessToken=<YOUR_ACCESS_TOKEN> --set clusterName=<YOUR_CLUSTER_NAME> --set agentVersion=<VERSION_NUMBER> --set signalFxRealm=<YOUR_SIGNALFX_REALM> signalfx/signalfx-agent --set kubernetesDistro=openshift`
+
+
+### PSP for SignalFx Agent via Helm Chart
+
+# SignalFx Monitoring with Pod Security Policy
```
keitwb: Heading should be deeper than `###` used above it.

Same for all headers below.

deployments/k8s/helm/signalfx-agent/README.md
```diff
@@ -54,3 +54,71 @@ If you are using OpenShift set `kubernetesDistro` to `openshift` to get
...
+- Usage of volume types
+- And a few other aspects including AppArmor, sysctl, and seccomp profiles
+
+## Pre-requisites 
+
+- All Chart dependencies should also be submitted independently
+- Must pass the linter (helm lint)
+- Must successfully launch with default values (helm install .)
+
+## Installation Procedures
```
keitwb: Helm does this for you and is unnecessary to document.

deployments/k8s/helm/signalfx-agent/restricted-sfx-agent.yaml
jaysfx: Looks like all of these need to be moved to templates and templatized.

chiragkumarp: Closed #321 PR and merge to here. Let's finalize and wrap up task if possible. Let me also revise Jay's above comments. Following is PR #321 summary for reference. 

keitwb 29 days ago
Here is an example of conditionalizing the template. This is yaml, so if statements aren't allowed here.

hiragkumarp commented 11 days ago
Thanks Ben for giving hint. Sorry for late update. I also come across this Helm Flow control standard (https://github.com/helm/helm/blob/master/docs/chart_template_guide/control_structures.md) and updated value YAML code accordingly with IF condition. Is this sounds good to approve?

keitwb commented 2 hours ago
@chiragkumarp This PR by itself doesn't really do anything -- can you combine it with #905 and present it as a single PR. Also see the comments on that other PR.

cpate3:
@jaysfx can you take a look now? Please clarify what need to be changed for templates and templatized? I have merged 2 PR in one now so we can fix and then close this finally after successful merge. Let me know what is still missing. The IF condition is applied in value.yaml file.

jaysfx:
@cpate3 thanks for the contribution. We're going to track it internally as part of our other work so we can appropriately prioritize.

emaderer:
Closing this PR since it's been pending for more than 2 years.

PR #906, opened 2019-08-09
Powershell updates

jchengsfx:


jchengsfx:
> Does this mean we were getting false positive results on Windows jobs?

It's possible, but hopefully the integration tests caught any obvious issues.  I did notice when I was manually testing that occasionally the build would pass even though there would be download failures with pip or other stuff, so hopefully these changes will catch them.

PR #907, opened 2019-08-09
Enable discovery of containers without ports

jaysfx:
Adds the kubernetes-scheduler monitor and some supporting changes:

* Add support for user-overridable ports on discovered services
* Allow tests to work with existing local minikube instance
* Add tar to docker image so that files can be copied into containers
  using `kubectl cp`

### DIFF
--- a/internal/observers/kubernetes/api.go
+++ b/internal/observers/kubernetes/api.go
@@ -58,6 +58,15 @@ const (
 // DIMENSION(container_spec_name): The short name of the container in the pod spec,
 // **NOT** the running container's name in the Docker engine
 
+// ENDPOINT_VAR(kubernetes_annotations): The set of annotations on the
+// discovered pod.
+
+// ENDPOINT_VAR(pod_spec): The full pod spec object, as represented by the Go
+// K8s client library (client-go): https://godoc.org/k8s.io/api/core/v1#PodSpec.
+
+// ENDPOINT_VAR(pod_metadata): The full pod metadata object, as represented by the Go
+// K8s client library (client-go): https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta.
+
 func init() {
 	observers.Register(observerType, func(cbs *observers.ServiceCallbacks) interface{} {
 		return &Observer{
@@ -231,13 +240,24 @@ func (o *Observer) endpointsInPod(pod *v1.Pod, client *k8s.Clientset, portAnnota
 		"kubernetes_namespace": pod.Namespace,
 	}
 
+	makeBaseEndpoint := func(idSuffix string, name string) *services.EndpointCore {
+		id := fmt.Sprintf("%s-%s-%s", pod.Name, pod.UID[:7], idSuffix)
+
+		endpoint := services.NewEndpointCore(id, name, observerType, podDims)
+
+		endpoint.Host = podIP
+
+		endpoint.AddExtraField("kubernetes_annotations", pod.Annotations)
+		endpoint.AddExtraField("pod_metadata", &pod.ObjectMeta)
+		endpoint.AddExtraField("pod_spec", &pod.Spec)
+
+		return endpoint
+	}
+
 	for _, container := range pod.Spec.Containers {
 		var containerState string
 		var containerID string
 		var containerName string
-
-		dims := utils.CloneStringMap(podDims)
-		dims["container_spec_name"] = container.Name
 
 		for _, status := range pod.Status.ContainerStatuses {
 			if container.Name != status.Name {
@@ -256,12 +276,21 @@ func (o *Observer) endpointsInPod(pod *v1.Pod, client *k8s.Clientset, portAnnota
 			continue
 		}
 
+		endpointContainer := &services.Container{
+			ID:      containerID,
+			Names:   []string{containerName},
+			Image:   container.Image,
+			Command: "",
+			State:   containerState,
+			Labels:  pod.Labels,
+		}
+
 		for _, port := range container.Ports {
 			portsSeen[port.ContainerPort] = true
 
-			id := fmt.Sprintf("%s-%s-%d", pod.Name, pod.UID[:7], port.ContainerPort)
+			endpoint := makeBaseEndpoint(fmt.Sprintf("%d", port.ContainerPort), port.Name)
 
-			endpoint := services.NewEndpointCore(id, port.Name, observerType, dims)
+			endpoint.AddDimension("container_spec_name", container.Name)
 
 			portAnnotations := annotationConfs.FilterByPortOrPortName(port.ContainerPort, port.Name)
 			monitorType, extraConf, err := configFromAnnotations(container.Name, portAnnotations, pod, client)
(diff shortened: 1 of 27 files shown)


tests/helpers/kubernetes/agent.py
```diff
@@ -105,9 +105,8 @@ def deploy(self, agent_yaml=None):
                 print(f"Created agent configmap:\n{configmap_base}")
 
                 daemonset_base = load_resource_yaml(AGENT_DAEMONSET_PATH)
-                daemonset_base["spec"]["template"]["spec"]["containers"][0]["imagePullPolicy"] = "Always"
                 daemonset_base["spec"]["template"]["spec"]["containers"][0]["image"] = self.agent_image_name
-                daemonset_base["spec"]["template"]["spec"]["containers"][0]["imagePullPolicy"] = "Always"
+                daemonset_base["spec"]["template"]["spec"]["containers"][0]["imagePullPolicy"] = "IfNotPresent"
```
keitwb: Why??

jaysfx: Because when the image that gets created with `make image` (quay.io/signalfx/signalfx-agent-dev) gets pushed to the docker instance having it set to Always will try to pull it from quay.io. Thinking about it more I guess this could break for the case where it's pulling from localhost:5000 repository. Might have to make the value configurable through pytest option? @jchengsfx

internal/monitors/kubernetes/prometheus/kubernetes.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/prometheus/common/expfmt"
+	"github.com/signalfx/signalfx-agent/internal/monitors/prometheusexporter"
+)
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"true"`
+
+	// Host of the exporter
+	Host string `yaml:"host" validate:"required"`
```
keitwb: Can you not embed the Prometheus exporter config struct?

internal/monitors/kubernetes/prometheus/kubernetes.go
```diff
@@ -0,0 +1,85 @@
...
+	}, nil
+}
+
+// GetInterval is a ConfigInterface method implementation for getting the configured monitor run interval.
+func (c *Config) GetInterval() time.Duration {
+	return time.Duration(c.IntervalSeconds) * time.Second
+}
+
+// GetMonitorType is a ConfigInterface method implementation for getting the monitor type.
+func (c *Config) GetMonitorType() string {
```
keitwb: Is this used anywhere?

internal/observers/kubernetes/api.go
```diff
@@ -277,21 +287,29 @@ func (o *Observer) endpointsInPod(pod *v1.Pod, client *k8s.Clientset, portAnnota
...
-				Container:     *container,
+				Container:     *endpointContainer,
+				Orchestration: *orchestration,
+			})
+		}
+
+		if len(container.Ports) == 0 {
+			// No ports were declared. Create an endpoint with no port that can later be user-defined.
+			id := fmt.Sprintf("%s-%s-portless-%d", pod.Name, pod.UID[:7], portlessCount)
+			portlessCount++
```
keitwb: What is the point of this counter?

internal/observers/kubernetes/api.go
```diff
@@ -277,21 +287,29 @@ func (o *Observer) endpointsInPod(pod *v1.Pod, client *k8s.Clientset, portAnnota
...
 				AltPort:       0,
-				Container:     *container,
+				Container:     *endpointContainer,
+				Orchestration: *orchestration,
+			})
+		}
+
+		if len(container.Ports) == 0 {
+			// No ports were declared. Create an endpoint with no port that can later be user-defined.
+			id := fmt.Sprintf("%s-%s-portless-%d", pod.Name, pod.UID[:7], portlessCount)
```
keitwb: Maybe use container name instead of portlessCount?

tests/helpers/kubernetes/agent.py
keitwb: How are you running the integration tests?

tests/monitors/kubernetes_scheduler/kubernetes_scheduler_test.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from tests.helpers.metadata import Metadata
+from tests.helpers.verify import verify
+
+pytestmark = [pytest.mark.kubernetes]
+METADATA = Metadata.from_package("kubernetes/scheduler")
+
+
+@pytest.mark.skip("only works with real minikube currently")
```
keitwb: This can conditionalize the skip based on whether the `--no-use-minikube` option was passed.

tests/helpers/kubernetes/agent.py
jaysfx: `pytest  --no-use-minikube --kube-context minikube --agent-image-name quay.io/signalfx/signalfx-agent-dev:4.7.8-post`

tests/monitors/kubernetes_scheduler/kubernetes_scheduler_test.py
jaysfx: That text isn't quite accurate. The real problem is that the test only works with 1.15 since new metrics were added. Going to add logic to only run for 1.15 or maybe just latest.

internal/monitors/kubernetes/prometheus/kubernetes.go
jaysfx: Yeah in prometheusexporter/monitor.go. It was part of the changes Bremmer did.

internal/monitors/kubernetes/prometheus/kubernetes.go
jaysfx: Prometheus exporter has options like username/password for basic auth that aren't used for Kubernetes APIs (I don't think).

internal/monitors/kubernetes/kubernetes.go
```diff
@@ -5,5 +5,6 @@ import (
 	_ "github.com/signalfx/signalfx-agent/internal/monitors/kubernetes/apiserver"
```
bjsignalfx: Is it possible to rename this file to modules.go?

jaysfx: Yeah I always have a hard time finding this file.

jaysfx:
> I could be wrong but I think we are doing way too much here #TeamTooMuch. It seems to me that we could use the k8s client API better like they did [here](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/metrics/metrics_grabber.go)?

The proxy is a good idea but I think it only makes sense if you were going to access the pod from a different host. In that case you would have to do your own discovery of the pods.

keitwb:
> > I could be wrong but I think we are doing way too much here #TeamTooMuch. It seems to me that we could use the k8s client API better like they did [here](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/metrics/metrics_grabber.go)?
> 
> The proxy is a good idea but I think it only makes sense if you were going to access the pod from a different host. In that case you would have to do your own discovery of the pods.

Accessing the pod from a different host isn't a problem if the agent runs in the same network as the rest of the cluster (our agent incidentally _doesn't_ since we still use `hostNetwork: true`).  We have just gotten lucky that the pod IP network is addressable from the host network almost always.

If we run into problems with addressing the pods directly, we will have to switch to using the proxy API endpoint to access metrics but isn't necessary unless we find problems with the current approach.

internal/observers/kubernetes/api.go
```diff
@@ -277,21 +286,28 @@ func (o *Observer) endpointsInPod(pod *v1.Pod, client *k8s.Clientset, portAnnota
...
 			endpoints = append(endpoints, &services.ContainerEndpoint{
 				EndpointCore:  *endpoint,
 				AltPort:       0,
-				Container:     *container,
+				Container:     *endpointContainer,
+				Orchestration: *orchestration,
+			})
+		}
+
+		if len(container.Ports) == 0 {
```
keitwb: Does this take into account the ports declared via annotations?

I like the concept of an endpoint being a whole container but I'm wary of how it will interact with some of the more indirect ways of defining ports (e.g. K8s annotations or Docker labels).

I kind of think we should _always_ emit a "container/portless" endpoint but I'm not sure how to make it backwards-compatible.

keitwb:
Also, rebase with latest master to get some integration test fixes.

internal/observers/kubernetes/api.go
jaysfx: @keitwb I looked into the annotations stuff more. I guess right now it's built around container port names. I don't see a clean way to extend the annotation naming convention that's there of `config.<port name>`. If we were doing it over maybe:

```
config.port.<name>
config.container.<name>
```

To keep backwards compat. might have to do `containerConfig.<name>` or something?

jaysfx: Or maybe we could look for a port of that name and if it doesn't exist look for a container of that name. 🤷‍♂

PR #911, opened 2019-08-15
Rename DimProperties to Dimension

keitwb:
This reflects what it is much better


keitwb:
Going to also add the new /v2/dimension PATCH functionality coming and will reopen when that is done.

PR #912, opened 2019-08-15
Haproxy: Add note for restrictive socket connection

mstumpfx:
request from the field


docs/monitors/collectd-haproxy.md
```diff
@@ -55,6 +55,19 @@ frontend stats-frontend
     acl ...
 ```
 
+If you are using a more restrictive socket such as the example below,
+you may need to add the signalfx-agent user to the haproxy group.
+This can be achieved by the following command:
+`sudo usermod -a -G haproxy signalfx-agent`
+
+```
+global
```
keitwb: This is kind of confusing how this HAProxy config snippet comes here by itself.  It says above that it is an example of setting up "a more restrictive socket", but it seems like you could omit this config or put a clearer explanation of it.

mstumpfx: re-worded and removed the example config

PR #914, opened 2019-08-16
K8s: Add resource syncing for jobs and cronjobs

mstumpfx:
**Work in progress**
PR to sync jobs and cronjob metrics

Metrics:
- kubernetes.job.completions
- kubernetes.job.parallelism
- kubernetes.job.active
- kubernetes.job.succeeded
- kubernetes.job.failed
- kubernetes.cronjob.active

Added a new job cache so we can link pods to the cronjobs they were created by,
via pods --> jobs --> cronjob link

To do:
- [x] add tests for cron jobs
- [x] add link for pod --> job --> cronjob to generate property
- [x] add test for cronjob property existing on kubernetes_pod_uid
- [x] add perf test for jobs (~5k jobs)
- [x] fix tests

### DIFF
--- a/internal/monitors/kubernetes/cluster/metrics/jobs.go
+++ b/internal/monitors/kubernetes/cluster/metrics/jobs.go
@@ -1,0 +1,78 @@
+package metrics
+
+import (
+	"time"
+
+	"github.com/signalfx/golib/datapoint"
+	k8sutil "github.com/signalfx/signalfx-agent/internal/monitors/kubernetes/utils"
+	atypes "github.com/signalfx/signalfx-agent/internal/monitors/types"
+	"github.com/signalfx/signalfx-agent/internal/utils"
+	batchv1 "k8s.io/api/batch/v1"
+)
+
+func datapointsForJob(job *batchv1.Job) []*datapoint.Datapoint {
+	dimensions := map[string]string{
+		"metric_source":        "kubernetes",
+		"kubernetes_namespace": job.Namespace,
+		"kubernetes_uid":       string(job.UID),
+		"kubernetes_name":      job.Name,
+	}
+
+	return []*datapoint.Datapoint{
+		datapoint.New(
+			"kubernetes.job.completions",
+			dimensions,
+			datapoint.NewIntValue(int64(*job.Spec.Completions)),
+			datapoint.Gauge,
+			time.Now()),
+		datapoint.New(
+			"kubernetes.job.parallelism",
+			dimensions,
+			datapoint.NewIntValue(int64(*job.Spec.Parallelism)),
+			datapoint.Gauge,
+			time.Now()),
+		datapoint.New(
+			"kubernetes.job.active",
+			dimensions,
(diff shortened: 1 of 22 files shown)


tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
```diff
@@ -96,3 +96,39 @@ def test_kubernetes_cluster_namespace_scope(k8s_cluster):
             assert ensure_always(
                 lambda: not has_datapoint(agent.fake_services, dimensions={"kubernetes_namespace": "bad"})
             ), "got pod metrics from unspecified namespace"
+
+
+@pytest.mark.kubernetes
+def test_jobs(k8s_cluster):
```
keitwb: See if you can add a CronJob test, maybe using a very short interval so the test doens't take too long to run.

mstumpfx: Added a test, shortest interval available was 1m

mstumpfx: but the job should kickoff as soon as it's created, and the test should not need 1 more than 1 job run so it shouldn't even take that full minute.

internal/monitors/kubernetes/cluster/metrics/jobs.go
```diff
@@ -0,0 +1,79 @@
...
+
+func dimPropsForJob(job *batchv1.Job) *atypes.DimProperties {
+	props, tags := k8sutil.PropsAndTagsFromLabels(job.Labels)
+
+	props["kubernetes_name"] = job.Name
+	props["k8s_workload"] = "Job"
+
+	for _, or := range job.OwnerReferences {
+		props[utils.LowercaseFirstChar(or.Kind)] = or.Name
+		props["k8s_"+utils.LowercaseFirstChar(or.Kind)+"_uid"] = string(or.UID)
```
mstumpfx:
```suggestion
		props["owner_"+utils.LowercaseFirstChar(or.Kind)+"_uid"] = string(or.UID)
```

internal/monitors/kubernetes/cluster/metrics/jobs.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/signalfx/signalfx-agent/internal/utils"
+	batchv1 "k8s.io/api/batch/v1"
+)
+
+func datapointsForJob(job *batchv1.Job) []*datapoint.Datapoint {
+	dimensions := map[string]string{
+		"metric_source":        "kubernetes",
+		"kubernetes_namespace": job.Namespace,
+		"kubernetes_uid":       string(job.UID),
+		"k8s_job_uid":          string(job.UID),
```
mstumpfx:
```suggestion
		"owner_job_uid":          string(job.UID),
```

internal/monitors/kubernetes/cluster/metrics/jobs.go
mstumpfx: actually, maybe not.

PR #916, opened 2019-08-19
Don't fail on first Prometheus exporter error

keitwb:


keitwb:
This doesn't really work since the prometheus parser parses everything at once and fails without emitting anything.

PR #917, opened 2019-08-19
Use microk8s for k8s tests in circleci

jchengsfx:


.circleci/config.yml
```diff
@@ -126,45 +126,27 @@ commands:
...
-            - /home/circleci/bin/kubectl
-
   install_helm:
     parameters:
       version:
         type: string
         default: v2.13.0
     steps:
       - restore_cache:
-          key: "v1-helm-binary-<< parameters.version >>"
```
keitwb: If this doesn't have to do with microk8s, move it to another PR.

PR #919, opened 2019-08-20
Adding smart agent version as a tag to spans

krunaljain:
Issue: https://signalfuse.atlassian.net/browse/APMB-242

### DIFF
--- a/internal/core/common/constants/constants.go
+++ b/internal/core/common/constants/constants.go
@@ -7,6 +7,6 @@ const (
 	AgentVersionEnvVar = "SIGNALFX_AGENT_VERSION"
 	// BundleDirEnvVar is a path to the root of the collectd/python bundle
 	BundleDirEnvVar = "SIGNALFX_BUNDLE_DIR"
-	// Tag key for signalfx smartagent version
+	// SmartAgentVersionTagKey is the tag key for signalfx smartagent version
 	SmartAgentVersionTagKey = "signalfx.smartagent.version"
 )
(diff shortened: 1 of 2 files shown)


internal/core/common/constants/constants.go
```diff
@@ -7,4 +7,6 @@ const (
 	AgentVersionEnvVar = "SIGNALFX_AGENT_VERSION"
 	// BundleDirEnvVar is a path to the root of the collectd/python bundle
 	BundleDirEnvVar = "SIGNALFX_BUNDLE_DIR"
+	// SmartAgentVersionTagKey is the tag key for signalfx smartagent version
+	SmartAgentVersionTagKey = "signalfx.smartagent.version"
```
keitwb: Don't really need a constant if it's only used once.

internal/core/writer/spans.go
```diff
@@ -151,6 +153,8 @@ func (sw *SignalFxWriter) preprocessSpan(span *trace.Span) {
 		delete(span.Tags, dpmeta.NotHostSpecificMeta)
 	}
 
+	// adding smart agent version as a tag
+	span.Tags[constants.SmartAgentVersionTagKey] = os.Getenv(constants.AgentVersionEnvVar)
```
keitwb: There is a generated module in the constants package you are importing here called `version.go` that contains a string const `Version` that is much cleaner than going through this envvar.   That module isn't committed to git so that's probably why you didn't see it. The envvar with the version only exists to make it possible to pass it down to the collectd subprocess.

PR #923, opened 2019-08-22
Update default  kubelet port to 10255

asuresh4:
Issue: https://signalfuse.atlassian.net/browse/INT-1510

### DIFF
--- a/docs/kubernetes-setup.md
+++ b/docs/kubernetes-setup.md
@@ -137,7 +137,7 @@ If you see errors like this in the agent logs:
 
 `Couldn't get machine info: Kubelet request failed - "401 Unauthorized", response: "Unauthorized"`
 
-`Couldn't get cAdvisor container stats" error="failed to get all container stats from Kubelet URL "https://localhost:10250/stats/container/": Kubelet request failed - "401 Unauthorized", response: "Unauthorized"`
+`Couldn't get cAdvisor container stats" error="failed to get all container stats from Kubelet URL "https://localhost:10255/stats/container/": Kubelet request failed - "401 Unauthorized", response: "Unauthorized"`
 
 This means that the agent cannot authenticate to the kubelet.  Assuming you
 have the ClusterRole and ClusterRoleBinding properly applied to the agent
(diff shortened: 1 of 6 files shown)


keitwb:
@asuresh4 A bunch of the special provider-specific instructions in kubernetes-setup.md are no longer necessary now with the new default (e.g. EKS).

docs/kubernetes-setup.md
```diff
@@ -137,7 +137,7 @@ If you see errors like this in the agent logs:
 
 `Couldn't get machine info: Kubelet request failed - "401 Unauthorized", response: "Unauthorized"`
 
-`Couldn't get cAdvisor container stats" error="failed to get all container stats from Kubelet URL "https://localhost:10250/stats/container/": Kubelet request failed - "401 Unauthorized", response: "Unauthorized"`
+`Couldn't get cAdvisor container stats" error="failed to get all container stats from Kubelet URL "https://localhost:10255/stats/container/": Kubelet request failed - "401 Unauthorized", response: "Unauthorized"`
```
keitwb: Leave this as it was, people might still be using the old config and will need this as it was.

docs/monitors/kubelet-stats.md
```diff
@@ -38,7 +38,7 @@ The **nested** `kubeletAPI` config object has the following fields:
 
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
-| `url` | no | `string` | URL of the Kubelet instance.  This will default to `https://<current node hostname>:10250` if not provided. |
+| `url` | no | `string` | URL of the Kubelet instance.  This will default to `https://<current node hostname>:10255` if not provided. |
```
keitwb: Typically port 10255 is HTTP, not HTTPS.

internal/core/common/kubelet/client.go
```diff
@@ -74,7 +74,7 @@ func NewClient(kubeletAPI *APIConfig) (*Client, error) {
 		if err != nil {
 			return nil, err
 		}
-		kubeletAPI.URL = fmt.Sprintf("https://%s:10250", hostname)
+		kubeletAPI.URL = fmt.Sprintf("https://%s:10255", hostname)
```
keitwb: `http`.  We should confirm on various platforms.  @asuresh4 can you check on Azure (AKS)?  I will confirm GKE and perhaps @mstumpfx can confirm on EKS since he was working with that recently.

docs/monitors/kubelet-stats.md
asuresh4: updated

PR #926, opened 2019-08-29
Add k8s_workload property to other workloads

mstumpfx:
For consistency with other workloads.

- Added dimProps for DaemonSets
- Added `k8s_workload` and `kubernetes_name` for Deployments (double publishing k8s name)
- Added `k8s_workload` and `kubernetes_name` for ReplicaSets (double publishing k8s name)
- Added `k8s_workload` for Pods, to be set to the OwnersReferences Kind

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,8 +20,7 @@ issues:
       linters:
         - gosec
 
-    - path: internal/monitors/kubernetes/cluster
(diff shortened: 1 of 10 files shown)


keitwb:
If you rebase with master the tests should pass now with the git.apache.org workaround.

PR #929, opened 2019-08-30
Incorporate kubernetes auth support into prometheus-exporter

jaysfx:
prometheus-exporter needs k8s auth support because we may want to monitor
something that's behind kube-rbac-proxy or some other k8s service we don't have
a wrapper monitor for.


jaysfx:
Note that this has kubernetes-scheduler from #907 as there's more work to be done with the port override thing and need to get the kubernetes auth asap.

internal/monitors/prometheusexporter/prometheus.go
```diff
@@ -0,0 +1,221 @@
...
+	var ctx context.Context
+	ctx, m.cancel = context.WithCancel(context.Background())
+	utils.RunOnInterval(ctx, func() {
+		dps, err := fetchPrometheusMetrics(fetch)
+		if err != nil {
+			m.logger.WithError(err).Error("Could not get prometheus metrics")
+			return
+		}
+
+		now := time.Now()
```
keitwb: Don't need to set timestamp.

PR #935, opened 2019-09-04
Update test_kubernetes_cluster_in_k8s to get StatefulSet metrics

jchengsfx:
Need to deploy the StatefulSet for the test since the kubernetes_cluster monitor now includes kubernetes.stateful_set.* metrics.

### DIFF
--- a/tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
+++ b/tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
@@ -21,6 +21,7 @@ def test_kubernetes_cluster_in_k8s(k8s_cluster):
         SCRIPT_DIR / "resource_quota.yaml",
         TEST_SERVICES_DIR / "nginx/nginx-k8s.yaml",
         SCRIPT_DIR / "cronjob.yaml",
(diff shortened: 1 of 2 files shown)


tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
```diff
@@ -21,6 +21,7 @@ def test_kubernetes_cluster_in_k8s(k8s_cluster):
         SCRIPT_DIR / "resource_quota.yaml",
         TEST_SERVICES_DIR / "nginx/nginx-k8s.yaml",
         SCRIPT_DIR / "cronjob.yaml",
+        SCRIPT_DIR / "statefulset.yaml",
```
keitwb: How did this pass before without this?

jchengsfx: > How did this pass before without this?

I think we were getting false positives in circleci if the `kubernetes_cluster_test:test_stateful_sets` was running at the same time.  Running the test by itself locally has been failing for me consistently.

keitwb: Ok, makes sense.

PR #937, opened 2019-09-04
Increase logstash test timeouts

jchengsfx:
Try to reduce test timeouts as seen in https://circleci.com/gh/signalfx/signalfx-agent/166523#tests/containers/3 and https://circleci.com/gh/signalfx/signalfx-agent/166788#tests/containers/2.

### DIFF
--- a/tests/monitors/logstash_tcp/logstash_tcp_test.py
+++ b/tests/monitors/logstash_tcp/logstash_tcp_test.py
@@ -58,7 +58,7 @@ def test_logstash_tcp_client(version):
         )
 
         with Agent.run(config) as agent:
-            assert wait_for(p(tcp_socket_open, host, 8900), timeout_seconds=120), "logstash didn't start"
+            assert wait_for(p(tcp_socket_open, host, 8900), timeout_seconds=180), "logstash didn't start"
             assert wait_for(p(has_datapoint, agent.fake_services, "logins.count", value=7, dimensions={}))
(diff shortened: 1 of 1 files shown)


tests/monitors/logstash_tcp/logstash_tcp_test.py
```diff
@@ -41,11 +41,13 @@ def test_logstash_tcp_client(version):
         f"docker.elastic.co/logstash/logstash:{version}",
         environment={"XPACK_MONITORING_ENABLED": "false", "CONFIG_RELOAD_AUTOMATIC": "true"},
     ) as logstash_cont:
+        host = container_ip(logstash_cont)
+        assert wait_for(p(tcp_socket_open, host, 5044), timeout_seconds=180), "logstash didn't start"
```
keitwb: I was intentionally not waiting for it to start because I wanted to make sure the monitor could handle conditions where the service wasn't up when the agent started (e.g. that the monitor waits for it to be up instead of just bombing out and never doing anything even after it comes up).

Can you try increasing the timeout of the first assert below to 180 seconds from 120 and see if that helps.  If it still fails there it might imply a bug with the monitor where it won't connect properly.

jchengsfx: @keitwb looks like the increasing the original assertion timeouts helped.

keitwb: Ok, logstash must just be really slow to start up then.

PR #941, opened 2019-09-06
Replace kubernetes_node with kubernetes_node_uid

keitwb:
This changes the global dimension and requires looking up the node UID
from the K8s API server.  The lookup is triggered if the envvar MY_NODE_NAME
is present in the agent.  The agent will refuse to start up if the node uid
cannot be resolved and MY_NODE_NAME is present.

Also removes old machineID and node name config options, as they are all replaced
with node uid.


asuresh4:
@keitwb I believe the metadata.yaml file also needs to be updated right? with the new dimension name

keitwb:
> @keitwb I believe the metadata.yaml file also needs to be updated right? with the new dimension name

Done.

PR #947, opened 2019-09-09
Add logstash monitor

seonsfx:


docs/monitors/logstash.md
```diff
@@ -0,0 +1,208 @@
...
+Configuration](../monitor-config.md#common-configuration).**
+
+
+| Config option | Required | Type | Description |
+| --- | --- | --- | --- |
+| `host` | no | `string` | The hostname of Logstash monitoring API (**default:** `127.0.0.1`) |
+| `port` | no | `integer` | The port number of Logstash monitoring API (**default:** `9600`) |
+| `useHTTPS` | no | `bool` | If true, the agent will connect to the host using HTTPS instead of plain HTTP. (**default:** `false`) |
+| `timeoutSeconds` | no | `integer` | The maximum amount of time to wait for API requests (**default:** `5`) |
+| `enableExtraHotThreadsMetrics` | no | `bool` | Whether it will send all extra hot threads metrics as well. (**default:** `false`) |
```
keitwb: Get rid of all of these.  This is what the new group functionality supersedes.

internal/monitors/logstash/logstash/monitor.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	return nil
+}
+
+func (m *Monitor) extractDatapoints(metricPath string, metricsJSON map[string]interface{}, dims map[string]string) []*datapoint.Datapoint {
+	var dps []*datapoint.Datapoint
+
+	for k, v := range metricsJSON {
+		childPath := fmt.Sprintf("%s.%s", metricPath, k)
```
keitwb: Probably going to be faster to do `metricPath + "." + k` since you don't need any of the conversion capabilities of Sprintf.

internal/monitors/logstash/logstash/monitor.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+	}
+
+	return dps
+}
+
+func (m *Monitor) fetchPluginMetrics(client *http.Client, endpoint string, prefix string, dimensions map[string]string) []*datapoint.Datapoint {
+	if metricsJSON, err := getJSON(client, endpoint); err == nil {
+		if total, exists := metricsJSON["total"]; exists {
+			metricName := fmt.Sprintf("%s.%s", prefix, "total")
```
keitwb: `Same thing about Sprintf.`

internal/monitors/logstash/logstash/monitor.go
```diff
@@ -0,0 +1,232 @@
...
+func (m *Monitor) fetchPipelineMetrics(client *http.Client, endpoint string, prefix string, dimensions map[string]string) []*datapoint.Datapoint {
+	var dps []*datapoint.Datapoint
+
+	if metricsJSON, err := getJSON(client, endpoint); err == nil {
+		if pipelines, exists := metricsJSON["pipelines"]; exists {
+			if pipelinesObj, isObject := pipelines.(map[string]interface{}); isObject {
+				for pipelineName, pipelineObj := range pipelinesObj {
+					if pipeline, converted := pipelineObj.(map[string]interface{}); converted {
+						dimsClone := make(map[string]string)
+						for dimk, dimv := range dimensions {
```
keitwb: I would invert the logic leading down to here and flatten these six levels of nesting down to one or two.  You can do that by returning on the "bad" paths.

internal/monitors/logstash/logstash/config.go
```diff
@@ -0,0 +1,82 @@
...
+	// The hostname of Logstash monitoring API
+	Host string `yaml:"host" default:"127.0.0.1"`
+	// The port number of Logstash monitoring API
+	Port uint16 `yaml:"port" default:"9600"`
+	// If true, the agent will connect to the host using HTTPS instead of plain HTTP.
+	UseHTTPS bool `yaml:"useHTTPS"`
+	// The maximum amount of time to wait for API requests
+	TimeoutSeconds int `yaml:"timeoutSeconds" default:"5"`
+	// Whether it will send all extra hot threads metrics as well.
+	EnableExtraHotThreadsMetrics bool `yaml:"enableExtraHotThreadsMetrics"`
```
keitwb: The groups functionality covers all of these boolean flags and they are unnecessary.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      default: true
+      type: gauge
+      group: pipeline
+    node.stats.pipelines.plugins.outputs.events.in:
+      description: Number of incoming events to output plugins
+      default: true
+      type: cumulative
+      group: pipeline
+    node.stats.pipelines.plugins.outputs.events.out:
+      description: Number of outgoing events from output plugins
```
keitwb: This description is somewhat misleading.  The way it is phrased it sounds like a gauge ("outgoing" is present participle, and gauges represent present state), not a cumulative counter.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      default: true
+      type: cumulative
+      group: pipeline
+    node.stats.pipelines.queue.events_count:
+      description: Number of events in queue
```
keitwb: Same thing.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      default: true
+      type: cumulative
+      group: events
+    node.stats.events.events.out:
+      description: Number of outgoing events
+      default: true
+      type: cumulative
+      group: events
+    node.stats.events.events.queue_push_duration_in_millis:
+      description: Queue push duration of events
```
keitwb: I cannot tell what this means or whether it is really supposed to be a gauge or not from this description.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      default: false
+      type: gauge
+      group: events
+    node.plugins.total:
+      description: Number of plugins
```
keitwb: This one seems a lot more like a gauge than a cumulative.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      group: process
+    node.stats.process.process.mem.total_virtual_in_bytes:
+      description: Size of the virtual memory of this process
+      default: false
+      type: gauge
+      group: process
+    node.stats.process.process.open_file_descriptors:
+      description: Number of currently open file descriptors
+      default: false
+      type: cumulative
```
keitwb: The description says "currently" which implies this should be a gauge.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      group: process
+    node.stats.process.process.peak_open_file_descriptors:
+      description: Peak number of currently open file descriptors
+      default: false
+      type: cumulative
```
keitwb: Same thing.  Being a "peak" value implies it might be monotonically increasing but it still isn't really the semantics of a cumulative counter.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      default: true
+      type: gauge
+      group: events
+    node.stats.events.events.filtered:
+      description: Number of filtered events
+      default: true
+      type: cumulative
+      group: events
+    node.stats.events.events.in:
+      description: Number of incoming events
```
keitwb: Rephrase to make it clear that this means the total number of events input to the system, not currently considered "incoming".

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      group: pipeline
+    node.stats.pipelines.reloads.successes:
+      description: Number of successful reloads
+      default: false
+      type: cumulative
+      group: pipeline
+    node.stats.events.events.duration_in_millis:
+      description: Duration of events
+      default: true
+      type: gauge
```
keitwb: Are you sure this isn't a cumulative counter?

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      group: jvm
+    node.stats.jvm.jvm.threads.count:
+      description: Number of JVM threads
+      default: true
+      type: cumulative
+      group: jvm
+    node.stats.jvm.jvm.threads.peak_count:
+      description: Highest number of JVM threads
+      default: true
+      type: cumulative
```
keitwb: "Highest number" is still a gauge, not cumulative counter.

internal/monitors/logstash/logstash/metadata.yaml
```diff
@@ -0,0 +1,427 @@
...
+      group: jvm
+    node.stats.jvm.jvm.uptime_in_millis:
+      description: Uptime length of JVM
+      default: false
+      type: gauge
+      group: jvm
+    node.os.os.available_processors:
+      description: Number of available processors
+      default: false
+      type: cumulative
```
keitwb: This is definitely more gauge-like.

PR #948, opened 2019-09-10
allow configuring the dbname

mteiste:
Not all services have "postgres" database available and
adding dbname= to connectionString had no effect as it
would be "overridden" by hardcoded dbname=postgres.

### DIFF
--- a/docs/monitors/postgresql.md
+++ b/docs/monitors/postgresql.md
@@ -88,6 +88,7 @@ Configuration](../monitor-config.md#common-configuration).**
 | --- | --- | --- | --- |
 | `host` | no | `string` |  |
 | `port` | no | `integer` |  (**default:** `0`) |
+| `masterDBName` | no | `string` | The "master" database to which the agent first connects to query the list of databases available in the server.  This database should be accessible to the user specified with `connectionString` and `params` below, and that user should have permission to query `pg_database`.  If you want to filter which databases are monitored, use the `databases` option below. (**default:** `postgres`) |
 | `connectionString` | no | `string` | See https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters. |
 | `params` | no | `map of strings` | Parameters to the connection string that can be templated into the connection string with the syntax `{{.key}}`. |
 | `databases` | no | `list of strings` | List of databases to send database-specific metrics about.  If omitted, metrics about all databases will be sent.  This is an [overridable set](https://docs.signalfx.com/en/latest/integrations/agent/filtering.html#overridable-filters). (**default:** `[*]`) |
--- a/internal/monitors/postgresql/monitor.go
+++ b/internal/monitors/postgresql/monitor.go
@@ -29,6 +29,13 @@ type Config struct {
 
 	Host string `yaml:"host"`
 	Port uint16 `yaml:"port"`
+	// The "master" database to which the agent first connects to query the
+	// list of databases available in the server.  This database should be
+	// accessible to the user specified with `connectionString` and `params`
+	// below, and that user should have permission to query `pg_database`.  If
+	// you want to filter which databases are monitored, use the `databases`
+	// option below.
+	MasterDBName string `yaml:"masterDBName" default:"postgres"`
 
 	// See
 	// https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters.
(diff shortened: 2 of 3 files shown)


keitwb:
Hi @mteiste, thanks for this.  Looks mostly good.  I tried to push an update to add a bit of documentation to the PR, but I got permission denied.  Can you add this patch?

```
diff --git a/docs/monitors/postgresql.md b/docs/monitors/postgresql.md
index c91bc352..99f1e2e1 100644
--- a/docs/monitors/postgresql.md
+++ b/docs/monitors/postgresql.md
@@ -88,6 +88,7 @@ Configuration](../monitor-config.md#common-configuration).**
 | --- | --- | --- | --- |
 | `host` | no | `string` |  |
 | `port` | no | `integer` |  (**default:** `0`) |
+| `masterDBName` | no | `string` | The "master" database to which the agent first connects to query the list of databases available in the server.  This database should be accessible to the user specified with `connectionString` and `params` below, and that user should have permission to query `pg_database`.  If you want to filter which databases are monitored, use the `databases` option below. (**default:** `postgres`) |
 | `connectionString` | no | `string` | See https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters. |
 | `params` | no | `map of strings` | Parameters to the connection string that can be templated into the connection string with the syntax `{{.key}}`. |
 | `databases` | no | `list of strings` | List of databases to send database-specific metrics about.  If omitted, metrics about all databases will be sent.  This is an [overridable set](https://docs.signalfx.com/en/latest/integrations/agent/filtering.html#overridable-filters). (**default:** `[*]`) |
diff --git a/internal/monitors/postgresql/monitor.go b/internal/monitors/postgresql/monitor.go
index da6899b3..cbae8809 100644
--- a/internal/monitors/postgresql/monitor.go
+++ b/internal/monitors/postgresql/monitor.go
@@ -27,9 +27,15 @@ func init() {
 type Config struct {
 	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"true"`
 
-	Host   string `yaml:"host"`
-	Port   uint16 `yaml:"port"`
-	DBName string `yaml:"dbname" default:"postgres"`
+	Host string `yaml:"host"`
+	Port uint16 `yaml:"port"`
+	// The "master" database to which the agent first connects to query the
+	// list of databases available in the server.  This database should be
+	// accessible to the user specified with `connectionString` and `params`
+	// below, and that user should have permission to query `pg_database`.  If
+	// you want to filter which databases are monitored, use the `databases`
+	// option below.
+	MasterDBName string `yaml:"masterDBName" default:"postgres"`
 
 	// See
 	// https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters.
@@ -83,7 +89,7 @@ func (m *Monitor) Configure(conf *Config) error {
 		return fmt.Errorf("could not render connectionString template: %v", err)
 	}
 
-	m.database, err = dbsql.Open("postgres", connStr+" dbname="+m.conf.DBName)
+	m.database, err = dbsql.Open("postgres", connStr+" dbname="+m.conf.MasterDBName)
 	if err != nil {
 		return err
 	}
@@ -212,7 +218,7 @@ func (m *Monitor) monitorServer() (*sql.Monitor, error) {
 
 	return sqlMon, sqlMon.Configure(&sql.Config{
 		MonitorConfig:    m.conf.MonitorConfig,
-		ConnectionString: connStr + " dbname=" + m.conf.DBName,
+		ConnectionString: connStr + " dbname=" + m.conf.MasterDBName,
 		DBDriver:         "postgres",
 		Queries:          defaultServerQueries,
 	})
diff --git a/selfdescribe.json b/selfdescribe.json
index 04479cc1..4b60ce8e 100644
--- a/selfdescribe.json
+++ b/selfdescribe.json
@@ -28780,6 +28780,14 @@
             "type": "uint16",
             "elementKind": ""
           },
+          {
+            "yamlName": "masterDBName",
+            "doc": "The \"master\" database to which the agent first connects to query the list of databases available in the server.  This database should be accessible to the user specified with `connectionString` and `params` below, and that user should have permission to query `pg_database`.  If you want to filter which databases are monitored, use the `databases` option below.",
+            "default": "postgres",
+            "required": false,
+            "type": "string",
+            "elementKind": ""
+          },
           {
             "yamlName": "connectionString",
             "doc": "See https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters.",
```

Let me know if the changes make sense to you.

mteiste:
patch applied

keitwb:
Ok thanks, will merge and release later today (or maybe tomorrow).

PR #950, opened 2019-09-10
Use US mirror for ubuntu images

jchengsfx:


jchengsfx:
> Does this see a significant speed increase?

The tests do seem to run faster, but will need to investigate more. Won't know about the agent image builds until these changes are pushed to the stage caches.

keitwb:
@jchengsfx Have you decided whether to commit to this or not?

jchengsfx:
> @jchengsfx Have you decided whether to commit to this or not?

I will hold off for now @keitwb.  It's been difficult to compare since the main mirror has not had any issues recently, i.e. both mirrors show similar performance and will need to wait for the main mirror to start having issues again to test.  Also, I don't want to negatively impact any developers outside of the US.

PR #957, opened 2019-09-13
Collect container status and reason for termiation/waiting

asuresh4:
Issue: https://signalfuse.atlassian.net/browse/INT-1537

### DIFF
--- a/internal/monitors/kubernetes/cluster/metrics/containers.go
+++ b/internal/monitors/kubernetes/cluster/metrics/containers.go
@@ -7,43 +7,32 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func dimPropsForContainer(pod *v1.Pod) []*atypes.DimProperties {
-	dimPropsList := make([]*atypes.DimProperties, 0)
+func dimPropsForContainer(cs v1.ContainerStatus) *atypes.DimProperties {
 
-	for _, cs := range pod.Status.ContainerStatuses {
+	containerProps := make(map[string]string)
 
-		if cs.ContainerID == "" {
-			continue
-		}
-
-		containerProps := make(map[string]string)
-
-		if cs.State.Running != nil {
-			containerProps["container_status"] = "running"
-		}
-
-		if cs.State.Terminated != nil {
-			containerProps["container_status"] = "terminated"
-			containerProps["container_status_reason"] = cs.State.Terminated.Reason
-		}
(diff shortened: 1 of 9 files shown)


internal/monitors/kubernetes/cluster/metrics/containers.go
```diff
@@ -0,0 +1,49 @@
...
+		if cs.State.Waiting != nil {
+			containerProps["container_status"] = "waiting"
+			containerProps["container_status_reason"] = cs.State.Waiting.Reason
+		}
+
+		if len(containerProps) > 0 {
+			dimProps := &atypes.DimProperties{
+				Dimension: atypes.Dimension{
+					Name:  "container_id",
+					Value: strings.Replace(cs.ContainerID, "docker://", "", 1),
```
keitwb: Pull this logic out to a function so it isn't duplicated between `pods.go` and here.

keitwb: Also please confirm that container_id is fully matching between what kubelet-stats and this derives.

tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
```diff
@@ -205,3 +205,17 @@ def test_cronjobs(k8s_cluster):
...
+def test_container_properties(k8s_cluster):
+    config = """
+    monitors:
+     - type: kubernetes-cluster
+    """
+    yamls = [SCRIPT_DIR / "resource_quota.yaml", TEST_SERVICES_DIR / "nginx/nginx-k8s.yaml"]
+    with k8s_cluster.create_resources(yamls):
+        with k8s_cluster.run_agent(agent_yaml=config) as agent:
+            assert wait_for(
+                p(any_dim_val_has_prop, agent.fake_services, dim_name="container_id", prop_name="container_status")
```
keitwb: Would be nice to actually test the values instead of just checking for the key.

internal/monitors/kubernetes/cluster/metrics/containers.go
asuresh4: Yes, confirmed the properties are getting propagated to MTSes from kubelet-stats monitor and addressed your comment.

asuresh4:
@keitwb PTAL when you get a chance. I believe I've addressed comments we discussed.

PR #958, opened 2019-09-13
Make kubernetes-apiserver metrics all custom

jaysfx:
We haven't decided yet which are bundled or not.


jaysfx:
> Doesn't this mean that the monitor emits no metrics at all when run with no special config? That seems very confusing to have a monitor that does nothing.

Yeah. I'm not sure what a solution would look like until we offer builtin content. I can add a note to the documentation?

PR #963, opened 2019-09-16
Fix unique id dimension for daemon sets

asuresh4:
`dimPropsForDaemonSet` puts properties on `kubernetes_uid`. However, the datapoints do not have that dimension. It's got `uid`.

@keitwb - making a PR against your branch since this is also a breaking change and it might make sense to release along with changes from #941


keitwb:
This PR was not based on the right branch.  It removes changes from the k8s_node_uid branch.

asuresh4:
@keitwb should be fixed now

PR #967, opened 2019-09-19
Introduce Hiera data structure and default agent config for convenience

kaypeter87:
It would be nice to have a default hiera data structure and a default agent config for convenience. In the future, I can contribute more on making the puppet code more modular (OS specific yaml files etc...).

For some, using reverse ip lookups when the key `hostname` is empty can sometimes be an issue. It would be better to set the FQDN of the machine itself instead of DNS. The default.yaml uses puppet facts to set the hostname, which is better for consistency.


keitwb:
Thanks for this @kaypeter87.  You probably know more about Puppet than I do so please let us know any other issues you might see in our Puppet module.

PR #971, opened 2019-09-20
Set agent version in daemonset during release

jchengsfx:


jaysfx:
Which file wasn't having the version bumped?

jchengsfx:
> Which file wasn't having the version bumped?

`deployments/k8s/daemonset.yaml` was not getting updated with `<version>` when running `./scripts/update-deployment-versions <version>`.

jaysfx:
I'm confused, there seems to be a bunch of commits updating that file for each release we've done.

PR #972, opened 2019-09-20
Add script to export all metrics in csv

seonsfx:
Thanks to @jadeblaquiere

### DIFF
--- a/scripts/dump-metrics-csv.py
+++ b/scripts/dump-metrics-csv.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+import csv
+import glob
+import pathlib
+import yaml
+
+all_metrics = ""
+
+path = pathlib.Path.cwd() / 'all_metrics.csv'
+with path.open(mode='w') as csvfile:
+    fieldnames = ['monitor', 'metric', 'description', 'default', 'type']
+    writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
+    writer.writeheader()
+    for filename in glob.glob("internal/monitors/**/metadata.yaml", recursive=True):
+        with open(filename) as f:
+            data = yaml.safe_load(f)
+            mons = data['monitors']
+            if mons is None:
+                continue
+            for m in mons:
+                if 'metrics' not in m.keys():
+                    continue
+                metrics = m['metrics']
+                if metrics is None:
+                    continue
+                for mm in metrics:
+                    if 'description' not in metrics[mm].keys():
+                        desc = "THIS METRIC HAS NO DESCRIPTION"
+                    else:
+                        desc = str(metrics[mm]['description']).replace("\n", " ")
+                    if desc is None:
+                        desc = "THE METRIC DESCRIPTION IS BLANK"
+                    if 'default' not in metrics[mm].keys():
+                        default = False
+                    else:
+                        default = metrics[mm]['default']
+                    if default is None:
+                        default = "DEFAULT IS BLANK"
+                    if 'type' not in metrics[mm].keys():
+                        mtype = "THIS METRICS HAS NO TYPE"
+                    else:
+                        mtype = metrics[mm]['type']
+                    if mtype is None:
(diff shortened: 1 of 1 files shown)


scripts/dump-metrics-csv.py
```diff
@@ -0,0 +1,49 @@
...
+
+import codecs
+import fnmatch
+import os
+import sys
+import yaml
+
+all_metrics = ""
+
+for root, dirs, files in os.walk("."):
```
jaysfx: `glob.glob("internal/monitors/**/metadata.yaml", recursive=True)`

scripts/dump-metrics-csv.py
```diff
@@ -0,0 +1,49 @@
...
+                            default = metrics[mm]['default']
+                        if default is None:
+                            default = "DEFAULT IS BLANK"
+                        if 'type' not in metrics[mm].keys():
+                            mtype = "THIS METRICS HAS NO TYPE"
+                        else:
+                            mtype = metrics[mm]['type']
+                        if mtype is None:
+                            mtype = "TYPE IS BLANK"
+                        all_metrics += ('"' + m['monitorType'] + '", "' + mm + '", "' + desc + '", "' + str(default) + '", "' + mtype + '"\n')
```
jaysfx: How is escaping of `,` handled in description? Should probably just use `csv` lib.

scripts/dump-metrics-csv.py
```diff
@@ -0,0 +1,49 @@
...
+output = codecs.open("all_metrics.csv", "w", "utf-8")
+output.write(all_metrics)
```
jaysfx: `Path.write_text`

scripts/dump-metrics-csv.py
keitwb: Same issue for `"`.

scripts/dump-metrics-csv.py
```diff
@@ -0,0 +1,49 @@
...
+                        if default is None:
+                            default = "DEFAULT IS BLANK"
+                        if 'type' not in metrics[mm].keys():
+                            mtype = "THIS METRICS HAS NO TYPE"
+                        else:
+                            mtype = metrics[mm]['type']
+                        if mtype is None:
+                            mtype = "TYPE IS BLANK"
+                        all_metrics += ('"' + m['monitorType'] + '", "' + mm + '", "' + desc + '", "' + str(default) + '", "' + mtype + '"\n')
+output = codecs.open("all_metrics.csv", "w", "utf-8")
```
keitwb: codecs package is overkill here.  You just need `Path` from `pathlib`.  It can handle writing it out in UTF8.

scripts/dump-metrics-csv.py
```diff
@@ -0,0 +1,49 @@
...
+import sys
+import yaml
+
+all_metrics = ""
+
+for root, dirs, files in os.walk("."):
+    for file in files:
+        if fnmatch.fnmatch(file, "metadata.yaml"):
+             with open(os.path.join(root, file)) as f:
+                data = yaml.load(f)
```
keitwb: Use `yaml.load_safe` so we don't get complains about unsafe things in security scanners.

scripts/dump-metrics-csv.py
```diff
@@ -0,0 +1,49 @@
...
+import fnmatch
+import os
+import sys
+import yaml
+
+all_metrics = ""
+
+for root, dirs, files in os.walk("."):
+    for file in files:
+        if fnmatch.fnmatch(file, "metadata.yaml"):
```
keitwb: If you use pathlib you won't need fnmatch at all.

scripts/dump-metrics-csv.py
```diff
@@ -0,0 +1,49 @@
...
+             with open(os.path.join(root, file)) as f:
```
keitwb: Get this as a pathlib.Path and you can just do `p.read_text()` without the context manager.

PR #974, opened 2019-09-23
Collect empeheral storage stats

asuresh4:


internal/monitors/cadvisor/converter/converter.go
```diff
@@ -11,6 +11,13 @@ import (
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/signalfx/golib/datapoint"
 	"github.com/signalfx/signalfx-agent/internal/utils"
+	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
+)
+
+const (
+	kubernetesPodUID    = "kubernetesPodUID"
```
keitwb: These string values aren't right.  They are supposed to be `_` separated.

keitwb: Also I'm not a huge fan of consting something that is only used twice and adds a level of indirection to figure out what the dim name really is.  I would just use string literals at the map set point.

internal/monitors/cadvisor/metadata.yaml
```diff
@@ -232,4 +232,13 @@ monitors:
...
   monitorType: kubelet-stats
+  metrics:
+    pod_ephemeral_storage_capacity_bytes:
+      description: Represents the storage space available (bytes) for the filesystem
+      default: false
+      type: gauge
+    pod_ephemeral_storage_used_bytes:
+      description: Represents the bytes used on the filesystem
+      default: false
+      type: cumulative
```
keitwb: This seems more like a gauge to me since it can go up and down over time.

asuresh4: Good point, updated to gauge.

internal/monitors/cadvisor/converter/converter.go
asuresh4: Replaced to using string literals.

selfdescribe.json
```diff
@@ -24780,6 +24782,18 @@
           "group": null,
           "default": true
         },
+        "pod_ephemeral_storage_capacity_bytes": {
+          "type": "gauge",
+          "description": "Represents the storage space available (bytes) for the filesystem",
```
keitwb: This is a pretty weak description.  What exactly is the ephemeral storage for a pod?  Is it the same as the root filesystem of the containers or is it something that you have to configure specially?

asuresh4: It's a combination of the root filesystem for containers and empty dir backed volumes (https://kubernetes.io/docs/concepts/storage/volumes/#emptydir)

asuresh4: I'll update the description with some more details

keitwb: Ok, yeah any links to official K8s docs would be good too.  Like the doc that says that it also includes emptyDir volumes.

asuresh4: Updated metric descriptions.

internal/monitors/cadvisor/kubeletstats.go
```diff
@@ -153,3 +154,18 @@ func (kip *kubeletInfoProvider) getAllContainersLatestStats() ([]info.ContainerI
 	}
 	return result, nil
 }
+
+func (kip *kubeletInfoProvider) GetEphemeralStatsFromPods() ([]stats.PodStats, error) {
+	req, err := kip.client.NewRequest("POST", "/stats/summary/", nil)
```
keitwb: Add these stats to a group and then conditionalize the request based on whether any metrics in that group are enabled.  There is a helper method on the Output object to do that.

asuresh4: Added metrics to a group called `podEphemeralStats`

internal/monitors/cadvisor/kubeletstats.go
```diff
@@ -29,6 +30,8 @@ type KubeletStatsConfig struct {
 	config.MonitorConfig
 	// Kubelet client configuration
 	KubeletAPI kubelet.APIConfig `yaml:"kubeletAPI" default:""`
+	// Whether or not to collect ephemeral storage stats from pods
+	IncludePodEphemeralStats bool `yaml:"includePodEphemeralStats" default:"false"`
```
keitwb: The whole point of doing the group thing was so that we don't need flags like this.  The use just has to specify `extraGroups: [podEphemeralStats]` and you can check it using `output.HasEnabledMetricInGroup(podEphemeralStats)`.  This will capture both if they use `extraGroups` or `extraMetrics`.

Get rid of this config.

internal/monitors/cadvisor/kubeletstats.go
```diff
@@ -153,3 +171,22 @@ func (kip *kubeletInfoProvider) getAllContainersLatestStats() ([]info.ContainerI
 	}
 	return result, nil
 }
+
+func (kip *kubeletInfoProvider) GetEphemeralStatsFromPods() ([]stats.PodStats, error) {
+	if !kip.includePodEphemeralStats {
```
keitwb: Do the conditional check in the function that calls this, not in here.  Also use `output.HasEnabledMetricInGroup`, not a config flag on the monitor.

internal/monitors/cadvisor/kubeletstats.go
```diff
@@ -49,8 +52,21 @@ func (ks *KubeletStatsMonitor) Configure(conf *KubeletStatsConfig) error {
...
+		conf.IncludePodEphemeralStats = true
+	}
+
 	return ks.Monitor.Configure(&conf.MonitorConfig, ks.Output.SendDatapoint,
-		newKubeletInfoProvider(client))
+		newKubeletInfoProvider(client, conf.IncludePodEphemeralStats))
+}
+
+// GetExtraMetrics returns additional metrics that should be allowed through.
+func (c *KubeletStatsConfig) GetExtraMetrics() []string {
```
keitwb: You can get rid of this function entirely.  For all new stuff we are purely using the metadata and no flags.

asuresh4: Removed this method

internal/monitors/cadvisor/kubeletstats.go
asuresh4: Since the caller of this method is common to both the `kubelet-stats` and `cadvisor` monitors, would it not be better to do this check here? This check is only valid for `kubelet-stat` monitor. `cadvisor` monitor does not have these metrics.

internal/monitors/cadvisor/kubeletstats.go
asuresh4: Done, removed the config. I hadn't realized the example I followed was more to maintain compatibility with old stuff.

internal/monitors/cadvisor/kubeletstats.go
keitwb: No, this function says it "gets the ephemeral stats from pods" and it should not lie.

asuresh4: Added a `GroupsEnablerHelper` to keep track of enabled groups in the converter. Let me know what you think.

internal/monitors/cadvisor/converter/converter.go
```diff
@@ -49,16 +57,24 @@ type machineInfoMetric struct {
...
+	infoProvider               InfoProvider
+	containerMetrics           []containerMetric
+	containerSpecMetrics       []containerSpecMetric
+	containerSpecMemMetrics    []containerSpecMetric
+	containerSpecCPUMetrics    []containerSpecMetric
+	machineInfoMetrics         []machineInfoMetric
+	podEphemeralStorageMetrics []podStatusMetric
+	sendDP                     func(*datapoint.Datapoint)
+	defaultDimensions          map[string]string
+	groupsEnablerHelper        GroupsEnablerHelper
```
keitwb: This can be the bool based on `HasEnabledMetricInGroup` output.

internal/monitors/cadvisor/converter/converter.go
```diff
@@ -657,6 +704,7 @@ func (c *CadvisorCollector) Collect() {
 	c.collectMachineInfo()
 	c.collectVersionInfo()
 	c.collectContainersInfo()
+	c.collectEphemeralStorageStatsFromPod()
```
keitwb: This is where the conditional goes.

asuresh4: Moved the condition into this method

internal/monitors/cadvisor/converter/converter.go
asuresh4: Updated to pass in just the boolean for this group as we discussed

PR #977, opened 2019-09-24
Reduce docker cache invalidation for agent targets

jchengsfx:
Reorder some steps in the main Dockerfile to help reduce cache invalidations when building the agent and stage caches.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,12 +17,13 @@ RUN cd /tmp &&\
 ENV GOPATH=/go
 WORKDIR /usr/src/signalfx-agent
 
+COPY go.mod go.sum ./
+RUN go mod download
+
 COPY cmd/ ./cmd/
 COPY scripts/make-templates scripts/make-versions ./scripts/
 COPY scripts/collectd-template-to-go ./scripts/
 COPY Makefile .
-COPY go.mod go.sum ./
-RUN go mod download
 COPY internal/ ./internal/
 
 ARG collectd_version=""
@@ -31,6 +32,7 @@ ARG GOOS="linux"
 
 RUN AGENT_VERSION=${agent_version} COLLECTD_VERSION=${collectd_version} make signalfx-agent &&\
     mv signalfx-agent /usr/bin/signalfx-agent
+
 
 ###### Collectd builder image ######
 FROM ubuntu:16.04 as collectd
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -250,23 +253,23 @@ RUN find $PYTHONHOME -name "*.pyc" -o -name "*.pyo" | xargs rm
...
-	  vim
-
-RUN apt install -y openjdk-8-jre-headless &&\
-    mkdir -p /opt/root &&\
+        curl \
+        host \
+        iproute2 \
+        netcat \
+        netcat.openbsd \
+        openjdk-8-jre-headless \
```
keitwb: Take the change to this extra-packages stage out, my PR #976 will fix this and also requires some changes to Java in general.

Dockerfile
```diff
@@ -403,10 +407,23 @@ ENV SIGNALFX_BUNDLE_DIR=/bundle \
     LC_ALL=C.UTF-8 \
     LANG=C.UTF-8
 
+RUN curl -fsSL get.docker.com -o /tmp/get-docker.sh &&\
+    sh /tmp/get-docker.sh
+
+# Get integration test deps in here
 RUN pip3 install ipython ipdb
+COPY python/setup.py /tmp/
+RUN pip3 install -e /tmp/
```
keitwb: I am skeptical of this.  In general the developer should be either using the python runner in the bundle dir in the dev image or referencing the packages directly in the `python` directory from the shared mount of the agent code (that is what the `PYTHONPATH=/usr/src/signalfx-agent/python` above accomplishes).  What were you trying to solve here?

keitwb: Ok, I see it is just copied from below.  Still, it doesn't seem right.

Dockerfile
jchengsfx: > Take the change to this extra-packages stage out, my PR #976 will fix this and also requires some changes to Java in general.

Will revert and wait for your changes.

Dockerfile
jchengsfx: > Ok, I see it is just copied from below. Still, it doesn't seem right.

Yeah, not sure why it was done this way to begin with.  Will change to install using python from the bundle dir instead.

jchengsfx: These lines have been removed.

PR #978, opened 2019-09-24
Collect persistent volume and volume claim phase as metrics

asuresh4:


internal/monitors/kubernetes/cluster/metrics/cache.go
```diff
@@ -161,6 +161,14 @@ func (dc *DatapointCache) HandleAdd(newObj runtime.Object) interface{} {
 		dps = datapointsForCronJob(o)
 		dimProps = dimPropsForCronJob(o)
 		kind = "CronJob"
+	case *v1.PersistentVolume:
```
keitwb: I'm pretty sure you have to start syncing them in the `clusterstate.go` module for this to ever get triggered.

PR #980, opened 2019-09-25
Collect CPU/Memory limits and requests for containers

asuresh4:


docs/monitors/kubernetes-cluster.md
```diff
@@ -77,6 +77,10 @@ Metrics that are categorized as
 (*default*) are ***in bold and italics*** in the list below.
 
 
+ - `kubernetes.container_cpu.limit` (*gauge*)<br>    Maximum CPU limit set for the container
```
keitwb: Change the `.` to `_` so that the metric names are e.g. `kubernetes.container_cpu_limit`.

internal/monitors/kubernetes/cluster/metadata.yaml
```diff
@@ -22,6 +22,22 @@ common:
     resource:
       description: The k8s resource that the quota applies to
   metrics:
+    kubernetes.container_cpu.limit:
+      description: Maximum CPU limit set for the container
```
keitwb: Expand this doc to explain that these is just derived straight from the pod spec and which fields.

internal/monitors/kubernetes/cluster/metrics/containers.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return dps
+}
+
+func datapointsForContainerSpec(c v1.Container, contDims map[string]string) []*datapoint.Datapoint {
+
+	dps := []*datapoint.Datapoint{
+		datapoint.New(
+			"kubernetes.container_cpu.request",
```
keitwb: Same thing here about metric name.

internal/monitors/kubernetes/cluster/metrics/containers.go
```diff
@@ -0,0 +1,110 @@
...
+			contDims,
+			datapoint.NewIntValue(c.Resources.Limits.Cpu().Value()),
```
keitwb: If I don't specify any value for cpu limit, will this just return a `0`?

Ideally if there is no specified limit/request in the pod spec, then no datapoints would be emitted for that metric.

internal/monitors/kubernetes/cluster/metrics/cache.go
```diff
@@ -204,7 +211,10 @@ func (dc *DatapointCache) handleAddPod(pod *v1.Pod) ([]*datapoint.Datapoint,
 		dc.podCache.AddPod(pod)
 	}
 	cachedPod := dc.podCache.GetCachedPod(pod.UID)
-	dps := datapointsForPod(pod)
+	dps, containerDimProps := datapointsForPod(pod)
+
+	dc.addDimPropsListToCache(containerDimProps)
```
keitwb: Let's drop these properties for now.

internal/monitors/kubernetes/cluster/metrics/pods.go
```diff
@@ -32,28 +31,32 @@ func datapointsForPod(pod *v1.Pod) []*datapoint.Datapoint {
...
-		contDims["container_spec_name"] = cs.Name
-		contDims["container_image"] = cs.Image
-
-		dps = append(dps, datapoint.New(
-			"kubernetes.container_restart_count",
-			contDims,
-			datapoint.NewIntValue(int64(cs.RestartCount)),
-			datapoint.Gauge,
-			time.Now()))
+		contDims := getAllContainerDimensions(cs.ContainerID, cs.Name, cs.Image, dimensions)
```
keitwb: `cs.ContainerID` does not get the `docker://` prefix stripped off here or in the `getAllContainerDimensions` function.  Is that an oversight or intentional?

tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
```diff
@@ -205,3 +205,29 @@ def test_cronjobs(k8s_cluster):
...
+CONTAINER_COMPUTE_RESOURCE_METRICS = [
+    "kubernetes.container_cpu.limit",
+    "kubernetes.container_cpu.request",
+    "kubernetes.container_memory.limit",
+    "kubernetes.container_memory.request",
+]
+
+
+@pytest.mark.kubernetes
+def test_container_compute_resouorce_metrics(k8s_cluster):
```
keitwb: Expand the integration tests to actually test for container dimensions, esp for the logic that strips off the `docker://` prefix.

You can use the K8s Python client to get the running pods and see what the expected pod names and container names should be and then assert that you are getting datapoints with those exact dimensions.

docs/monitors/kubernetes-cluster.md
asuresh4: Updated metric names

internal/monitors/kubernetes/cluster/metrics/containers.go
asuresh4: Yeah, if no limit is set, this will return a zero. Will handle that case.

internal/monitors/kubernetes/cluster/metrics/cache.go
asuresh4: Done

internal/monitors/kubernetes/cluster/metrics/pods.go
asuresh4: I might have lost it during a rebase I did.  Added a test for this.

tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
```diff
@@ -275,3 +275,30 @@ def test_deployments(k8s_cluster):
                     },
                 )
             )
+
+
+@pytest.mark.kubernetes
+def test_containers(k8s_cluster):
```
keitwb: Test that containers with limits/requests go through as metrics.  Also test the case where there is a missing limit/request and make sure no metrics are generated for it.

internal/monitors/kubernetes/cluster/metadata.yaml
asuresh4: Added details about metrics

tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
asuresh4: Updated the tests to check for this.

PR #982, opened 2019-09-26
Container properties

keitwb:


internal/monitors/kubernetes/cluster/metrics/containers.go
```diff
@@ -0,0 +1,45 @@
...
+	atypes "github.com/signalfx/signalfx-agent/internal/monitors/types"
+	v1 "k8s.io/api/core/v1"
+)
+
+func dimPropsForContainer(cs v1.ContainerStatus) *atypes.DimProperties {
+
+	containerProps := make(map[string]string)
+
+	if cs.State.Running != nil {
+		containerProps["container_status"] = "running"
```
keitwb: Make it set `containerProps["container_status_reason"] = ""` and then add the `MergeIntoExisting: true` flag to the `DimProperties` struct below so that this thing can work with the container property updates that I'm implementing for the container <---> service correlation feature.

asuresh4:
Will open a PR after rebasing with recent changes.

PR #984, opened 2019-09-26
Reorder endpoint config rendering so output gets everything

keitwb:
Before the output object was getting only the values specified in the agent yaml config
and not config that came from the endpoint itself

### DIFF
--- a/internal/monitors/activemonitor.go
+++ b/internal/monitors/activemonitor.go
@@ -30,38 +30,40 @@ type ActiveMonitor struct {
 	doomed bool
 }
 
-// Does some reflection magic to pass the right type to the Configure method of
-// each monitor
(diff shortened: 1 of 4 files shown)


keitwb:
> I feel like there has to be a better way to structure the config code in general. It's just way too tricky.

Any specific suggestions on how?  There are multiple config sources that have to be merged together.  I thought about making a separate struct to aggregate everything but it would basically just have that `renderConfig` method on it so it seemed simpler to just have a function.

jaysfx:
> Any specific suggestions on how?

Not really, it's one of those things I'd have to play around with. I'm talking about all the config code not just this particular piece.

keitwb:
> > Any specific suggestions on how?
> 
> Not really, it's one of those things I'd have to play around with. I'm talking about all the config code not just this particular piece.

Alright, well I think the `renderConfig` function at least marginally simplified it by putting it in to a pure function.

PR #986, opened 2019-09-26
add haproxy_status metric

bjsignalfx:


internal/monitors/haproxy/metadata.yaml
```diff
@@ -59,6 +59,10 @@ monitors:
       unique_proxy_id:
         description: 'A unique proxy id'
     metrics:
+      haproxy_status:
+        default: false
+        description: '. Values reported for listeners, frontends, backends and servers.'
```
keitwb: Leading `.`.

keitwb: Also this description is quite vague.  What kind of values is this supposed to be?  Is it some kind of enum?

keitwb: Ok, better now but you need to remake the docs.

bjsignalfx: Remade the docs

PR #988, opened 2019-09-26
Pin rubyzip version for windows chef tests

jchengsfx:
Chef tests on windows are failing because`rubyzip` was recently updated to require ruby >= v2.4. 
The chef-client currently provides ruby <= v2.3 and the windows cookbook (which has a dependency on rubyzip) does not pin the rubyzip version.

### DIFF
--- a/tests/deployments/chef/chef_test.py
+++ b/tests/deployments/chef/chef_test.py
@@ -189,14 +189,16 @@ def run_win_chef_client(backend, agent_version, stage):
 def run_win_chef_setup(chef_version):
     assert has_choco(), "choco not installed!"
     uninstall_win_agent()
+    if WIN_CHEF_BIN_DIR not in os.environ.get("PATH"):
+        os.environ["PATH"] = WIN_CHEF_BIN_DIR + ";" + os.environ.get("PATH")
     if run_win_command("chef-client --version", []).returncode == 0:
         run_win_command("choco uninstall -y -f chef-client")
     if chef_version == "latest":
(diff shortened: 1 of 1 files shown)


tests/deployments/chef/chef_test.py
```diff
@@ -189,14 +189,16 @@ def run_win_chef_client(backend, agent_version, stage):
...
     if run_win_command("chef-client --version", []).returncode == 0:
         run_win_command("choco uninstall -y -f chef-client")
     if chef_version == "latest":
         run_win_command(f"choco upgrade -y -f chef-client")
     else:
         run_win_command(f"choco upgrade -y -f chef-client --version {chef_version}")
-    if WIN_CHEF_BIN_DIR not in os.environ.get("PATH"):
-        os.environ["PATH"] = WIN_CHEF_BIN_DIR + ";" + os.environ.get("PATH")
+    # pin rubyzip to v1.3.0 to prevent the windows cookbook from installing newer incompatible versions
+    run_win_command(r"C:\opscode\chef\embedded\bin\gem install rubyzip -f -q -N -v 1.3.0")
```
keitwb: Is this not something we should fix in the cookbook dependency metadata instead?  Won't people using the cookbook run into the same potential issue?

jchengsfx: Possibly, will take a look. Just didn't want to introduce any side effects that could cause potential conflicts for any customers that may not be using the embedded ruby.

PR #993, opened 2019-09-27
Update quick-install.md

galpizarsfx:
This is testing the Smart Agent tile to see if we can manipulate text that appears both in the UI and in docs.signalfx. This needs to be merged so that we can see the results in lab. Afterwards, whether or not the test is successful, this will need to be unmerge so that in another PR, I will make the official change, along with some other updates.

### DIFF
--- a/docs/quick-install.md
+++ b/docs/quick-install.md
@@ -70,17 +70,16 @@ _Note: if you have previously configured another metric collection agent on your
 
 __Linux:__ Dependencies are completely bundled along with the agent, including a Java JRE runtime and a Python runtime, so there are no additional dependencies required. The agent works on any modern Linux distribution (kernel version 2.6+).
 
-If you are not installing from the tile on the Integrations page:
-
-- Get your API_TOKEN from: __Organization Settings => Access Token__ tab in the SignalFx application. 
-
-- Determine YOUR\_SIGNAL_FX_REALM from your [profile page](https://docs.signalfx.com/en/latest/getting-started/get-around-ui.html#user-profile-avatar-and-color-theme) in the SignalFx web application.
-
 To install the Smart Agent on a single Linux host, enter:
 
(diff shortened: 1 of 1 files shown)


docs/quick-install.md
```diff
@@ -70,17 +70,16 @@ _Note: if you have previously configured another metric collection agent on your
...
-- Determine YOUR\_SIGNAL_FX_REALM from your [profile page](https://docs.signalfx.com/en/latest/getting-started/get-around-ui.html#user-profile-avatar-and-color-theme) in the SignalFx web application.
-
 To install the Smart Agent on a single Linux host, enter:
 
-```sh
-curl -sSL https://dl.signalfx.com/signalfx-agent.sh > /tmp/signalfx-agent.sh
+```sh curl -sSL https://dl.signalfx.com/signalfx-agent.sh > /tmp/signalfx-agent.sh
 sudo sh /tmp/signalfx-agent.sh --realm YOUR_SIGNALFX_REALM YOUR_SIGNALFX_API_TOKEN
+
+<!-- In the above command, YOUR_SIGNALFX_REALM represents the realm that is shown on your profile page
```
keitwb: I don't think you need to make this a comment, just make it regular markdown.

galpizarsfx:
I thought the comment would help to manipulate the appearance of the text
between the tile and the doc site. In other words,

* In the tile, the text would display the user's specific realm and token.
* In the doc site, the same text would literally display
YOUR_SIGNALFX_REALM.

Does that make sense?

George

keitwb:
No, it will just make it disappear everywhere if it is a comment.

galpizarsfx:
Okay, then in that case, we can simply ignore and close this request. As an FYI, in the near future, I'll be sending another PR which will reorganize the Smart Agent content with some editing as well.

PR #994, opened 2019-09-27
Set max rubyzip version for chef cookbook

jchengsfx:
Latest `rubyzip` gem is incompatible with ruby versions < 2.4 bundled with chef-client.

### DIFF
--- a/deployments/chef/metadata.rb
+++ b/deployments/chef/metadata.rb
@@ -17,5 +17,7 @@ supports 'windows'
 
 depends 'windows', '>= 4.3.4'
 
+gem 'rubyzip', '< 2.0.0'
+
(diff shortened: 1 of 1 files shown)


deployments/chef/metadata.rb
```diff
@@ -17,5 +17,7 @@
 
 depends 'windows', '>= 4.3.4'
 
+gem 'rubyzip', '< 2.0.0'
```
keitwb: I'm a bit wary of how this will interact with people's environments, but I guess we can try it as long as we make it a separate release that people can roll back if problematic.

jchengsfx: > I'm a bit wary of how this will interact with people's environments, but I guess we can try it as long as we make it a separate release that people can roll back if problematic.

I had the same concerns, but the chef-client should be able to handle multiple versions of the same gem if necessary.

PR #996, opened 2019-09-27
Update to-integrations-repo script to handle logstash docs

asuresh4:


docs/monitors/logstash-tcp.md
```diff
@@ -94,6 +93,10 @@ output {
 }
 ```
 
+Once Logstash is configured with the above configuration. The logstash-tcp monitor will
+collect `lognis.count` and `process_time.<timer_field>`. See config options for default
```
keitwb: `lognis`..

PR #1003, opened 2019-09-30
Changes for Non-Docker Kubernetes runtime

keitwb:
See individual commits.

This gives us a decent support for cri-o and containerd, but there are still some gaps:

 - cri-o is missing pod networks stats
 - containerd is missing container filesystem usage stats
 - both cri-o and containerd are missing collectd/df metrics.  This is mitigated by #1016.

### DIFF
--- a/deployments/k8s/configmap.yaml
+++ b/deployments/k8s/configmap.yaml
@@ -46,6 +46,14 @@ data:
     - type: kubelet-stats
       kubeletAPI:
         authType: serviceAccount
+      datapointsToExclude:
+      - dimensions:
+          container_image:
+           - '*pause-amd64*'
+           - 'k8s.gcr.io/pause*'
+        metricNames:
+          - '*'
+          - '!*network*'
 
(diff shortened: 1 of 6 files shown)


deployments/k8s/configmap.yaml
```diff
@@ -46,6 +46,14 @@ data:
     - type: kubelet-stats
```
asuresh4: looks like we've also got `docker-container-stats` monitors enabled. do we it when we've . got `kubelet-stats`monitor enabled?

keitwb: I need to make that a flag so people not using Docker can easily turn it off.

jchengsfx:
K8s test jobs removed from circleci for versions < 1.12.  Will update required PR checks when this is merged.

deployments/k8s/configmap.yaml
keitwb: I checked and this is already there.

PR #1006, opened 2019-10-01
Update to-integrations-repo script to handle logstash docs

asuresh4:


asuresh4:
@keitwb - I'll be adding subsequent commits to handle config tables. But it would great if you could take a look. Results of running the script: https://github.com/signalfx/integrations/pull/759

scripts/docs/templates/integrations_doc.txt
```diff
@@ -0,0 +1,157 @@
+
```
keitwb: Call this `integrations_docs.md.jinja2` to make it clear it is a template for a markdown file.

scripts/docs/to-integrations-repo_v2
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python3
```
keitwb: Just replace the old script whenever you are done with these changes.

scripts/docs/to-integrations-repo_v2
```diff
@@ -0,0 +1,186 @@
...
+# Syncs all the relevant docs/metadata from the agent to the integrations repo.
+
+import copy
+import json
+import os
+import re
+import sys
+
+from collections import OrderedDict
+from jinja2 import Environment, FileSystemLoader
```
keitwb: Add the jinja dep to a requirements.txt file somewhere and pin it to a specific version.

scripts/docs/templates/integrations_doc.txt
asuresh4: updated

scripts/docs/to-integrations-repo_v2
asuresh4: yep, that's the plan. wanted to make it easier to review.

scripts/docs/to-integrations-repo_v2
asuresh4: done

scripts/docs/to-integrations-repo_v2
```diff
@@ -0,0 +1,312 @@
...
+    j2_env.filters["surround_by_tick"] = surround_by_tick
+    j2_env.filters["has_atleast_one_monitor_with_section"] = has_atleast_one_monitor_with_section
+    j2_env.filters["send_all_metrics_from_all_monitors"] = send_all_metrics_from_all_monitors
+
+    return j2_env
+
+
+def get_template():
+    j2_env = get_jinja_env()
+    template = j2_env.get_template("integrations_doc.md.jinja2")
```
keitwb: Make sure this script will work the same no matter what the PWD is when run.

asuresh4: A file system loader object is passed to the environment `j2_env`. Any template in the path passed into the filesystem loader object, can be accessed in this manner. `INTEGRATIONS_DOC_TEMPLATE_DIR` is passed in and is derived from the current script directory. That should suffice right?

keitwb: I suppose so, just make sure that you can run the script equally well when your shell is in the root of the repo as from the same directory where the script is.

docs/monitors/logstash-tcp.md
```diff
@@ -93,6 +93,9 @@ output {
   }
 }
 ```
+Once Logstash is configured with the above configuration. The logstash-tcp monitor
+will collect `logins.count` and `process_time.<timer_field>`. See config options
+for default
```
keitwb: Period needed at end of sentence.

keitwb:
```suggestion
for default values.
```

asuresh4: Done

PR #1014, opened 2019-10-07
add hpa metrics to k8s cluster monitor

bjsignalfx:


internal/monitors/kubernetes/cluster/clusterstate.go
```diff
@@ -76,6 +78,7 @@ func (cs *State) Start() {
 		cs.beginSyncForType(ctx, &quota.ClusterResourceQuota{}, "clusterresourcequotas", v1.NamespaceAll,
 			cs.quotaClient.RESTClient())
 	}
+	v2beta1Client := cs.clientset.AutoscalingV2beta1().RESTClient()
```
keitwb: Call this `hpaV2Beta1Client` instead to be more specific.

PR #1016, opened 2019-10-08
Fix filesystems monitor for Linux

keitwb:


internal/monitors/filesystems/filesystems.go
```diff
@@ -152,21 +152,21 @@ func (m *Monitor) emitDatapoints() {
 
 		// disk utilization
 		m.Output.SendDatapoint(datapoint.New(diskUtilization,
-			utils.MergeStringMaps(map[string]string{"plugin": types.UtilizationMetricPluginName}, commonDims),
+			utils.MergeStringMaps(commonDims, map[string]string{"plugin": types.UtilizationMetricPluginName}),
```
asuresh4: looks like we don't send in common dimensions for `diskSummaryUtilization`, is that intentional?

keitwb: It is intentional because the summary utilization is an aggregate of all partitions so all the dimensions are rolled up.

PR #1021, opened 2019-10-09
Make haproxy monitor work with auto-discovery

keitwb:


internal/monitors/haproxy/config.go
```diff
@@ -1,16 +1,35 @@
...
+	// then specify the `url` config option with the format `unix://...` and
+	// omit `host`, `port` and `useHTTPS`.
+	UseHTTPS bool `yaml:"useHTTPS"`
+	// The path to HAProxy stats. The default is `stats?stats;csv`.
+	// This is used to construct the `url` option if not provided.
+	Path string `yaml:"path" default:"stats?stats;csv"`
+
+	// URL on which to scrape HAProxy. Scheme `http://` for http-type and
+	// `unix://` socket-type urls.  If this is not provided, it will be derive
+	// from the `host`, `port`, and `useHTTPS` options.
```
keitwb:
```suggestion
	// from the `host`, `port`, `path`, and `useHTTPS` options.
```

internal/monitors/haproxy/config.go
```diff
@@ -23,3 +42,27 @@ type Config struct {
 	// If empty then metrics for all proxies will be reported.
 	Proxies []string `yaml:"proxies"`
 }
+
+// Validate the config
+func (c *Config) Validate() error {
```
keitwb: This validation is too constraining.  There are various edge-cases where it is useful to be able to mix and match discovery rules and configs in ways that we cannot foresee, so I am more inclined towards letting the user do whatever they want.  As long as it is clearly documented how the various config options relate to each other (I think it is), then it should be fine.

Get rid of this method.

tests/monitors/haproxy/haproxy_test.py
```diff
@@ -33,7 +33,23 @@ def test_haproxy_default_metrics_from_stats_page(version):
              url: http://{host}:8080/stats?stats;csv
            """
         ) as agent:
+            verify(agent, EXPECTED_DEFAULTS - EXPECTED_DEFAULTS_FROM_SOCKET, 10)
+            assert not has_log_message(agent.output.lower(), "error"), "error found in agent output!"
+
 
+@pytest.mark.parametrize("version", ["1.8"])
+def test_haproxy_default_metrics_from_stats_page_by_discovery_rule(version):
+    with run_service("haproxy", buildargs={"HAPROXY_VERSION": version}, name="haproxy", ports={8080: 8080}):
```
keitwb: In `test-services/haproxy/Dockerfile`, add a line that says `EXPOSE 8080` and then you shouldn't have to do the `ports={8080:8080}` thing, which is not very robust since you couldn't run multiple tests with the same port simultaneously.

internal/monitors/haproxy/config.go
bjsignalfx: ok

tests/monitors/haproxy/haproxy_test.py
bjsignalfx: ok

PR #1025, opened 2019-10-09
897 update agent script

galpizarsfx:
The Windows command was recently updated; however, in that recent update, the REALM and TOKEN text did not auto-populate in the UI. This update will (hopefully) update the command to properly populate those fields.


galpizarsfx:
@keitwb Just as a reminder to merge my updates into master.

PR #1030, opened 2019-10-14
Fix azure tag sync when hostname is not lower cased

jaysfx:
On the asynctasker side we lower case the resource id when reporting values. So
that the resource ids match up we also need to lower case them on the agent
side.

Also fix an issue where the build would fail removing genmetadata.go when
running from a shared folder (theory is that there's a path too long somewhere
in the directory structure) so just limit the cleanup to internal/monitors.


asuresh4:
We could sync properties to both dimensions from asynctasker. That'll help in not creating new MTSes.

keitwb:
@jaysfx is going to do an analysis of how many MTSs we have that have upper cased azure_resource_id and see what the impact will be.

PR #1031, opened 2019-10-15
K8s v1.16

jchengsfx:
Dropped K8s v1.9.0, v1.10.0, and v1.11.0 test jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ executors:
 
   helm:
     docker:
-      - image: alpine/helm:2.13.0
+      - image: alpine/helm:2.15.0
 
 
 commands:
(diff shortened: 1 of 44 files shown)


.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ executors:
 
   helm:
     docker:
-     - image: alpine/helm:2.13.0
+     - image: alpine/helm:2.15.0-rc.1
```
keitwb: Don't merge this until this becomes a final release.

deployments/k8s/generate-from-helm
```diff
@@ -19,7 +19,8 @@ for f in $templates; do
     -x ${AGENT_CHART_DIR%/}/templates/$f \
     $AGENT_CHART_DIR | \
       sed -e 's/[[:space:]]*$//' | \
-      sed -e 's@apiVersion: apps/v1beta2@apiVersion: extensions/v1beta1@' | \
+      sed -e 's@apiVersion: apps/v1beta2@apiVersion: apps/v1@' | \
```
keitwb: Just fix the original resource templates to use apps/v1 and get rid of this sed stuff.  The only reason we had it to begin with was for when we had the Helm chart in the managed repository and they required all apiVersions to be the latest available.

PR #1032, opened 2019-10-15
Fix some doc links to open in a new window

seonsfx:


docs/faq.md
```diff
@@ -114,8 +114,8 @@ the agent instead of the default `/bin/signalfx-agent`, as well by adding the
 ...
 ```
 
-The source for the script `/bin/umount-hostfs-non-persistent` can be [found
-here](https://github.com/signalfx/signalfx-agent/blob/master/scripts/umount-hostfs-non-persistent),
+The source for the script `/bin/umount-hostfs-non-persistent` can be <a target="_blank" href="https://github.com/signalfx/signalfx-agent/blob/master/scripts/umount-hostfs-non-persistent">found
```
keitwb: I thought the whole point of your changes in the script was to make this unnecessary?

seonsfx: thought we would keep the current workflow which statically moves "agent_docs" without any modification as the number of links are just handful. I can do whatever we think makes more sense though.

keitwb: Yeah I want the original copies of the agent docs to be as "pure" Markdown as possible.

asuresh4: in the context of agent docs in product-docs, it may be a safe assumption to say links not going to "https://docs.signalfx.com/..." or "https://github.com/signalfx/signalfx-agent/docs/..." should open in a new tab right?

seonsfx: @asuresh4 True

PR #1033, opened 2019-10-15
Add container status as property

asuresh4:


asuresh4:
@keitwb - i've addressed the comment you made on https://github.com/signalfx/signalfx-agent/pull/982 about making these changes compatible with the service correlation feature you're working on

tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
```diff
@@ -309,6 +309,15 @@ def test_containers(k8s_cluster):
                     container_id = container_status.container_id.replace("docker://", "").replace("cri-o://", "")
                     containers_cache[container_status.name] = container_id
                     assert wait_for(p(has_datapoint, agent.fake_services, dimensions={"container_id": container_id}))
+                    assert wait_for(
+                        p(
+                            has_dim_prop,
```
keitwb: Be more specific here on the expected value of `container_status`.  Would be nice to test for the terminated/waiting states too but not sure how easy it is to force a container into that state when managed by Kubernetes.  If it isn't too convoluted to do, test for those other states as well.

asuresh4: Updated the tests to pick up container status from the container objects retrieved. Let me know what you think.

internal/monitors/kubernetes/cluster/metrics/pods.go
```diff
@@ -11,7 +11,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func datapointsForPod(pod *v1.Pod) []*datapoint.Datapoint {
+func datapointsForPod(pod *v1.Pod) ([]*datapoint.Datapoint, []*atypes.DimProperties) {
```
keitwb: This function should only return datapoints.  Why can't the dimProps stuff be done in the `dimPropsForPods` function below?

tests/monitors/kubernetes_cluster/kubernetes_cluster_test.py
keitwb: Seems good.

internal/monitors/kubernetes/cluster/metrics/pods.go
asuresh4: `dimPropsForPods` does not have context about the pod object itself. It makes use of what we cache. I've moved this logic out to `handleAddPod` now.

asuresh4:
Closing this as per discussion with @keitwb.

asuresh4:
Changes LGTM but can't approve since I originally opened the PR.

PR #1034, opened 2019-10-16
Chef apt version fix

jchengsfx:


keitwb:
Don't add tests for 12.04.  We don't support it and shouldn't waste CI resources testing it.

PR #1035, opened 2019-10-17
Add kubernetes_workload_name property

asuresh4:


internal/monitors/kubernetes/cluster/metrics/deployments.go
```diff
@@ -30,6 +30,7 @@ func datapointsForDeployment(dep *v1beta1.Deployment) []*datapoint.Datapoint {
 func dimPropsForDeployment(dep *v1beta1.Deployment) *atypes.DimProperties {
 	props, tags := k8sutil.PropsAndTagsFromLabels(dep.Labels)
 	props["kubernetes_workload_name"] = dep.Name
+	props["deployment"] = dep.Name
```
asuresh4: Put "deployment" property back here since it seems like it may be useful to have per comment in https://github.com/signalfx/signalfx-agent/pull/621

PR #1037, opened 2019-10-20
Improve hiera default template and add chocolatey comment block

kaypeter87:
Using the `%{::hostname}` interpolation fact will have SFX use shortnames on the hostname key. This will be better for filtering on the SFX platform. 

Also, since Windows can have inconsistent casing, adding a line in the `agent.yaml.erb` ruby remplate to downcase all hostname values will allow us to filter using wildcards. Currently, the filtering for plots in SFX is case sensitive. 

Lastly, I added a comment block to account for the future `chocolatey` package release. This will pave the way to put in conditionals for either using chocolatey or direct SFX repos.


kaypeter87:
Regarding the circleci test failing on puppet:
I've seen this error before, you probably need to add facts to the stubbing so that hiera could find/lookup the correct keys. This is only a speculation.

keitwb:
Thanks @kaypeter87, but I am wary of this because we have certain assumptions in some of our built-in content and monitors that the `host` dimension is unique at any given time across an entire org.  If you reuse short hostnames across various environments, this could be a problem.  Even if your organization doesn't, I know it is not terribly uncommon in others.  That is why we use FQDN here, and in fact if you don't provide the `hostname` config option, the agent tries to resolve a FQDN by doing a reverse lookup on the IP address that the short host name resolves to.

If you are fairly confident that you don't have much short hostname reuse, you can just set that config to the short hostname in your own facts data, but I don't want to make this default.  I'm assuming you can do the lowercasing part without having to modify the template?

For the chocolately thing, we will add support for that whenever a chocolately package is available.

kaypeter87:
@keitwb Fair point on other orgs re-using short hostnames.

The only issue we have with the default FQDN doing a reverse lookup on the IP is that it will retrieve that data from DNS. How is the case handled for multiple records for any given IP addresses? Since reverse lookups are based on DNS' query system, it would return whatever DNS decides what the first record is(I believe this is alphanumerical order). This is an assumption that most orgs have stale DNS records (in many cases they do).

The `%{::hostname}` interpolated value uses facter to get an already known fact. It is essentially the node's hostname, as derived from its validated certificate name, which would be a more accurate representation of the host itself.

As far as the downcasing on the template level, it just makes more sense to extrapolate there as oppose to creating additional puppet classes and file declarations for solving just a casing problem. But, it would be a non-issue if filtering on plots aren't case sensitive :).

keitwb:
There is a config option `useFullyQualifiedHost: false` that you can set in the agent if you want it to use the default short hostname (the hostname that the kernel reports) and skip the reverse lookup process.  This only has any effect if you are not setting `hostname` in the config, as the agent will automatically add a `host` dimension if you don't.  By setting `useFullyQualifiedHost: false` and omitting `hostname` from your config, you will get the effect you are looking for (except for the downcasing thing).  I assume it is possible to somehow tell Puppet to unset a "default" fact?  (I'm not terribly familiar with Puppet)

I really haven't dealt with mixed case host names before, so I'm not sure what the ideal solution is.  I suppose some people might want the casing preserved if they crafted host names with mixed case to begin with.  I would just do it with a custom fact or something that is derived from hostname, as this is too specific of an issue to force on everybody who may or may not care about casing.

kaypeter87:
Thanks for taking the time to respond on these issues. My issue with `useFullyQualifiedHost: false` is that this does not seem to work as intended. We have tested across our org without success. I'll see if we can file an issue regarding this and provide reports on re-producing the issue.

Mix-casing has always been a problem in mixed-environments of Linux/Windows. This issue only arises on the Windows side as uppercasing is allowed on hostnames. This causes the org to create two plots for filtering on 1 subset of hosts.

ie.
Hosts: `linux-web1.domain.com`, `linux-web2.domain.com`, `WINDOWS-WEB1.DOMAIN.COM`, `WINDOWS-WEB2.DOMAIN.COM`

Plot A - Filter: `*-web*`
Plot B - Filter: `*-WEB*`

I'll file a feature request for this separate issue on filtering.

PR #1039, opened 2019-10-22
Service container link

keitwb:
This is a new feature which does two main things:

1) Adds `container_id` and `kubernetes_pod_uid` tags to spans that come through the agent that identify where the spans came from.

2) Add `service-name` and `cluster-name` properties to the `container_id` and `kubernetes_pod_uid` dimensions to identify the trace span service and cluster associated with those container/pods.

It also involved refactoring a lot of things around dimension handling.

See individual commits for more details.


internal/monitors/kubernetes/cluster/metrics/pods.go
```diff
@@ -50,56 +51,68 @@ func datapointsForPod(pod *v1.Pod) []*datapoint.Datapoint {
...
 
-	_ = getPropsFromTolerations(cachedPod.Tolerations)
+	_ = getPropsFromTolerations(pod.Spec.Tolerations)
 
-	// if pod is selected by a service, sync service as a tag
-	serviceTags := sc.GetMatchingServices(cachedPod)
-	for _, tag := range serviceTags {
-		tags["kubernetes_service_"+tag] = true
+	if len(props) == 0 && len(tags) == 0 {
+		return nil
```
asuresh4: is this check required? it seems redundant since we set `MergeIntoExisting` to true.

internal/monitors/kubernetes/utils/servicecache.go
```diff
@@ -29,98 +28,82 @@ func NewServiceCache() *ServiceCache {
...
+		Selector:  svc.Spec.Selector,
+	}
 }
 
 // AddService adds or updates a service in cache
-// This function should only be called after sc.IsCached
-// to prevent unnecessary updates to the internal cache.
 // Returns true if the service is added, false if it was ignored
-func (sc *ServiceCache) AddService(svc *v1.Service) bool {
+func (sc *ServiceCache) AddService(svc *v1.Service) {
```
asuresh4: Comment above the function needs to be updated reflecting the update

internal/monitors/kubernetes/utils/servicecache.go
```diff
@@ -29,98 +28,82 @@ func NewServiceCache() *ServiceCache {
...
-func (sc *ServiceCache) Delete(svc *v1.Service) ([]types.UID, error) {
+func (sc *ServiceCache) Delete(svc *v1.Service) error {
 	return sc.DeleteByKey(svc.UID)
 }
 
 // DeleteByKey removes a service from the cache given a UID
 // Returns pods that were previously matched by this service
 // so their properties may be updated accordingly
-func (sc *ServiceCache) DeleteByKey(svcUID types.UID) ([]types.UID, error) {
+func (sc *ServiceCache) DeleteByKey(svcUID types.UID) error {
```
asuresh4: Same thing regarding comments on the function

internal/monitors/kubernetes/cluster/metrics/pods.go
keitwb: No, and in fact returning nil is very bad so I will remove this whole conditional.

internal/core/writer/dimensions/client.go
```diff
@@ -0,0 +1,305 @@
...
+
+		// Don't further delay it if it gets updated so that we are always
+		// guaranteed to update a dimension at least some times, even if it
+		// continually gets updated more frequently than `sendDelay` seconds
+		// (which should be dealt with somewhere else).
+
+		// If the dim is a merge request, then update the existing one in
+		// place, otherwise replace it.
+		if delayedDim.MergeIntoExisting {
+			delayedDim.Properties = utils.MergeStringMaps(delayedDim.Properties, filteredDim.Properties)
```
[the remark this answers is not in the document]
keitwb: I don't think so because the filtering always happens before it gets to this point in the code.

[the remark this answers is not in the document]
keitwb: Do you see anything specific that is concerning?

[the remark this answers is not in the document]
asuresh4: I was wondering if the dimensions set cache could end up with stale properties. But I don't think that's the case because it's guaranteed that we have a clean cache whenever the property filters are updated.

internal/monitors/kubernetes/cluster/metrics/dimensions.go
```diff
@@ -243,3 +255,67 @@ func (dh *DimensionHandler) handleAddJob(job *batchv1.Job) {
 		dh.emitDimension(dimensionForPodCronJob(pod, cronJobRef.Name, cronJobRef.UID))
 	}
 }
+
+type dimensionDeduplicator struct {
```
asuresh4: this seems like something that does not necessarily need to confined to this monitor. do you want to move this some place more global?

keitwb: Yes, I thought about it some more and decided to move it back into the writer's dimension client.

PR #1041, opened 2019-10-22
Add monitor for kube-proxy

seonsfx:


docs/monitors/kubernetes-proxy.md
```diff
@@ -0,0 +1,164 @@
...
+Exports Prometheus metrics from the [kube-proxy](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy)
+metrics in Prometheus format. The monitor queries path `/metrics` by default when no path is configured. The monitor converts
+the Prometheus metric types to SignalFx metric types as described [here](prometheus-exporter.md)
+
+Example YAML Configuration
+
+```yaml
+monitors:
+- type: kubernetes-proxy
+  host: 127.0.0.1
```
keitwb: Once #907 is merged, you will be able to do a discovery rule like this:

```
discoveryRule: pod_name == "kube-proxy" && target == "pod"
port: 10249
```

seonsfx: @keitwb `kube-proxy` metrics endpoint is only accessible through localhost so it will be discovered but still won't get any metrics. For example, on my `seon-eks` cluster on AWS, the agent is able to discover and retrieve host of `kube-proxy` which is 192.168.6.21, but connection is refused when the agent tries to get the metrics.

keitwb:
Remove the portless branch commits from this.

PR #1045, opened 2019-10-24
Run patch-interpreter during rpm/deb stages

jchengsfx:
https://signalfuse.atlassian.net/browse/INT-1609


jchengsfx:
> Looks good. As a double check can you deploy this on a real centos 6 machine and make sure it still works?

Installed and ran without issue on centos6 host in EC2.

packaging/rpm/signalfx-agent.spec
```diff
@@ -49,7 +49,7 @@ install -p -m 644 %{_sourcedir}/signalfx-agent.1 $RPM_BUILD_ROOT/%{_mandir}/man1
 
 %files
 
-%config(noreplace) /etc/signalfx/agent.yaml
+%config(noreplace) %verify(not md5 size mtime) /etc/signalfx/agent.yaml
```
jchengsfx: @jaysfx @keitwb let me know if I should exclude any other attributes from rpm verification for the agent.yaml.

List of available attributes from http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html:
- Owner (owner)
- Group (group)
- Mode (mode)
- MD5 Checksum (md5)
- Size (size)
- Major Number (maj)
- Minor Number (min)
- Symbolic Link String (symlink)
- Modification Time (mtime)

jaysfx: lgtm, does marking it as a config file not implicitly turn off verification?

jchengsfx: I assumed it would ignore config files, but looks like that is the intended behavior for rpm, so the workaround is to exclude the attributes in the spec (see similar issue in centos https://bugs.centos.org/view.php?id=11349).

PR #1051, opened 2019-10-28
Add etcd prometheus monitor

asuresh4:


docs/monitors/gitlab.md
```diff
@@ -174,6 +174,9 @@ Configuration](../monitor-config.md#common-configuration).**
 | `password` | no | `string` | Basic Auth password to use on each request, if any. |
 | `useHTTPS` | no | `bool` | If true, the agent will connect to the exporter using HTTPS instead of plain HTTP. (**default:** `false`) |
 | `skipVerify` | no | `bool` | If useHTTPS is true and this option is also true, the exporter's TLS cert will not be verified. (**default:** `false`) |
+| `caCertPath` | no | `string` | Path to the CA cert that has signed the Kubelet's TLS cert, unnecessary if `skipVerify` is set to false. |
+| `clientCertPath` | no | `string` | Path to the client TLS cert to use if `authType` is set to `tls` |
```
keitwb: This is highly misleading -- there is no `authType` setting on this monitor.

asuresh4: Yes, that was not intentional. Removed.

docs/monitors/gitlab-gitaly.md
```diff
@@ -36,6 +36,9 @@ Configuration](../monitor-config.md#common-configuration).**
 | `password` | no | `string` | Basic Auth password to use on each request, if any. |
 | `useHTTPS` | no | `bool` | If true, the agent will connect to the exporter using HTTPS instead of plain HTTP. (**default:** `false`) |
 | `skipVerify` | no | `bool` | If useHTTPS is true and this option is also true, the exporter's TLS cert will not be verified. (**default:** `false`) |
+| `caCertPath` | no | `string` | Path to the CA cert that has signed the TLS cert, unnecessary if `skipVerify` is set to false. |
+| `clientCertPath` | no | `string` | Path to the client TLS cert to use for TLS required connections |
```
keitwb: Ok, so for these monitors if the user sets these options and is also using HTTPS, then these configured client certs will be used automatically?

internal/monitors/etcd/metadata.yaml
```diff
@@ -0,0 +1,867 @@
+monitors:
+- monitorType: etcd
+  dimensions:
+  doc: |-
+    This monitor reports etcd server metrics under the `/metrics` path on its
```
keitwb: Discuss the difference between this and `collectd/etcd` and also give a sample config for a typical K8s env.

docs/monitors/gitlab-gitaly.md
asuresh4: Yes, if the certs are provided, it will be used by the http client in prometheus exporter

internal/monitors/etcd/metadata.yaml
asuresh4: Added a comment regarding the differences between the monitors and also provided a sample config

PR #1055, opened 2019-10-29
Add CoreDNS prometheus monitor

mstumpfx:


keitwb:
> `check_links` looks in master for the coredns doc, which doesn't exist yet in master. That is the only link that's failing though, so probably safe to merge without it.

Yeah that's fine.  That's why that test isn't required but just make sure there aren't any other link failures.

mstumpfx:
There was a timeout on `https://wiki.jenkins.io/display/JENKINS/Metrics+Plugi` but it's resolved since.

PR #1058, opened 2019-10-30
Add kubernetesCluster helm config option

jaysfx:
A user wanted to be able to use a different value for kubernetes_cluster than
that of cluster.

Also cleanup required values. Removed conditional from daemonset and configmap
because it would confusingly result in half the helm chart being deployed.

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/templates/secrets.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/secrets.yaml
@@ -10,4 +10,4 @@ metadata:
     release: {{ .Release.Name }}
 type: Opaque
 data:
-  access-token: {{ .Values.signalFxAccessToken | b64enc | quote }}
+  access-token: {{ (required ".Values.signalFxAccessToken is required. If you need an access token visit https://www.signalfx.com for a free trial." .Values.signalFxAccessToken) | b64enc | quote }}
--- a/deployments/k8s/helm/signalfx-agent/templates/secrets.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/secrets.yaml
@@ -10,4 +10,4 @@ metadata:
     release: {{ .Release.Name }}
 type: Opaque
 data:
-  access-token: {{ (required ".Values.signalFxAccessToken is required. If you need an access token visit https://www.signalfx.com for a free trial." .Values.signalFxAccessToken) | b64enc | quote }}
+  access-token: {{ (required ".Values.signalFxAccessToken is required. If you need an access token, check your profile page in the SignalFx web application, or visit https://www.signalfx.com for a free trial." .Values.signalFxAccessToken) | b64enc | quote }}
--- a/deployments/k8s/helm/signalfx-agent/values.yaml
+++ b/deployments/k8s/helm/signalfx-agent/values.yaml
@@ -2,7 +2,7 @@
 # docker image tag if not overridden with imageTag
 agentVersion: 4.13.0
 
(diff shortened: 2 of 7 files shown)


deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -83,13 +83,17 @@ traceEndpointUrl:
...
-# clusterName must be provided.  It is an arbitrary value that identifies this
-# K8s cluster in SignalFx.  This will be the value of the 'kubernetes_cluster'
-# dimension on every metric sent by the agent.  It will also be the value of
-# the `cluster` config option that is used to set a `cluster` property on the
-# `kubernetes_node` dimension.
+# An arbitrary value that identifies this K8s cluster in SignalFx.  This will be
+# the value of the 'kubernetes_cluster' dimension on every metric sent by the
+# agent (unless overriden by `kubernetesClusterName`).  It will also be the
+# value of the `cluster` config option that is used to set a `cluster` property
+# on the `kubernetes_node` dimension. (REQUIRED)
```
mstumpfx: This is a good explanation of _how_ this works but doesn't really explain _why_ one would need this. I think this is going to be confusing for users who are unfamiliar with the need for the `cluster` property.
We should add something about setting this value to the same value as your Smart Gateway (or gateway cluster?).
Better to be explicit in this case I think.

deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -83,13 +83,17 @@ traceEndpointUrl:
...
 apiUrl:
 
-# clusterName must be provided.  It is an arbitrary value that identifies this
-# K8s cluster in SignalFx.  This will be the value of the 'kubernetes_cluster'
-# dimension on every metric sent by the agent.  It will also be the value of
-# the `cluster` config option that is used to set a `cluster` property on the
-# `kubernetes_node` dimension.
+# An arbitrary value that identifies this K8s cluster in SignalFx.  This will be
+# the value of the 'kubernetes_cluster' dimension on every metric sent by the
+# agent (unless overriden by `kubernetesClusterName`).  It will also be the
```
keitwb: The actual config value is `kubernetesCluster`, not `kubernetesClusterName`.

deployments/k8s/helm/signalfx-agent/values.yaml
keitwb: Yeah should probably mention the Smart Gateway aspect.

deployments/k8s/helm/signalfx-agent/templates/secrets.yaml
```diff
@@ -10,4 +10,4 @@ metadata:
     release: {{ .Release.Name }}
 type: Opaque
 data:
-  access-token: {{ .Values.signalFxAccessToken | b64enc | quote }}
+  access-token: {{ (required ".Values.signalFxAccessToken is required. If you need an access token visit https://www.signalfx.com for a free trial." .Values.signalFxAccessToken) | b64enc | quote }}
```
mstumpfx: If you need an access token, check your profile page in the SignalFx web application, or visit....

jaysfx:
```suggestion
  access-token: {{ (required ".Values.signalFxAccessToken is required. If you need an access token, check your profile page in the SignalFx web application, or visit https://www.signalfx.com for a free trial." .Values.signalFxAccessToken) | b64enc | quote }}
```

PR #1061, opened 2019-10-31
Generalize container id parsing

asuresh4:


internal/utils/k8sutil/k8s.go
```diff
@@ -82,5 +82,6 @@ func PortByNumber(pod *v1.Pod, port int32) *v1.ContainerPort {
 
 // StripContainerID returns a pure container id without the runtime scheme://
 func StripContainerID(id string) string {
-	return strings.Replace(strings.Replace(id, "docker://", "", 1), "cri-o://", "", 1)
+	re := regexp.MustCompile(`^[\w_-]+://`)
```
keitwb: Move the `regexp.MustCompile` var to a global variable and reuse it in this function.  That will increase the efficiency of this func by many times.

asuresh4: Done

tests/helpers/util.py
```diff
@@ -380,6 +380,10 @@ def get_container_file_content(container, path):
     return container.exec_run("cat %s" % path)[1].decode("utf-8")
 
 
+def get_stripped_container_id(container_id):
+    return re.sub(r"^[\w_-]+://", "", container_id)
```
keitwb: This is a test anti-pattern by implementing the test logic identical to the app logic.  If you happened to have a mistake in the regexp, you would never know it.  Make this something more explicit with `.replace` like it was originally (but still keep the separate function so we can add more easily later).

asuresh4: Ah good point, will update

PR #1067, opened 2019-11-04
Skip dim updates if container_id is empty

asuresh4:


internal/monitors/kubernetes/cluster/monitor_test.go
```diff
@@ -183,6 +192,17 @@ var _ = Describe("Kubernetes plugin", func() {
 			MergeIntoExisting: true,
 		}))
 
+		Expect(dims).ShouldNot(ConsistOf(&types.Dimension{
```
keitwb: What is this test for?

keitwb: Also it is redundant to the previous assertion.

keitwb: ConsistOf will test the length as well as the elements matching.

asuresh4: Oh ok. I wasn't sure if we also needed to assert that a particular Dimension should not exist in the output.

asuresh4: Removed the additional check

PR #1075, opened 2019-11-08
Mark openshift metrics as non-default

mstumpfx:
None of these metrics are used in built-in content so this violates our "rule" of default vs non-default metrics. These metrics have been categorized as custom on the backend already, so this is just a doc fix.


asuresh4:
Closing in favor of #1077 - cherry picked this commit.

PR #1079, opened 2019-11-12
Added missing HTTP paths for Zipkin listener

owais:
Added the following paths to Zipkin listener in addition to /v1/trace

- /api/v1/spans
- /api/v2/spans

Depends on: https://github.com/signalfx/gateway/pull/268


owais:
@keitwb Yes, I opened this so reviewers for the gateway change can see where the new function is needed.

keitwb:
Looks good.  Merge at will.

Issue #1080, opened 2019-11-13
python-monitor should log unhandled exceptions


xp-1000:
Hello,

In my current environment (debian9), signalfx agent logging works fine through systemd.

Nevertheless, trying to create a new python-monitor script I noticed that if errors happen during script execution by the agent, this is not retrieved in agent logs.

For example, a python stacktrace due to unhandled exception will not be visible anywhere (in my knowledge) and so the script will fail silently.

This would be far better to provider logs from python script to signalfx agent logs

jaysfx:
Thanks for the report. The fix has been merged to master.

xp-1000:
awesome, thanks for this quick action.
I will test for the next released version.

PR #1082, opened 2019-11-13
Fix python-monitor error logging

jaysfx:
Logging messages at error level was a noop and exceptions were logged as errors
so those were never logged.

Issue #1080


mstumpfx:
I wonder why we left it out in the first place?

keitwb:
> I wonder why we left it out in the first place?

I probably thought it would fall through like in other languages but apparently it doesn't work that way in Go.

PR #1083, opened 2019-11-14
Use go:generate consistently

keitwb:
This makes the agent buildable across all platforms with standard Go
commands instead of requiring custom make targets.  This also makes
working with the agent in an IDE easier.


cmd/monitorcodegen/main.go
```diff
@@ -32,35 +32,7 @@ func buildOutputPath(pkg *selfdescribe.PackageMetadata) string {
 	return path.Join(outputDir, genMetadata)
 }
 
-// shouldRegenerate determines whether the metadata file needs regenerated based on its existence and timestamps.
-func shouldRegenerate(pkg *selfdescribe.PackageMetadata, outputFile string) (bool, error) {
```
jaysfx: Does this not result in a speedup with the new method?

internal/monitors/zcodegen/codegen.go
```diff
@@ -0,0 +1,6 @@
+// Package codegen exists purely to manipulate the go:generate build order so
+// that collectd templates are generated first before the monitor metadata
+// module is generated.
+package codegen
+
+//go:generate bash -c "go build -o ./monitorcodegen ../../../cmd/monitorcodegen && ./monitorcodegen"
```
jaysfx: This will require having bash installed on Windows right?

keitwb: Yeah I missed switching the windows script to use go generate.  Will fix.

keitwb: Ok, fixed it to work with windows and to make windows CI actually use go generate.

cmd/monitorcodegen/main.go
keitwb: The speed of `go generate ./...` is sub-1 second after the first invocation, so I'm not too worried about the speed.

PR #1084, opened 2019-11-15
881 update agent instructions

galpizarsfx:
@barbara-sfx Please review my proposed changes to the Smart Agent tile.


docs/quick-install.md
```diff
@@ -1,117 +1,90 @@
...
+## Review pre-installation requirements for the Smart Agent
 
-* _Monitors_ that collect metrics from the host and applications
+Before you download and install the Smart Agent on a **single** host, review the requirements below.
 
-* _Observers_ that discover applications and services running on the host
+(For other installation options, including bulk deployments, see [Advanced Installation Options](./advanced-install-options.md).)
 
-* a _Writer_ that sends the metrics collected by monitors to SignalFx
+| General requirements   |     Linux requirements      |  Windows requirements |
```
barbara-sfx: I don't think a table format works here, because it's column 1 for everyone, then either column 2 or 3, which is a weird way for a table to work. 

I think 3 subheadings would work better.

docs/quick-install.md
```diff
@@ -1,117 +1,90 @@
...
+|----------|:-------------:|------:|
+| <p>You must have access to your command line interface.</p> <p>You must uninstall or disable any previously installed collector agent from your host, such as collectd.</p>| <p>You must run kernel version 2.6 or higher for your Linux distribution.</p> <p>The Smart Agent is bundled with additional required dependencies, including a Java JRE runtime and a Python runtime. As a result, there is no need to proactively install additional dependencies.</p>| <p>You must run .Net Framework 3.5 on Windows 8 or higher.</p> <p>You must run Visual C++ Compiler for Python 2.7.</p>  |
```
barbara-sfx: Why is this here if it's not something they need to do? >The Smart Agent is bundled with additional required dependencies, including a Java JRE runtime and a Python runtime. As a result, there is no need to proactively install additional dependencies.</p>

docs/quick-install.md
```diff
@@ -1,117 +1,90 @@
...
-instance of applications that match a discovery rule. See [Endpoint
-Discovery](./auto-discovery.md) for more information.
+### Step 1. Install the SignalFx Smart Agent on Single Host
 
-Many of the monitors are built around [collectd](https://collectd.org), an open
-source third-party monitor, and use it to collect metrics. Some other monitors
-do not use collectd. However, either type is configured in the same way.
+<details>
+<summary>Linux</summary>
+<br>
```
barbara-sfx: I like this Summary thing! Never saw it before.

docs/quick-install.md
```diff
@@ -1,117 +1,90 @@
...
-### Monitors
+## Install the Smart Agent
 
-Monitors collect metrics from the host system and services.  They are
-configured under the `monitors` list in the agent config.  For
-application specific monitors, you can define discovery rules in your monitor
-configuration. A separate monitor instance is created for each discovered
-instance of applications that match a discovery rule. See [Endpoint
-Discovery](./auto-discovery.md) for more information.
+### Step 1. Install the SignalFx Smart Agent on Single Host
```
barbara-sfx: ... on your host (no caps)

Do we need to repeat " a single host" again?

docs/quick-install.md
barbara-sfx: But all I see when I expand Linus is the sh statement. So I'm guessing headings aren't supported. So you need to rework this and review it live, and then I'll look again. 

Also I'm not sure "Optional 1" and "Optional 2" work as headings, even if they were visible. Maybe use Summary for those too?

galpizarsfx:
@barbara-sfx I think I found a potential miscommunication taking place here, which may clear up some of your earlier comments:

- This request is to review the **881-update-agent-instructions** branch within the **agent** repo. 
- The readthedocs site you refer to in your comments is related to another branch, called **881-update-agent** that lives within the product repo. That branch hasn't been updated in awhile, so it's probably best to ignore it, and simply focus on the branch within the agent repo. My apologies for any confusion.

docs/quick-install.md
barbara-sfx: This content still isn't showing up in the build. Maybe we should discuss this PR in our 1x1 tomorrow.

image

docs/quick-install.md
galpizarsfx: This has been updated.

docs/quick-install.md
galpizarsfx: That's a fair point. This has been removed.

docs/quick-install.md
galpizarsfx: Hmm bizarre; I see it working fine within GitHub. 
Screen Shot 2019-12-03 at 7 08 27 PM

docs/quick-install.md
galpizarsfx: On one hand, we can follow the tips from DTTW, which was to make headers and titles as specific as possible. But, earlier in the document we do say, "single host," so using here "your host" should be clear enough that you are not going to install the agent on multiple hosts.

docs/quick-install.md
```diff
@@ -1,117 +1,90 @@
...
 
- - [Concepts](#concepts)
- - [Installation](#installation)
+The Smart Agent contains three main components:
 
+| Component | Description |
+|-----------|-------------|
+| Monitors  |  This component collects metrics from the host and applications. For a list of supported monitors and their configurations, see [Monitor Configuration](./monitor-config.md).            |
+| Observers |   This component collects metrics from services that are running in your environment. For a list of supported observers and their configurations, see [Observer Configuration](./observer-config.md).           |
+| Writer    |   This component collects metrics from configured monitors and then sends these metrics to SignalFx on a regular basis. If you are expecting your monitors to send large volumes of metrics through a single agent, then you must update the configurations. To learn more, see [Agent Configurations](./config-schema.md#writer).          |
```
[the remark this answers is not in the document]
barbara-sfx: Turns out the script fixes these, so never mind going to html instead. But they should open in a new window.

docs/quick-install.md
```diff
@@ -1,175 +1,184 @@
...
+***
 
-To install the Smart Agent on a single Windows host, enter:
+### Step 2. Confirm your installation
 
-```sh
-& {Set-ExecutionPolicy Bypass -Scope Process -Force; $script = ((New-Object System.Net.WebClient).DownloadString('https://dl.signalfx.com/signalfx-agent.ps1')); $params = @{access_token = "YOUR_SIGNALFX_API_TOKEN"};; 
-ingest_url = "https://ingest.YOUR_SIGNALFX_REALM.signalfx.com"; api_url = "https://api.YOUR_SIGNALFX_REALM.signalfx.com"}; Invoke-Command -ScriptBlock ([scriptblock]::Create(". {$script} $(&{$args} @params)"))}
-```
+1. From the command line, enter the following command to confirm that your Smart Agent is functional:
```
barbara-sfx: I thought this was Linux only at first, but the live doc says this command works on both platforms. (Because I didn't know Windows supported sudo.) For clarity, I'd say "From the Linux or Windows command line,...."

docs/quick-install.md
```diff
@@ -1,175 +1,184 @@
...
+    Discovered Endpoint Count:        6
+    Bad Monitor Config:               None
+    Global Dimensions:                {host: my-host-1}
+    Datapoints sent (last minute):    1614
+    Events Sent (last minute):        0
+    Trace Spans Sent (last minute):   0
+    ```
 
-To confirm the SignalFx Smart Agent installation is functional on either platform, enter:
+2. Enter the following commands to verify the installation:
```
barbara-sfx: How is "confirm that your Smart Agent is functional" in step 1 different from "verify the installation" in step 2?

docs/quick-install.md
barbara-sfx: But also, my standard is to put what you are trying to do first, so it would be "To confirm..., enter the following command on the Linux or Windows command line."

docs/quick-install.md
```diff
@@ -1,175 +1,184 @@
...
 
-```sh
-signalfx-agent status config - show resolved config in use by agent
-signalfx-agent status endpoints - show discovered endpoints
-signalfx-agent status monitors - show active monitors
-signalfx-agent status all - show everything
-```
+***
+ 
+#### Update your realm
```
barbara-sfx: If the install from the UI, they should never have to do anything with realms, and even if they did, these instructions don't really tell them what to fix or how to fix it. Ditto for those endpoints. 

In fact, this is why we are now people to install from the UI -- so they don't have to deal with any of this stuff.

docs/quick-install.md
```diff
@@ -1,175 +1,184 @@
...
+2. Click **My Profile**.
+3. Next to **Organizations**, review the listed realm.
 
-##### Realm 
+***
 
-By default, the Smart Agent will send data to the us0 realm. If you are not in this realm, you will need to explicitly set the signalFxRealm option in your config like this:
+#### Set the endpoints
 
+To explicitly set the ingest, API server, and trace endpoint URLs, review the following examples:
```
barbara-sfx: Same comment as above

docs/quick-install.md
```diff
@@ -1,175 +1,184 @@
...
 
 ```sh
-ingestUrl: "https://ingest.YOUR_SIGNALFX_REALM.signalfx.com"
-apiUrl: "https://api.YOUR_SIGNALFX_REALM.signalfx.com"
-traceEndpointUrl: "https://ingest.YOUR_SIGNALFX_REALM.signalfx.com/v1/trace"
+journalctl -u signalfx-agent | tail -100
 ```
 
-This will default to the endpoints for the realm configured in signalFxRealm if not set.
+For Windows, simply review the event logs.
```
barbara-sfx: Remove "simply"

docs/quick-install.md
```diff
@@ -1,175 +1,184 @@
...
+### Review additional documentation
 
-Installation is complete.
+After a successful installation, you can learn more about:
 
-To continue your exploration of SignalFx Smart Agent capabilities, see [Advanced Installation Options](./advanced-install-options.md).
+* The capabilities of the SignalFx Smart Agent. See [Advanced Installation Options](./advanced-install-options.md).
 
-To learn more about how your data is presented in SignalFx, see the [15-Minute SignalFx Quick Start](https://docs.signalfx.com/en/latest/getting-started/quick-start.html).
+* The SignalFx UI, including how data is displayed. See [15-Minute SignalFx Quick Start](https://docs.signalfx.com/en/latest/getting-started/quick-start.html).
```
barbara-sfx: I'm not a fan of bullets that start with sentence fragments. So I don't really see this as an improvement on the old text, except that adding bullets would be nice.

For the quick start, you should take them right to the target (https://docs.signalfx.com/en/latest/getting-started/quick-start.html#step-3-view-infrastructure-status). Because it dumps them at  Step 3, so it might be good to say something like:

To learn more about how your data is presented in SignalFx, start with [this section](...html#target) in the **15-Minute SignalFx Quick Start**.

docs/quick-install.md
galpizarsfx: I guess they are both trying to achieve the same purpose, which is to validate / confirm / verify that the installation was successful. 

I updated both steps to say, "To confirm your installation, enter the following command on the Linux or Windows command line:..."

docs/quick-install.md
galpizarsfx: The doc site currently says, "By default, the Smart Agent will send data to the us0 realm. If you are not in this realm, you will need to explicitly set the signalFxRealm option in your config like this:..." 

This statement is confusing for a number of reasons, including your previous comment about installing from the UI where the realm is already populated. Does the above statement imply that if the pre-populated command in the UI says us0, then you need to confirm elsewhere that is indeed your realm, and not simply a default / placeholder realm? Similarly, to me, the statement implies that you can possibly run the install command without specifying a realm because, by default, the installation will be configured to the us0 realm.

Long story short, I think maybe it's best to remove this section entirely. What do you think?

docs/quick-install.md
galpizarsfx: You're right. It's not overly clear how configuring this information will help the installation process, let alone how to configure this information. Maybe we should consider removing it completely?

docs/quick-install.md
galpizarsfx: It's a very short section, so there wasn't much to improve.; at least the content flows better now, in my opinion. I'll update the link, and turn the sentence fragments into a complete sentence.

docs/quick-install.md
barbara-sfx: Yes, everything related to realm should go. This was here because we provided template for them to enter at the command line themselves. Now we don't so this should not matter to them.

docs/quick-install.md
barbara-sfx: Yep, this also goes.

docs/quick-install.md
galpizarsfx: The "update your realm" section has been removed.

docs/quick-install.md
galpizarsfx: This section has been removed.

galpizarsfx:
@asuresh4 @keitwb I did some fairly major re-work to smart agent tile. As you can see above, there are some conflicts. Am I responsible for resolving those conflicts?

PR #1085, opened 2019-11-18
Add support for helm v3

jchengsfx:
These changes will require upgrading any locally installed helm to v3 in order to run `generate-from-helm` outside of the dev-image.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,8 +25,7 @@ executors:
 
   helm:
     docker:
-      - image: alpine/helm:2.15.0
-
+      - image: alpine/helm:3.0.0
 
 commands:
   goexecutor_setup:
@@ -175,7 +174,7 @@ commands:
     parameters:
       version:
         type: string
-        default: v2.15.0
+        default: v3.0.0
     steps:
       - download_and_cache:
           key: v4-helm-binary-<< parameters.version >>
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -185,8 +184,10 @@ commands:
           tar -zxvf /tmp/helm.tar.gz -C /tmp
           sudo mv -f /tmp/linux-amd64/helm /usr/bin/helm
           chmod a+x /usr/bin/helm
-          sudo mv -f /tmp/linux-amd64/tiller /usr/bin/tiller
-          chmod a+x /usr/bin/tiller
+          if [[ -f /tmp/linux-amd64/tiller ]]; then
```
keitwb: Why are we installing tiller if it was removed from Helm 3?

keitwb: Is this so that we can run both Helm 2 and 3 tests?

tests/deployments/helm/helm_test.py
```diff
@@ -192,40 +209,60 @@ def install_helm_chart(k8s_cluster, values_path):
...
+    with run_service("helm", **opts) as cont:
+        get_kubeconfig(k8s_cluster, cont)
+        yield cont
+
+
+@pytest.mark.parametrize("helm_version", ["2.15.0", "3.0.0"])
+def test_helm(k8s_cluster, helm_version):
+    helm_major_version = int(helm_version.split(".")[0])
+    with run_helm_image(k8s_cluster, helm_version) as cont:
+        with k8s_cluster.create_resources([APP_YAML_PATH]), tiller_rbac_resources(
```
keitwb: If it is 3.0, we really shouldn't be installing anything related to Tiller if it was removed.

.circleci/config.yml
jchengsfx: > Is this so that we can run both Helm 2 and 3 tests?

I left it in just for posterity, but it's actually not being used by pytest anymore since we're running helm from docker images now in the test.

tests/deployments/helm/helm_test.py
jchengsfx: > If it is 3.0, we really shouldn't be installing anything related to Tiller if it was removed.

I updated the tests for both helm v2 and v3.

.circleci/config.yml
keitwb: Ok then get rid of it.  Git is our posterity.

PR #1086, opened 2019-11-18
Use new writer from signalfx-go for datapoints and spans

keitwb:
It is mostly a port from the agent, abstracted out to support different
types of instances.  It uses a ring buffer to prioritize newer elements
while keeping memory consumption bounded.

Also changed the monitor output interface to send datapoints and spans
in groups (as a slice) instead of individually to improve efficiency in
batching.  This should result in larger batches which will improve agent
efficiency.

Also increasing the default datapoint buffer size to 25,000 from 5,000.
The memory cost of this is 20,000 * 8 bytes.


keitwb:
Still hasn't been merged to master in signalfx-go.  Will update this when it is.

PR #1090, opened 2019-11-22
Allow Helm chart to use existing secret containing SFX token

fiftoine:


deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
```diff
@@ -48,3 +48,10 @@ Get namespace to deploy agent and its dependencies.
 {{- define "signalfx-agent.namespace" -}}
     {{- default .Release.Namespace .Values.namespace -}}
 {{- end -}}
+
+{{/*
+Get namespace to deploy agent and its dependencies.
```
keitwb: Copy/pasted without change.

deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -2,9 +2,14 @@
...
 agentVersion: 4.14.1
 
-# The access token for SignalFx. (REQUIRED)
+# The access token for SignalFx. (REQUIRED if signalFxAccessTokenExistingSecret not set)
 signalFxAccessToken: ""
 
+# Use an existing secret for the SFX token (if set does not create a secret)
+# See templates/secrets.yaml for the format of the secret to use
+# If set, signalFxAccessToken is not required
+signalFxAccessTokenExistingSecret: ""
```
keitwb: Would prefer this called `signalFxAccessTokenSecretName` instead so that it is clear that is is the _name_ of the secret resource.

keitwb:
Thanks @fiftoine for this.  We will release the updated Helm chart onto our repo sometime today.

PR #1091, opened 2019-11-22
Rerun pytest failures in circleci

jchengsfx:
Rerun all pytest failures in circleci to help reduce job failures due to environmental or file descriptor issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,9 @@ commands:
       markers:
         type: string
         default: ""
+      workers:
+        type: string
+        default: auto
       pytest_options:
         type: string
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -340,7 +340,7 @@ jobs:
       - import_image
       - run: .circleci/scripts/setup-integration-tests.sh
       - run_pytest:
-          pytest_options: -n auto
+          pytest_options: -n1
```
keitwb: I suspect it might be more of an issue when there is more than one executor.

keitwb:
Closing in favor of #1109

PR #1095, opened 2019-11-27
Make host bind mount propagated to container in K8s

keitwb:


mstumpfx:
@keitwb lgtm but why ?

keitwb:
> @keitwb lgtm but why ?

To pick up new mounts on the host filesystem for the filesystem utilization monitors to watch after the agent has already started.

PR #1098, opened 2019-12-02
Remove bad link from genericjmx doc

keitwb:


jaysfx:
Could use wayback machine link if the content is still desired: https://web.archive.org/web/20170313145920/http://realjenius.com/2012/11/21/java7-jmx-tunneling-freedom/

keitwb:
> Could use wayback machine link if the content is still desired: https://web.archive.org/web/20170313145920/http://realjenius.com/2012/11/21/java7-jmx-tunneling-freedom/

Nope, not really that useful after rereading it.

PR #1102, opened 2019-12-03
Add ansible integration tests

jchengsfx:


tests/deployments/ansible/images/Dockerfile.centos6
```diff
@@ -0,0 +1,17 @@
+FROM centos:6
```
keitwb: Drop centos6 and Amazon Linux 1 since we are already planning on dropping support for them.  No use wasting CI run time on them.

.circleci/scripts/run-deployment-tests.sh
```diff
@@ -68,6 +68,11 @@ ansible)
     docker run --rm \
         signalfx-agent-ansible-dev \
         ansible-lint .
+    if [ $CIRCLE_NODE_INDEX -eq 0 ]; then
```
keitwb: Does this only run it when changes are made to the playbook dir?

jchengsfx: Will only run for PRs if there are changes to the playbook dir or the ansible test dir (same as chef and puppet).

PR #1104, opened 2019-12-04
Regex for json keys in metric value path definition in expvar monitor

bjsignalfx:
- We want to allow the use of regex for json keys when defining the path of metric values in a json object.
- Changed path separator to be configurable.
- Changed to set the Name field for MetricConfig during configuration to avoid processing strings constantly using when using name().


docs/monitors/expvar.md
```diff
@@ -190,6 +190,7 @@ The **nested** `metrics` config object has the following fields:
 | `JSONPath` | **yes** | `string` | JSON path of the metric value |
 | `type` | **yes** | `string` | SignalFx metric type. Possible values are "gauge" or "cumulative" |
 | `dimensions` | no | `list of objects (see below)` | Metric dimensions |
+| `pathSeparator` | no | `string` | Metric value path separator character in JSON. The default character is / (**default:** `/`) |
```
keitwb: No need to say "The default character is /" because the doc already will this for you.

internal/monitors/expvar/config.go
```diff
@@ -57,6 +56,24 @@ type MetricConfig struct {
 	Type string `yaml:"type" validate:"required,oneof=gauge cumulative"`
 	// Metric dimensions
 	DimensionConfigs []DimensionConfig `yaml:"dimensions"`
+	// Metric value path separator character in JSON. The default character is /
+	PathSeparator string `yaml:"pathSeparator" default:"/"`
+	// The PathSeparator string character as rune
+	pathSeparator rune
```
keitwb: Do not treat config objects as scratch pads for caching data or deriving values. This makes things like config comparison when determining which montiors to restart more complicated.

docs/monitors/expvar.md
```diff
@@ -65,7 +65,7 @@ monitors:
     metric_source: expvar-aws
 ```
 The expvar monitor can be configured to extract metric values from complex JSON objects such as the one shown
-below. Suppose the `memstats` variable shown below is exposed at endpoint `http://172.17.0.5:5000/debug/vars`
+below. Suppose the `memstats` variable shown below is exposed at endpoint `http://172.17.0.5:5000/debug/vars`.
```
keitwb: Should not have period at end.

internal/monitors/expvar/config.go
```diff
@@ -57,6 +58,8 @@ type MetricConfig struct {
 	Type string `yaml:"type" validate:"required,oneof=gauge cumulative"`
 	// Metric dimensions
 	DimensionConfigs []DimensionConfig `yaml:"dimensions"`
+	// Path separator character of metric value in JSON object
+	PathSeparator string `yaml:"pathSeparator" default:"/"`
```
keitwb: Is this default fully backwards compatible with what this monitor did before?

keitwb: It kind of seems like the old default was `.`.

tests/monitors/expvar/expvar_test.py
```diff
@@ -62,6 +62,7 @@ def test_expvar_custom_metric():
               port: 8080
               metrics:
               - JSONPath: queues.count
+                pathSeparator: '.'
```
keitwb: If you have to add this to make this integration test work, then it means it isn't backwards compatible and you will break existing deployments.

internal/monitors/expvar/config.go
bjsignalfx: @keitwb Ok I'm made the old default character the default. But I have noticed that the default path separator character is not set if you configure the embedded MetricConfig and don't provide a value for pathSeparator!

tests/monitors/expvar/expvar_test.py
bjsignalfx: The character '.' is now the default as before, but somehow the path separator is not set if you don't explicitly configure it in your metrics section. If you don't have a metrics section all together in your config then the default path separator is set!

internal/monitors/expvar/config.go
```diff
@@ -32,7 +33,7 @@ type Config struct {
 	// If true, sends metrics memstats.alloc, memstats.by_size.size, memstats.by_size.mallocs and memstats.by_size.frees
 	EnhancedMetrics bool `yaml:"enhancedMetrics"`
 	// Metrics configurations
-	MetricConfigs []MetricConfig `yaml:"metrics"`
+	MetricConfigs []*MetricConfig `yaml:"metrics"`
```
keitwb:
```suggestion
	MetricConfigs []*MetricConfig `yaml:"metrics" default:"[]"`
```

tests/monitors/expvar/expvar_test.py
bjsignalfx: @keitwb I implemented a temp fix for this by setting the default path separator value manually.

PR #1107, opened 2019-12-05
Make prometheus-exporter timeout user-configurable

jaysfx:
Keeps original default of 10s while allowing user to override.

### DIFF
--- a/internal/monitors/prometheusexporter/prometheus.go
+++ b/internal/monitors/prometheusexporter/prometheus.go
@@ -56,7 +56,8 @@ type Config struct {
 	// Path to the client TLS key to use for TLS required connections
 	ClientKeyPath string `yaml:"clientKeyPath"`
 
-	// Number of seconds before timing out while requesting data from target host.
+	// HTTP timeout duration for both read and writes. This should be a
+	// duration string that is accepted by https://golang.org/pkg/time/#ParseDuration
 	HTTPTimeout time.Duration `yaml:"httpTimeout" default:"10s"`
 
 	// Use pod service account to authenticate.
--- a/internal/monitors/prometheusexporter/prometheus.go
+++ b/internal/monitors/prometheusexporter/prometheus.go
@@ -56,6 +56,9 @@ type Config struct {
 	// Path to the client TLS key to use for TLS required connections
 	ClientKeyPath string `yaml:"clientKeyPath"`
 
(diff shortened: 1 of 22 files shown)


internal/monitors/prometheusexporter/prometheus.go
```diff
@@ -56,6 +56,9 @@ type Config struct {
 	// Path to the client TLS key to use for TLS required connections
 	ClientKeyPath string `yaml:"clientKeyPath"`
 
+	// Number of seconds before timing out while requesting data from target host.
+	HTTPTimeout time.Duration `yaml:"httpTimeout" default:"10"`
```
keitwb: time.Duration has units of nanoseconds not seconds.

You can do a default of `10s` and I think it will parse it from yaml properly (see other config options that are time.Duration).

internal/monitors/prometheusexporter/prometheus.go
```diff
@@ -124,7 +127,7 @@ func (m *Monitor) Configure(conf *Config) error {
 	}
 
 	client := &http.Client{
-		Timeout: 10 * time.Second,
+		Timeout: conf.HTTPTimeout * time.Second,
```
keitwb: If you're going to convert it to seconds, then call the config option `httpTimeoutSeconds` and make it an int instead of fudging the meaning of time.Duration.

internal/monitors/prometheusexporter/prometheus.go
jaysfx: Ah yeah I was later multiplying it by time.Second so it was wrong.. but worked. :P Will update.

internal/monitors/prometheusexporter/prometheus.go
jaysfx: Yeah will just use time.Duration properly like other options, didn't know default groked time units.

internal/monitors/prometheusexporter/prometheus.go
```diff
@@ -56,6 +56,9 @@ type Config struct {
 	// Path to the client TLS key to use for TLS required connections
 	ClientKeyPath string `yaml:"clientKeyPath"`
 
+	// Number of seconds before timing out while requesting data from target host.
```
keitwb: A better doc for the config option is something like this: https://github.com/signalfx/signalfx-agent/blob/master/internal/monitors/forwarder/monitor.go#L30

PR #1109, opened 2019-12-10
Attempt to fix EBADF errors in pytest

keitwb:
- Upgrade Python to latest
 - Upgrade pytest and pytest-xdist
 - Don't close file descriptors for stdin/out/err when running agent as a subprocess
 - Write the config file in binary mode
 - Add decorator to catch and retry EBADF for certain functions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,15 +44,15 @@ commands:
             docker load -i /tmp/workspace/image.tar
 
   install_pytest:
-    description: Install pytest and dependencies with pyenv 3.6.3.
+    description: Install pytest and dependencies with pyenv 3.8.0.
     steps:
       - restore_cache:
           key: v1-pytest-{{ checksum "tests/requirements.txt" }}
       - run:
           name: Install pytest and dependencies
           command: |
-            pyenv install --skip-existing 3.6.3
-            pyenv global 3.6.3
+            pyenv install --skip-existing 3.8.0
+            pyenv global 3.8.0
             if which pip; then
                 pip install --upgrade 'pip==10.0.1'
             else
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,15 +44,20 @@ commands:
             docker load -i /tmp/workspace/image.tar
 
   install_pytest:
-    description: Install pytest and dependencies with pyenv 3.8.0.
+    description: Install pytest and dependencies with pyenv.
+    parameters:
+      python_version:
+        type: string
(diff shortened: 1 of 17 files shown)


.circleci/config.yml
```diff
@@ -544,7 +545,7 @@ workflows:
             - helm_check
       - k8s_integration_tests:
           name: k8s_crio_integration_tests
-          k8s_version: v1.15.0
+          k8s_version: v1.16.0
```
keitwb: Do this in a separate PR, let's not mix things up here.

.circleci/config.yml
```diff
@@ -368,6 +368,7 @@ jobs:
       - run:
           environment:
             K8S_VERSION: << parameters.k8s_version >>
+            WITH_CRIO: "<< parameters.with_crio >>"
```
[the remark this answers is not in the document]
jchengsfx: > Split to separate PR

Yep, I just needed something quick to trigger all k8s tests for now.

.circleci/scripts/collect_tests.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def pytest_collection_finish(self, session):
+        for item in session.items:
+            self.collected.add(item.location[0])
+
+
+collect_tests = CollectTests()
+markers = sys.argv[1]
+directory = sys.argv[2]
+pytest.main(['--collect-only', '-m', markers, '-p', 'no:terminal', directory], plugins=[collect_tests])
```
keitwb:
```suggestion
pytest.main(['--collect-only', '-m', markers, '-p', 'no:terminal', directory], plugins=[CollectTests()])
```

tests/helpers/util.py
```diff
@@ -156,11 +157,15 @@ def get_unique_localhost():
...
-    )
+    try:
+        proc = subprocess.Popen(
+            [str(c) for c in command], env=env, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs
+        )
+    except OSError as e:
+        sys.stderr.write(f"{e}\n")
 
     get_output = pull_from_reader_in_background(proc.stdout)
+    assert proc.poll() is None, f"agent is not running:\n{get_output}"
```
keitwb: `get_output` is a function, not a string value.

Also this is super racy.  If the agent dies immediately after this check it will still go on happily.

keitwb: Also `proc` will be None if the OSError happens since it will throw the exception before `proc` gets assigned.

PR #1111, opened 2019-12-10
Update installer test for circleci parallelism

jchengsfx:


tests/packaging/installer_test.py
```diff
@@ -67,9 +75,15 @@ def _run_tests(base_image, init_system, installer_args, **extra_run_kwargs):
             print_lines(get_agent_logs(cont, init_system))
 
 
-@pytest.mark.parametrize("base_image,init_system", SUPPORTED_DISTROS)
+@pytest.mark.parametrize(
+    "base_image,init_system",
+    [pytest.param(distro, init, marks=pytest.mark.deb) for distro, init in DEB_DISTROS]
```
keitwb: Maybe just make individual functions for each distro (e.g. `def test_installer_debian_8`) and then they can call a helper function that actually runs the tests with the appropriate init system specified.  These functions would have markers on them for deb/rpm. Then we can also get rid of the DEB/RPM_DISTROS global variables.

In general I find that much easier to understand than all of these function decorators with list comprehensions with `pytest.param` calls.

PR #1112, opened 2019-12-11
Fixed expvar monitor crashing for empty map value and metric name mangling for wildcard in JSON path

bjsignalfx:
Fixed the following issues with the expvar monitor reported by a customer:
1. Using a JSONPath with willplayads.* results in a single metric with the names smushed together instead of separate metrics.
2. Having a map value be `{}` causes the monitor to crash.

### DIFF
--- a/internal/monitors/expvar/expvar.go
+++ b/internal/monitors/expvar/expvar.go
@@ -32,8 +32,8 @@ const (
 func init() {
 	monitors.Register(&monitorMetadata, func() interface{} {
 		return &Monitor{
-			metricsKeys:      map[string][]string{},
-			dimsKeys:         map[string][]string{},
+			keysFromPaths:    map[string][]string{},
+			keysFromDimPaths: map[string][]string{},
 			allMetricConfigs: nil,
 		}
 	}, &Config{})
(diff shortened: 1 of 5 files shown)


keitwb:
@bjsignalfx Please change the title of the PR to use proper title case and amend your commit message in git to provide more details on what the problem was and why these changes are necessary, and generally what it does to solve it.

It looks like you are doing a lot more than just "check[ing] for non-numeric metric values".

internal/monitors/expvar/expvar.go
```diff
@@ -164,90 +164,113 @@ func (m *Monitor) findKeys(pattern string, aMap map[string]interface{}) []string
...
-	switch jsonValue := jsonValue.(type) {
+func (m *Monitor) addDps(dps map[string][]*datapoint.Datapoint, dp *datapoint.Datapoint, keys []string, keyFromPathIndex int, value interface{}, conf *MetricConfig) {
+	keysFromPath, pathKeysLen, nextKeyFromPathIndex := m.keysFromPaths[conf.Name], len(m.keysFromPaths[conf.Name]), keyFromPathIndex+1
+	atPathEnd := nextKeyFromPathIndex == pathKeysLen
+	switch value := value.(type) {
 	case map[string]interface{}:
-		nextJSONKeyPattern := jsonKeyPatterns[nextJSONKeyPatternIndex]
-		dpCopy, metric, dims := dp, dp.Metric, dp.Dimensions
-		for i, nextJSONKey := range m.findKeys(nextJSONKeyPattern, jsonValue) {
+		if atPathEnd {
```
keitwb: This function is getting pretty unwieldy.  Also it's pretty confusing having a function called `addDps` and another called `addDp` that do radically different things.

Issue #1115, opened 2019-12-13
Enable building custom monitors from extenal repos


captncraig:
I have some custom integrations I would like to create as native signalfx agent monitors. I definitely prefer this to alternatives (python / collectd exec / prometheus) because it is fairly straightforward, and we already have a body of go collectors that can be fairly easily transformed into monitors.

The main problems are in logistics. Primarily the fact that all of the interfaces I need to create a monitor are in the `internal` packages. This means I can only build them **inside** of this repository. There are a number of reasons I (or others) might not want to do that:

- We have very specific integrations that will never be of interest to the community and will not be open-sourced.
- We would happily open source an integration, but would prefer to keep it separate from this repo. Possibly for governance or licensing or ownership reasons.
- We would like to be able to test and compile the integration in isolation from the rest of the agent code as much as possible.

Currently to make custom integrations I must build the agent myself. I agree this is probably unavoidable, and plugins are probably not worth the trouble for all the headache they give. I have another suggestion for making this part a little easier, but that will need to be another issue.

But right now my only option is to essentially:

1. Fork this repo
2. Add my own monitors in (adding files that are extremely unlikely to introduce future conflicts.
3. Add a new file somewhere sole to `import _` my package(s).
4. Compile the code.

I don't like this for a few reasons:

1. It is a bit of a pain to merge in upstream changes.
2. New developers on our end don't know the boundaries between our code and your code.

I would much rather have a dedicated monitors repo that has scripts to build the entire agent as appropriate. So my proposal is simply this:

I would like to move some packages from `github.com/signalfx/signalfx-agent/internal` into `github.com/signalfx/signalfx-agent/pkg`.

Specifically, I just need to move the packages that need to be directly imported to build a monitor. Going from the documentation, that includes:

- `github.com/signalfx/signalfx-agent/internal/core/config`
- `github.com/signalfx/signalfx-agent/internal/monitors`
- `github.com/signalfx/signalfx-agent/internal/monitors/types`
- `github.com/signalfx/signalfx-agent/internal/utils`

It is a fairly simple change, but it does touch a large number of files. Is there a significant reason not to make that change? I have tested this change, and have gotten things compiling after moving the packages. I will make a PR shortly.

keitwb:
Would you write your own `main` method that proxies the agent's `main`, imports your new monitors, and then compile that?  I'm curious what the best workflow would be for what you are trying to do and if there is anything else the agent could do to facilitate it other than moving around packages.

[...]

PR #1116, opened 2019-12-13
Move some packages from internal to pkg

captncraig:
This is my attempt to implement the changes I proposed in #1115.

I use the following snippet to rename a package and rewrite imports:

```
SFX="\"github.com/signalfx/signalfx-agent"
PKG="monitors"
mkdir -p pkg/$PKG
git mv internal/$PKG/*.go pkg/$PKG
grep -lR "$SFX/internal/$PKG\"" | xargs sed -i "s=$SFX/internal/$PKG\"=$SFX/pkg/$PKG\"=g"
goimports -w -l .
```

Currently I have moved the following packages from `internal` to `pkg`:

- `monitors`
- `core/config`
- `monitors/types`
- `utils`

These are the ones required to implement a simple monitor. More can be moved in the future as needed or requested.


keitwb:
Thanks @captncraig.  This is something that I've figured would need to happen someday so thanks for doing it!

Looks like the monitor code gen binary is dependent on non-local agent code based on the error at https://circleci.com/gh/signalfx/signalfx-agent/196587?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link.  I will fix it and push it to this branch.

captncraig:
I think I figured out how to regen the docs properly. Fingers crossed.

captncraig:
@keitwb I see the circleci failure you are talking about now. I am definitely not sure how to fix it.

keitwb:
> @keitwb I see the circleci failure you are talking about now. I am definitely not sure how to fix it.

Yes, will take a look.

I'm also considering just moving everything to `pkg` and not have the split between `internal` and `pkg` but will think about it some more.

captncraig:
> I'm also considering just moving everything to pkg and not have the split between internal and pkg but will think about it some more.

That would certainly be simpler. I assumed things were in internal for a reason, so didn't really want to rock that too hard. Keeping things in internal is a little nice because you can make breaking api changes easier, but having a split like I have done also makes things confusing.

keitwb:
Yes, the encapsulation in `internal/` is nice for signifying things that might break but I think the confusion of having things split is not really worth it.  Also others might want to add/extend other aspects of the agent beyond just adding new monitors.

As far as "breaking changes", I would say that we are willing to help out anybody who is trying to do what you are doing but we aren't making any guarantees about compatibility of code in `pkg` between even minor releases of the agent.  The agent's versioning scheme is centered around compatibility of the configuration and MTSs emitted by the agent, and adding another layer of API compatibility would make things excessively inflexible.  In practice, given the history of the packages you would have to touch, I doubt it would break very often or in ways that are difficult to fix up.  Any "breaking changes" to the monitor interface requires us to fix up several dozen monitors, so it isn't something we do often!

captncraig:
I made a branch that completely moves everything:

https://github.com/signalfx/signalfx-agent/compare/master...captncraig:nointernal

For some reason github crashes every time I try to make a pr from it. I have no idea why.

keitwb:
Yeah I've seen crashes when you make a PR that is very big.  I think I've figured out the issue with the build that this PR has.  I'll try making a PR from a separate branch with the agent main package split out and the code generated modules committed.

keitwb:
Ok, I've got it working in a separate branch but I'm running into the same Github limitation, arg!  I'll see if somebody else in the office can make it go through but if not I might have to bypass the PR route.

captncraig:
I wonder if it could be hacked by opening a very simple pr, then pushing more changes into it?

keitwb:
That worked! See https://github.com/signalfx/signalfx-agent/pull/1118.  Going to close this one and continue everything in that PR.

PR #1117, opened 2019-12-16
Add jaegergrpc monitor

codesmith14:


internal/monitors/jaegergrpc/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+monitors:
+- dimensions:
+  doc: |
+    Runs a GRPC server that listens for Jaeger trace batches
+    and forwards them to SignalFx (or the configured ingest host in the
+    `writer` section of the agent config).  By default, the server listens on
+    localhost port 14250 but can be configured to anything.
+  metrics:
+  monitorType: jaegergrpc
```
keitwb: Make the monitor type `jaeger-grpc` so it is easier to read in the config.

internal/monitors/jaegergrpc/monitor.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return creds, err
+}
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `yaml:",inline" acceptsEndpoints:"false" singleInstance:"true"`
+	// The host:port on which to listen for datapoints.  The listening server
+	// accepts datapoints on the same HTTP path that ingest/gateway accepts
```
keitwb: This doc doesn't seem very relevant for this monitor.

internal/monitors/jaegergrpc/monitor.go
```diff
@@ -0,0 +1,161 @@
...
+		// send the spans on through the agent
+		m.Output.SendSpans(spans...)
+	}
+
+	return &api_v2.PostSpansResponse{}, nil
+}
+
+// Configure the monitor and kick off volume metric syncing
+func (m *Monitor) Configure(conf *Config) error {
+	m.lock.Lock()
```
keitwb: You don't need a lock because `Configure` can now only be called once per monitor instance.

keitwb: Also `Configure` and `Shutdown` happen in the same goroutine so there won't be any issues of memory coherency.

internal/monitors/jaegergrpc/monitor.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	return sourceIP, true
+}
+
+// PostSpans implements the jeager api_v2.CollectorServiceServer interface.  The grpc server will pass the jaeger
+// batches it receives to this method.  This method will convert the jaeger batches to SignalFx spans and pass them
+// on to the output writer.
+func (m *Monitor) PostSpans(ctx context.Context, r *api_v2.PostSpansRequest) (*api_v2.PostSpansResponse, error) {
+	if r != nil {
```
keitwb: It is a little clearer to just return the empty response (or maybe some error since getting a nil request seems abnormal) if `r == nil` instead of nesting all of the core logic of the function in this if statement.

internal/monitors/jaegergrpc/monitor.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		grpcOptions = []grpc.ServerOption{grpc.Creds(creds)}
+	}
+
+	// create the grpc server
+	m.grpc = grpc.NewServer(grpcOptions...)
+
+	var err error
+	// create a listener with the configured ListenAddress
+	m.ln, err = net.Listen("tcp", conf.ListenAddress)
```
keitwb: Move this into the goroutine and also implement a retry loop that retries indefinitely to listen on this port until shutdown.  The key thing is that if something happens to be listening on this port when the agent start but later stops listening, then the monitor should not just fail out at the beginning and then never work until an agent restart.  Make it log an error message whenever it can't listen since that is still a bad condition.

One of the core properties of monitors that we want is that they are persistent in the face of errors and can reach the desired monitoring state when it is possible to do so.

PR #1119, opened 2019-12-16
Add support for exporting traces in SAPM protocol

owais:


owais:
Depends on: https://github.com/signalfx/golib/pull/167

owais:
Looking into the CI failure

go.mod
```diff
@@ -6,10 +6,16 @@ replace github.com/creasty/defaults => github.com/signalfx/defaults v1.2.2-0.201
 
 replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
 
+replace k8s.io/api => k8s.io/api v0.0.0-20181110191121-a33c8200050f
```
keitwb: This isn't necessary.  Just do `go get k8s.io/api@a33c8200050f` and it will set the version in the `require` section below.

internal/core/config/config.go
```diff
@@ -22,6 +22,11 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	protoSAPM = "sapm"
+	protoSFX  = "sfx"
```
keitwb: The protocol is actually Zipkin v2.  My understanding is that we are trying to avoid calling it a "SignalFx" protocol.

go.mod
owais: I tried it but it did not work for me. Was seeing this and then gave up :)

```
build github.com/signalfx/signalfx-agent/cmd/monitorcodegen: cannot load k8s.io/api/admissionregistration/v1alpha1: module k8s.io/api@latest found (v0.17.0), but does not contain package k8s.io/api/admissionregistration/v1alpha1
```

internal/core/writer/writer.go
```diff
@@ -115,6 +115,10 @@ func New(conf *config.WriterConfig, dpChan chan []*datapoint.Datapoint, eventCha
 		dpChan:            dpChan,
 		spanChan:          spanChan,
 	}
+	if conf.ExportSAPM() {
+		sw.client = sfxclient.WithSAPMExporter(sw.client)
+	}
```
tigrannajaryan: Nit: move `sfxclient.NewHTTPSink()` case to `else` here for symmetry?

keitwb: Yes, it would be a bit clearer to see the two distinct clients set closer together in code.

I'm assuming the WithSAPMExporter client behaves the same as the NewHTTPSink client for datapoints and events?

go.mod
keitwb: Also I just realized you are downgrading this package which is generally the opposite direction we want to go.  Let me check and see if I can get it building without this.

owais: Thanks

internal/core/config/config.go
owais: Updated

internal/core/writer/writer.go
owais: Updated by using the options pattern

owais: > I'm assuming the WithSAPMExporter client behaves the same as the NewHTTPSink client for datapoints and events?

Yes, it only affect trace exporter. I'll update it to be more descriptive.

owais: Updated

owais:
Updated PR.

owais:
Thanks @keitwb. Can you please also take a look at https://github.com/signalfx/golib/pull/167

PR #1120, opened 2019-12-16
Add vSphere monitor

pmcollins:


internal/monitors/vsphere/model/model.go
```diff
@@ -0,0 +1,100 @@
...
+	Host                 string `yaml:"host"`
+	Port                 uint16 `yaml:"port"`
+	// The vSphere username.
+	Username string `yaml:"username"`
+	// The vSphere password.
+	Password string `yaml:"password"`
+	// Whether we verify the server's certificate chain and host name
+	InsecureSkipVerify bool `yaml:"InsecureSkipVerify"`
+	// How often to reload the inventory and inventory metrics.
+	InventoryRefreshIntervalSeconds int `yaml:"inventoryRefreshIntervalSeconds"`
```
jaysfx: You can make this a `time.Duration` and set the default to be "10s" and it'll magically convert it. Example: https://github.com/signalfx/signalfx-agent/blob/e3e5a713f34ffc053b447f6da9aa939be2f0a09e/internal/monitors/logstash/tcp/monitor.go#L48

internal/monitors/vsphere/service/log.go
```diff
@@ -0,0 +1,5 @@
+package service
+
+import "github.com/sirupsen/logrus"
+
+var Log = logrus.WithField("monitorType", "vsphereInfo")
```
jaysfx: Does Log need to be exported?

pmcollins: Yeah because it's used by the `vsphere` package. Not sure if this is an idiomatic way to log though.

internal/monitors/vsphere/model/model.go
pmcollins: Nice, I'll do that instead.

internal/monitors/vsphere/metadata.yaml
```diff
@@ -0,0 +1,1093 @@
...
+      It may work with versions 5.5 and 5.1, but neither are officially supported.
+
+      Sample YAML configuration:
+      ```yaml
+      monitors:
+        - type: vsphere
+          host: "172.16.248.140"
+          username: "administrator@vsphere.local"
+          password: "S3cr3t"
+          InsecureSkipVerify: true
```
keitwb:
```suggestion
          insecureSkipVerify: true
```

internal/monitors/vsphere/metadata.yaml
```diff
@@ -0,0 +1,1093 @@
...
+    doc: |
+      A VMware vSphere deployment includes physical hosts, ESXi hypervisors, virtual machines, and vCenter Server.
+      This monitor reports metrics for a vSphere deployment by logging into a vCenter Server and periodically
+      retrieving data about the deployment and its real time performance data.
+
+      When this monitor starts up, it logs into vCenter Server and traverses the inventory, gathering and caching all of
+      the hosts and virtual machines, and their available metrics.
+
+      Then it periodically queries the vCenter for the performance data for those metrics. It does so at a rate of
+      once every 20 seconds, the interval at which vCenter makes real-time performance data avaialable.
```
keitwb: So if I set `intervalSeconds: 10` on the montior what does that do?  That will be the default btw.

internal/monitors/vsphere/metadata.yaml
```diff
@@ -0,0 +1,1093 @@
...
+        default: false
+        group: disk
+      vsphere.disk.kernel_latency_ms:
+        description: Average amount of time, in milliseconds, spent by VMkernel to process each SCSI command.
+        type: gauge
+        default: false
+        group: disk
+      vsphere.disk.kernel_read_latency_ms:
+        description: Average amount of time, in milliseconds, spent by VMkernel to process each SCSI read command.
+        type: gauge
```
keitwb: Should any of these millisecond metrics be gauges?  I'm confused why some are counters and some are gauges.

internal/monitors/vsphere/model/model.go
keitwb: Be sure to take the suffix "Seconds" out if you do.

internal/monitors/vsphere/monitor.go
```diff
@@ -0,0 +1,38 @@
...
+		func() interface{} { return &Monitor{} },
+		&model.Config{},
+	)
+}
+
+func (m *Monitor) Configure(conf *model.Config) error {
+	var ctx context.Context
+	ctx, m.cancel = context.WithCancel(context.Background())
+	r := newRunner(ctx, conf, m)
+	utils.RunOnInterval(ctx, r.run, time.Duration(conf.IntervalSeconds)*time.Second)
```
keitwb: This will default to 10 seconds. Is that going to cause problems with the 20 second metric cycle in vSphere?

internal/monitors/vsphere/service/log.go
keitwb: A better pattern that existing agent code isn't terribly consistent with is to have the logger be instantiated in the `Configure` method with fields for the monitor id (`conf.MonitorID`) and type being equal to the registered monitor type (`vsphere`, not `vsphereInfo`).  Then pass that logger instance around to the various structs that need logging.  This allows you to distinguish between multiple instances of the monitor is the case ever arose.

In general it is best to try and "push up" errors to the top level of the monitor or current go routine and log them there but sometimes for things like info/debug messages you will need the logger instance in child structs of the monitor so you will have to put it on those structs.

pkg/monitors/vsphere/metadata.yaml
```diff
@@ -0,0 +1,1093 @@
...
+        default: false
+        group: cpu
+      vsphere.cpu_swapwait_ms:
+        description: CPU time spent waiting for swap-in.
+        type: gauge
+        default: false
+        group: cpu
+      vsphere.cpu_system_ms:
+        description: Amount of time spent on system processes on each virtual CPU in the virtual machine.
+        type: counter
```
jaysfx: We use `cumulative` for cumulative counters in the metadata.

internal/monitors/vsphere/model/model.go
pmcollins: Done.

internal/monitors/vsphere/metadata.yaml
pmcollins: Done. Updated `model.Config` yaml tags as well.

internal/monitors/vsphere/metadata.yaml
pmcollins: Any value should be fine (I did test 10s). If it's too soon to request an interval, the number of intervals requested will be zero and no query will occur. If a lot of time has passed, as many intervals required to fill in the gap will be requested.

internal/monitors/vsphere/metadata.yaml
pmcollins: I just went by the [vmware docs](https://code.vmware.com/apis/197/vsphere/doc/cpu_counters.html) here -- if it said it's a "delta" type, then I mapped that to a counter, otherwise a gauge. But this is just for the docs -- the actual type sent to the sfx api comes from this function in points.go:

```
func statsTypeToMetricType(statsType types.PerfStatsType) datapoint.MetricType {
	if statsType == types.PerfStatsTypeDelta {
		return datapoint.Counter
	}
	return datapoint.Gauge
}
```
It's always possible that I made a data entry error in the yaml though.

internal/monitors/vsphere/monitor.go
pmcollins: Should be fine. As above, if it's not time to fetch data, then it won't try.

pkg/monitors/vsphere/metadata.yaml
pmcollins: Thx. Replaced all.

internal/monitors/vsphere/service/log.go
pmcollins: Sounds good, will do this.

pkg/monitors/vsphere/metadata.yaml
keitwb: Make sure the value being emitted is actually a cumulative and not just a delta incremental value.

pmcollins: Reverted these to `counter` based on feedback from Ben. Basically, everything is a `datapoint.Gauge` except for `PerfStatsTypeDelta` which is a `datapoints.Count`.

pkg/monitors/vsphere/model/model.go
```diff
@@ -0,0 +1,93 @@
...
+	Host                 string `yaml:"host"`
+	Port                 uint16 `yaml:"port"`
+	// The vSphere username
+	Username string `yaml:"username"`
+	// The vSphere password
+	Password string `yaml:"password"`
+	// Whether we verify the server's certificate chain and host name
+	InsecureSkipVerify bool `yaml:"insecureSkipVerify"`
+	// How often to reload the inventory and inventory metrics
+	InventoryRefreshInterval time.Duration `yaml:"inventoryRefreshInterval" default:"600s"`
```
keitwb: Maybe default to 60s?

pkg/monitors/vsphere/monitor.go
```diff
@@ -0,0 +1,40 @@
...
+		&model.Config{},
+	)
+}
+
+func (m *Monitor) Configure(conf *model.Config) error {
+	var ctx context.Context
+	ctx, m.cancel = context.WithCancel(context.Background())
+	log := logrus.WithField("monitorType", "vsphere")
+	r := newRunner(ctx, log, conf, m)
+	utils.RunOnInterval(ctx, r.run, time.Duration(conf.IntervalSeconds)*time.Second)
```
keitwb: Hardcode to 20s and add doc telling people that IntervalSeconds is a noop on this monitor.

pmcollins: Done.

pkg/monitors/vsphere/model/model.go
pmcollins: Done.

PR #1124, opened 2019-12-19
Allow setting event endpoint separate from datapoints and traces

keitwb:
This might be useful for a setup in which both the agent and OpenTelemetry
Collector are used together.

### DIFF
--- a/pkg/core/config/config.go
+++ b/pkg/core/config/config.go
@@ -35,8 +35,13 @@ type Config struct {
 	// The URL of SignalFx ingest server.  Should be overridden if using the
 	// SignalFx Gateway.  If not set, this will be determined by the
 	// `signalFxRealm` option below.  If you want to send trace spans to a
-	// different location, set the `traceEndpointUrl` option.
+	// different location, set the `traceEndpointUrl` option.  If you want to
+	// send events to a different location, set the `eventEndpointUrl` option.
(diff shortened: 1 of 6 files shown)


pkg/core/config/config.go
```diff
@@ -35,8 +35,13 @@ type Config struct {
...
 	// SignalFx Gateway.  If not set, this will be determined by the
 	// `signalFxRealm` option below.  If you want to send trace spans to a
-	// different location, set the `traceEndpointUrl` option.
+	// different location, set the `traceEndpointUrl` option.  If you want to
+	// send events to a different location, set the `eventEndpointUrl` option.
 	IngestURL string `yaml:"ingestUrl"`
+	// The full URL (including path) to the event ingest server.  If this is
+	// not set, all events will be sent to the same place as `ingestUrl`
+	// above.
+	EventEndpointURL string `yaml:"eventEndpointUrl"`
```
pjanotti: note: I see that this follows the pattern of `traceEndpointUrl` but there is also others without `Endpoint`, eg.: `ingestUrl` and `apiUrl`. I'm not sure which pattern is the preferred one... assuming that is this one.

keitwb: The intent behind the different naming convention is that `ingestUrl` and `apiUrl` are "base" URLs that the writer appends paths to, whereas `event/traceEndpointUrl` are full URLs including path.

PR #1127, opened 2019-12-20
Add heroku metadata monitor

asuresh4:
Collects Heroku dyno metadata if enabled: https://devcenter.heroku.com/articles/dyno-metadata#dyno-metadata


docs/monitors/heroku-metadata.md
```diff
@@ -0,0 +1,31 @@
...
+
+# heroku-metadata
+
+Monitor Type: `heroku-metadata` ([Source](https://github.com/signalfx/signalfx-agent/tree/master/pkg/monitors/heroku))
+
+**Accepts Endpoints**: **Yes**
+
+**Multiple Instances Allowed**: Yes
+
+## Overview
```
keitwb: This will need docs on how to enable the metadata in Heroku.

pkg/monitors/heroku/monitor.go
```diff
@@ -0,0 +1,64 @@
...
+		properties["heroku_app_name"] = os.Getenv("HEROKU_APP_NAME")
+		properties["heroku_slug_commit"] = os.Getenv("HEROKU_SLUG_COMMIT")
+		properties["heroku_release_creation_timestamp"] = os.Getenv("HEROKU_RELEASE_CREATED_AT")
+		properties["heroku_app_id"] = os.Getenv("HEROKU_APP_ID")
+
+		m.Output.SendDimensionUpdate(&types.Dimension{
+			Name:              "dyno_id",
+			Value:             dynoID,
+			Properties:        properties,
+			MergeIntoExisting: true,
```
keitwb: You probably want this to be `false` unless you know of something else manipulating the dyno_id dimension.

pkg/monitors/heroku/monitor.go
```diff
@@ -0,0 +1,64 @@
...
+func init() {
+	monitors.Register(&monitorMetadata, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Configure monitor
+func (m *Monitor) Configure(c *Config) error {
+	m.logger = utils.NewThrottledLogger(log.WithFields(log.Fields{"monitorType": "heroku-metadata"}), 20*time.Second)
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+
+	utils.RunOnInterval(m.ctx, func() {
```
keitwb: This doesn't need to run on an interval, only once.

keitwb: But do the run in a goroutine.

pkg/monitors/heroku/monitor.go
```diff
@@ -0,0 +1,64 @@
...
+			MergeIntoExisting: true,
+		})
+
+	}, time.Duration(c.IntervalSeconds)*time.Second)
+
+	return nil
+}
+
+// Shutdown stops the metric sync
+func (m *Monitor) Shutdown() {
```
keitwb: No need for a shutdown.

pkg/monitors/heroku/monitor.go
asuresh4: Removed

pkg/monitors/heroku/monitor.go
asuresh4: Updated

keitwb:
So the stuff from the collector app that will have dyno_id is the memory/load metrics and any custom metrics emitted by the app?  But router metrics will only have dyno name as a dimension?

pkg/monitors/heroku/monitor.go
```diff
@@ -0,0 +1,57 @@
...
+
+	go func() {
+		properties := map[string]string{}
+		dynoID := os.Getenv("HEROKU_DYNO_ID")
+
+		properties["heroku_release_version"] = os.Getenv("HEROKU_RELEASE_VERSION")
+		properties["heroku_app_name"] = os.Getenv("HEROKU_APP_NAME")
+		properties["heroku_slug_commit"] = os.Getenv("HEROKU_SLUG_COMMIT")
+		properties["heroku_release_creation_timestamp"] = os.Getenv("HEROKU_RELEASE_CREATED_AT")
+		properties["heroku_app_id"] = os.Getenv("HEROKU_APP_ID")
```
keitwb: Is there any way to get the dyno type?  e.g. Hobby, etc.

asuresh4:
> So the stuff from the collector app that will have dyno_id is the memory/load metrics and any custom metrics emitted by the app? But router metrics will only have dyno name as a dimension?

yes, that's right

pkg/monitors/heroku/monitor.go
asuresh4: I looked through the docs, I don't this information being exposed through the dyno metadata "endpoint". Seems like this information is only exposed via the platform api.

docs/monitors/heroku-metadata.md
asuresh4: Added docs

asuresh4:
> > So the stuff from the collector app that will have dyno_id is the memory/load metrics and any custom metrics emitted by the app? But router metrics will only have dyno name as a dimension?
> 
> yes, that's right

This not actually true. Custom metrics being logged by application will does not have `dyno_id`. Here's an example,

`164 <190>1 2019-12-21T22:21:26.705132+00:00 host app web.1 - gauge#quota_used=20 cumulative#response_bytes=100 sfxdimension#service=backend sfxdimension#client=sfx_app`

where the message

`gauge#quota_used=20 cumulative#response_bytes=100 sfxdimension#service=backend sfxdimension#client=sfx_app`

is what is logged out from an application and it will not contain Dyno id. However, it could be added by users since the application has access to the env variables on the dyno.

PR #1128, opened 2019-12-23
Update quick-install.md

galpizarsfx:
Updating headers to troubleshoot smart agent tile


galpizarsfx:
Tables seem to exist in other tiles, so I left those in, but I did remove the drop-down menus. @keitwb

galpizarsfx:
@keitwb Please review this PR to see if the changes I made to the Smart Agent tile were enough to troubleshoot. As a reminder, previously, I had made several changes to the Smart Agent tile, and when those changes were merged into master, the tile in lab did not display correctly. Your feedback was to remove any "fancy" features, so I removed the drop-down menus that I had previously added. (I kept the tables because tables seems to exist in other tiles, so I'm assuming tables should be fine to keep.)

PR #1132, opened 2020-01-03
standardize http client config

jaysfx:


jaysfx:
@keitwb please re-review, I made a few changes, mostly added testing.

pkg/core/common/auth/tls.go
```diff
@@ -72,15 +58,15 @@ func TLSConfig(tlsConfig *tls.Config, caCertPath string, clientCertPath string,
 	return tlsConfig, nil
 }
 
+// CertPool returns the system cert pool for non-Windows platforms
 func CertPool() (*x509.CertPool, error) {
```
jaysfx: @keitwb I was getting test failures on win32 and realized this function was returning a nil CertPool on win32 even when error was also nil. I think it makes sense to return an empty pool on win32. See what you think.

keitwb: Yeah makes sense to normalize nil to an empty pool for easier handling in caller funcs.

PR #1134, opened 2020-01-06
Remove default puppet ingest/api urls

### LINKED ISSUE #1133: Puppet module defaults are broken
Using this puppet module with its defaults results in the SignalFX agent getting 401's trying to send data, the problems seems to be with the apiUrl and ingestUrl, the current config is defined as:

https://github.com/signalfx/signalfx-agent/blob/6b8c8c1064ff21b8eb4fa992880ac9f27a84d4b2/deployments/puppet/data/default.yaml#L4

```
---
signalfx_agent::config:
  signalFxAccessToken: "MY_TOKEN"
  ingestUrl: https://ingest.signalfx.com
  apiUrl: https://api.signalfx.com
  hostname: "%{facts.fqdn}"
  intervalSeconds: 10
  logging:
    level: info
  observers: [
    {"type": "host"}
  ]
```

The [config schema](https://github.com/signalfx/signalfx-agent/blob/master/docs/config-schema.md) defines these as optional, the values of which are derived from the `signalFxRealm`. The schema doc also states that if these values are defined they should be done using the following format:

```
ingestUrl: "https://ingest.YOUR_SIGNALFX_REALM.signalfx.com"
apiUrl: "https://api.YOUR_SIGNALFX_REALM.signalfx.com"
traceEndpointUrl: "https://ingest.YOUR_SIGNALFX_REALM.signalfx.com/v1/trace"
```

Couple options to fix this, either remove these defaults from the config template, or add the full enpoints as per the schema, which would also require adding in the `signalFxRealm` with us1 as a default value.

keitwb:
Remove any references to default ingest/api urls.  The user should specify these without us providing any defaults.

Fixes #1133.

### DIFF
--- a/deployments/puppet/README.md
+++ b/deployments/puppet/README.md
@@ -14,6 +14,7 @@ class accepts the following parameters:
     ```ruby
     $config = {
       signalFxAccessToken: "MY_TOKEN",
+      signalFxRealm: "us1",
       enableBuiltInFiltering: true,
       monitors: [
         {type: "collectd/cpu"},
--- a/deployments/puppet/data/default.yaml
+++ b/deployments/puppet/data/default.yaml
@@ -1,8 +1,5 @@
 ---
 signalfx_agent::config:
-  signalFxAccessToken: "MY_TOKEN"
-  ingestUrl: https://ingest.signalfx.com
-  apiUrl: https://api.signalfx.com
   hostname: "%{facts.fqdn}"
   intervalSeconds: 10
   logging:


deployments/puppet/README.md
```diff
@@ -14,6 +14,7 @@ class accepts the following parameters:
     ```ruby
     $config = {
       signalFxAccessToken: "MY_TOKEN",
+      signalFxRealm: "us1",
```
jaysfx: Why defaulting to us1? 

Also if they specify signalFxRealm they won't have to specify ingesturl and apiurl right?

keitwb: That's not a default, just a documented example.

keitwb: Yes, if they specify realm the other options are automatically derived by the agent.

PR #1135, opened 2020-01-06
Update to kubernetes_workload property

mstumpfx:
**Breaking changes from 4.x**:
Previously we synced the `kubernetes_workload` property to _always_ be the direct parent type. This change will set the `kubernetes_workload` to possibly be the super-parents type, as explained below:

- If a pod is created by a deployment, we should set the `kubernetes_workload` property to be `Deployment` instead of a `ReplicaSet`.
- If a pod is created by a ReplicaSet directly, and that ReplicaSet was *not* created by a Deployment, we should set `kubernetes_workload` to be `ReplicaSet`.

---

- Also, if a pod is created by a CronJob, we should set the `kubernetes_workload` property to be `CronJob` instead of `Job`.
- If a pod is created by a Job directly, and that Job was *not* created by a CronJob, we should set `kubernetes_workload` to be `Job`.

### DIFF
--- a/pkg/monitors/kubernetes/cluster/metrics/dimensions.go
+++ b/pkg/monitors/kubernetes/cluster/metrics/dimensions.go
@@ -154,20 +154,28 @@ func (dh *DimensionHandler) handleAddPod(pod *v1.Pod) {
 		}
 	}
 
+	// Check if pod owner of type ReplicaSet is cached. Check owners reference on ReplicaSet to see if it was
+	// created by a Deployment. Sync properties accordingly
 	rsRef := k8sutil.FindOwnerWithKind(pod.OwnerReferences, "ReplicaSet")
 	if rsRef != nil {
 		if replicaSet := dh.replicaSetCache.Get(rsRef.UID); replicaSet != nil {
 			if deployRef := k8sutil.FindOwnerWithKind(replicaSet.OwnerReferences, "Deployment"); deployRef != nil {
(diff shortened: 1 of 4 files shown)


mstumpfx:
@keitwb 
> The dimension will not flap correct?

It shouldn't as far as I can tell. The property should only ever be set once. I will use ReplicaSet / Deployment / Pod events as an example but replace ReplicaSet/Deployment with Job/CronJob and it is the same logic:

Watch for Pod / ReplicaSet events:
1. Pod Event:
    - We sync dimensions for the pod, if the OwnerReference (OR) is a ReplicaSet or Job, we defer until later to set the `kubernetes_workload` property. Otherwise, it is set right away.
    - In the deferred function, we check if the ReplicaSet is cached (aka have we seen the RS event yet?). If we have, we check the OR for the cached ReplicaSet. If the cached OR has an OR of a deployment, set the `kubernetes_workload` property to `Deployment`. If it does *not* have an OR reference of a deplyoment, set the `kubernetes_workload` to ReplicaSet.
      - If the RS was *not* cached, that means we may have gotten the events out of order and we do *not* sync any `kubernetes_workload` property at this time. It is deferred until we see the RS event.

2. ReplicaSet event:
    - We add the ReplicaSet to the cache, and cache it's OwnerReference.
    - Loop through all pods in the same namespace as the RS
    - For each one, check if the pod was created by the RS
    - If the pod was created by the RS, check the RS OR and if it was a deployment, sync `kubernetes_workload : Deployment`. If the RS was not created by a deployment, sync `kubernetes_workload: ReplicaSet`

PR #1140, opened 2020-01-13
Add non-collectd load monitor and make default

keitwb:
The old collectd/load monitor is still present for backwards compatibility

The metrics are exactly the same except for the lack of a 'plugin' dimension


pkg/monitors/load/load.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/signalfx/signalfx-agent/pkg/core/config"
+	"github.com/signalfx/signalfx-agent/pkg/monitors"
+	"github.com/signalfx/signalfx-agent/pkg/monitors/types"
+	"github.com/signalfx/signalfx-agent/pkg/utils"
+	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
+)
+
+func init() {
+	if runtime.GOOS != "windows" {
```
asuresh4: would it make sense add a log message saying it isn't supported in case this monitor is configured on windows?

keitwb: It will essentially say that the monitor doesn't exist if they try and run it on windows and then if they look in the docs it clearly says that it only works on Linux, so I think that is enough.

asuresh4:
we should also probably add a deprecation notice in the the `collectd/load` monitor

PR #1141, opened 2020-01-14
Make net-io monitor default on Linux

keitwb:
- Added dropped packet metrics to match collectd/interface
- Replaced collectd/interface with net-io monitor in distributed configs
- Removed plugin and plugin_instance dimensions from datapoints

Breaking Changes:
 - plugin and plugin_dimension dimensions are gone and only the interface dimension remains

### DIFF
--- a/deployments/ansible/README.md
+++ b/deployments/ansible/README.md
@@ -37,7 +37,7 @@ This role sources the following variables:
         - type: collectd/cpufreq
         - type: filesystems
         - type: collectd/disk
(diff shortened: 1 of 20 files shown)


jaysfx:
Looks like a bunch of previous changes got pulled in (e.g. Python 3 switch)

keitwb:
> Looks like a bunch of previous changes got pulled in (e.g. Python 3 switch)

I needed to rebase it on the latest 5.0-staging branch.

PR #1143, opened 2020-01-15
Make vmem monitor default on Linux

keitwb:
- Remove beta warning for Linux
 - Remove plugin dimension
 - Clean up memory monitor

### DIFF
--- a/pkg/monitors/memory/memory_linux.go
+++ b/pkg/monitors/memory/memory_linux.go
@@ -1,0 +1,19 @@
+package memory
+
+import (
+	"time"
+
+	"github.com/shirou/gopsutil/mem"
+	"github.com/signalfx/golib/v3/datapoint"
+)
+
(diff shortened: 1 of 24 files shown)


pkg/monitors/memory/memory_linux.go
```diff
@@ -0,0 +1,19 @@
+package memory
+
+import (
+	"time"
+
+	"github.com/shirou/gopsutil/mem"
+	"github.com/signalfx/golib/v3/datapoint"
+)
+
+func (m *Monitor) makeMemoryDatapoints(memInfo *mem.VirtualMemoryStat, dimensions map[string]string) []*datapoint.Datapoint {
```
bjsignalfx: Wait, how is this magic happening? Same function signature in the same package. How is this not a compile error?

keitwb: See https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool, esp. the section of "File Suffixes".

PR #1144, opened 2020-01-15
RFC: Example of collectable monitor

jaysfx:
One of the most common patterns is for a monitor to do collection of metrics every
`intervalSeconds`. This adds an (optional) Collectable interface where if
the monitor implements it we will call it every `intervalSeconds`.

This simplifies monitors that follow this pattern as well as make it easier to implement
other features like checking how long collect intervals are taking and logging if
they're taking longer than `intervalSeconds` which would result in lagging datapoints.


pkg/monitors/diskio/diskio_others.go
```diff
@@ -82,10 +77,9 @@ func (m *Monitor) Configure(conf *Config) error {
...
 
-	// gather metrics on the specified interval
-	utils.RunOnInterval(ctx, func() {
-		m.emitDatapoints()
-	}, time.Duration(conf.IntervalSeconds)*time.Second)
-
 	return nil
 }
+
+func (m *Monitor) Collect(ctx context.Context) {
```
keitwb: Add an interface type assertion in this module (`var _ Collectable = new(Monitor)`)

PR #1147, opened 2020-01-17
Make cpu monitor default instead of collectd/cpu

keitwb:
- Also remove collectd/cpufreq from standard configs
 - Replace collectd/cpu with cpu in some of the integration tests

### DIFF
--- a/pkg/monitors/collectd/cpu/metadata.yaml
+++ b/pkg/monitors/collectd/cpu/metadata.yaml
@@ -1,6 +1,10 @@
 monitors:
 - dimensions:
   doc: |
+    **This monitor is deprecated in favor of the `cpu` monitor.  Please switch
+    to that monitor, as this monitor will be removed in a future agent
+    release.**
+
     This monitor collects cpu usage data using the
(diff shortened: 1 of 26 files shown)


pkg/monitors/collectd/cpu/metadata.yaml
```diff
@@ -1,6 +1,10 @@
 monitors:
 - dimensions:
   doc: |
+    **This monitor is deprecated in favor of the `cpu` monitor.  Please switch
+    to that monitor, as this monitor will be remove in a future agent
```
mstumpfx:
```suggestion
    to that monitor, as this monitor will be removed in a future agent
```

mstumpfx: might be useful to just put "removed in 5.0" instead of future

keitwb: > might be useful to just put "removed in 5.0" instead of future

It isn't being removed in 5.0, it is just not part of the default config.  It will be totally removed in some later major release.

PR #1153, opened 2020-01-21
Improve performance of host observer

jaysfx:
This skips the uid lookup since we don't use it. See upstream PRs
for some more background:

https://github.com/shirou/gopsutil/issues/784
https://github.com/shirou/gopsutil/pull/783

This requires upgrading gopsutil which changed some structs which in
turn requires upgrading telegraf. Apparently the `Stolen` field was never set
(https://github.com/shirou/gopsutil/pull/677)

Even with this change there are still some performance issues in the host
observer on machines with a large number of connections. Something like this
change would be required to deal with the remaining issues:

https://github.com/shirou/gopsutil/issues/695

### DIFF
--- a/pkg/monitors/cpu/cpu.go
+++ b/pkg/monitors/cpu/cpu.go
@@ -240,8 +240,7 @@ func cpuTimeStatTototalUsed(t *cpu.TimesStat) *totalUsed {
 		t.Softirq +
 		t.Steal +
 		t.Guest +
-		t.GuestNice +
-		t.Stolen
+		t.GuestNice
 
 	return &totalUsed{
 		Total: total,
--- a/go.mod
+++ b/go.mod
@@ -82,7 +82,7 @@ require (
 	github.com/influxdata/influxdb v1.7.4 // indirect
 	github.com/influxdata/platform v0.0.0-20190117200541-d500d3cf5589 // indirect
 	github.com/influxdata/tail v1.0.0 // indirect
-	github.com/influxdata/telegraf v0.10.2-0.20190319005412-5e88824c153e
+	github.com/influxdata/telegraf v0.10.2-0.20200121190823-6dad859d74c2
 	github.com/influxdata/toml v0.0.0-20180607005434-2a2e3012f7cf // indirect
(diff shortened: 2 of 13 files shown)


pkg/monitors/cpu/cpu.go
```diff
@@ -240,8 +240,7 @@ func cpuTimeStatTototalUsed(t *cpu.TimesStat) *totalUsed {
 		t.Softirq +
 		t.Steal +
 		t.Guest +
-		t.GuestNice +
-		t.Stolen
+		t.GuestNice
```
keitwb: What does this have to do with "Improve performance of host observer"?

keitwb: Also, we probably need the stolen metric for consistency.

keitwb: See https://github.com/signalfx/signalfx-agent/pull/1093/files#diff-1d0f8cf86bd7c90de47e9a231a73e489R234.

keitwb: See https://github.com/signalfx/signalfx-agent/pull/1093/files#diff-1d0f8cf86bd7c90de47e9a231a73e489R234.

keitwb:
I don't really like gopsutil anymore given how clunky the interface is and how little control it gives you over the data that it gathers and how it does it.

When I did the new processlist montior for Linux I just found a library more specific to that purpose and avoided gopsutil altogether. That might be the better solution here since there are still performance issues even with this that require changing gopsutil to fix.

keitwb:
@jaysfx let's go ahead and push through this PR and we'll deal with the other issues later.  This seems like it would help quite a bit as is.  Just get the CI tests passing.

jaysfx:
Looks like mssql server monitor has a lot of new metrics. May just let the "ignore undocumented metrics" handle them since there's a lot of new ones to document.

keitwb:
> Looks like mssql server monitor has a lot of new metrics. May just let the "ignore undocumented metrics" handle them since there's a lot of new ones to document.

Yeah, looks pretty complicated: https://github.com/influxdata/telegraf/tree/master/plugins/inputs/sqlserver.  It's not even clear that there is a full list of metrics from telegraf's side anyway.

Issue #1151, opened 2020-01-19
Agent crashing on kubernetes cluster


coeyal:
Agent keeps crashing on kubernetes.
I'm using EKS cluster 
`Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.12-eks-eb1860", GitCommit:"eb1860579253bb5bf83a5a03eb0330307ae26d18", GitTreeState:"clean", BuildDate:"2019-12-23T08:58:45Z", GoVersion:"go1.11.13", Compiler:"gc", Platform:"linux/amd64"}`

Here is the Error from the logs:
`time="2020-01-19T21:47:49Z" level=info msg="Starting K8s API resource sync"
E0119 21:47:49.547438       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 2170 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x2260c00, 0x40b14c0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.0/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.0/pkg/util/runtime/runtime.go:48 +0x82
panic(0x2260c00, 0x40b14c0)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/signalfx/signalfx-agent/pkg/monitors/kubernetes/cluster/metrics.datapointsForJob(0xc00103e000, 0x2488e00, 0x2, 0x20ed403)
	/usr/src/signalfx-agent/pkg/monitors/kubernetes/cluster/metrics/jobs.go:25 +0x1ff
github.com/signalfx/signalfx-agent/pkg/monitors/kubernetes/cluster/metrics.(*DatapointCache).HandleAdd(0xc000cb3800, 0x2b75220, 0xc00103e000, 0x2b75220, 0xc00103e000)
	/usr/src/signalfx-agent/pkg/monitors/kubernetes/cluster/metrics/cache.go:109 +0x78d
github.com/signalfx/signalfx-agent/pkg/monitors/kubernetes/cluster.(*State).beginSyncForType.func3(0xc000f98480, 0x17, 0x17, 0xc000ddd808, 0x8, 0x0, 0x0)
[...]

PR #1154, opened 2020-01-21
kubernetes-cluster monitor bug fix for job metrics

mstumpfx:
To address user reported https://github.com/signalfx/signalfx-agent/issues/1151

The `kubernetes.job.completions` and `kubernetes.job.paralellism` metrics
may be unset, resulting in null pointer errors.
This commit adds checks for existence of each attribute, as well as a
new test for checking their lack of existence.

### DIFF
--- a/pkg/monitors/kubernetes/cluster/metrics/jobs.go
+++ b/pkg/monitors/kubernetes/cluster/metrics/jobs.go
@@ -18,38 +18,53 @@ func datapointsForJob(job *batchv1.Job) []*datapoint.Datapoint {
 		"kubernetes_name":      job.Name,
 	}
 
-	return []*datapoint.Datapoint{
-		datapoint.New(
-			"kubernetes.job.completions",
-			dimensions,
-			datapoint.NewIntValue(int64(*job.Spec.Completions)),
-			datapoint.Gauge,
-			time.Time{}),
-		datapoint.New(
-			"kubernetes.job.parallelism",
-			dimensions,
-			datapoint.NewIntValue(int64(*job.Spec.Parallelism)),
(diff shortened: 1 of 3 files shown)


pkg/monitors/kubernetes/cluster/metrics/jobs.go
```diff
@@ -18,38 +18,53 @@ func datapointsForJob(job *batchv1.Job) []*datapoint.Datapoint {
 		"kubernetes_name":      job.Name,
 	}
 
-	return []*datapoint.Datapoint{
-		datapoint.New(
-			"kubernetes.job.completions",
-			dimensions,
-			datapoint.NewIntValue(int64(*job.Spec.Completions)),
-			datapoint.Gauge,
-			time.Now()),
```
keitwb: Take out all of the `time.Now()`s and replace them with `time.Time{}`.  You are making a bunch of system calls every time you make a datapoint and you don't really need a timestamp anyway since higher up code in this monitor will set the timestamp later.

mstumpfx: Should we just change all of these in a separate PR? These are littered all throughout the `kubernetes-monitor`?

keitwb: You can if you want but let's do it in this same release as this.

mstumpfx: I can rebase after https://github.com/signalfx/signalfx-agent/pull/1156

keitwb:
I ran the tests manually and they passed.  Merging now.

PR #1158, opened 2020-01-22
Convert haproxy, conviva, ecs, logstash, expvar to use httpconfig

jaysfx:
# 5.0 release notes
These monitors have been updated to have standard HTTP/HTTPS configuration options:

* haproxy
* conviva
* ecs
* logstash
* expvar

These include options to enable HTTPS as well as configure TLS certificates and timeouts.

Note that if the monitor had an option like `timeoutSeconds: 5` before the option is now `httpTimeout: 5` or `httpTimeout: 5s` (note the `s`). The durations can be explicitly specified as part of the value using Go's [ParseDuration syntax](https://golang.org/pkg/time/#ParseDuration). If the duration is left off it will be interpreted as seconds.

## conviva
`pulseUsername` option is now `username`

## haproxy
`sslVerify` which defaulted tor `true` is now `skipVerify` which defaults to `false`. The defaults are equivalent but if you had the option set the value is now inverse.

### DIFF
--- a/docs/monitors/conviva.md
+++ b/docs/monitors/conviva.md
@@ -149,9 +149,14 @@ Configuration](../monitor-config.md#common-configuration).**
 
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
-| `pulseUsername` | **yes** | `string` | Conviva Pulse username required with each API request. |
-| `pulsePassword` | **yes** | `string` | Conviva Pulse password required with each API request. |
-| `timeoutSeconds` | no | `integer` |  (**default:** `10`) |
+| `httpTimeout` | no | `int64` | HTTP timeout duration for both read and writes. This should be a duration string that is accepted by https://golang.org/pkg/time/#ParseDuration (**default:** `10s`) |
+| `username` | no | `string` | Basic Auth username to use on each request, if any. |
+| `password` | no | `string` | Basic Auth password to use on each request, if any. |
+| `useHTTPS` | no | `bool` | If true, the agent will connect to the exporter using HTTPS instead of plain HTTP. (**default:** `false`) |
+| `skipVerify` | no | `bool` | If useHTTPS is true and this option is also true, the exporter's TLS cert will not be verified. (**default:** `false`) |
+| `caCertPath` | no | `string` | Path to the CA cert that has signed the TLS cert, unnecessary if `skipVerify` is set to false. |
+| `clientCertPath` | no | `string` | Path to the client TLS cert to use for TLS required connections |
+| `clientKeyPath` | no | `string` | Path to the client TLS key to use for TLS required connections |
 | `metricConfigs` | no | `list of objects (see below)` | Conviva metrics to fetch. The default is quality_metriclens metric with the "All Traffic" filter applied and all quality_metriclens dimensions. |
 
 
--- a/docs/monitors/haproxy.md
+++ b/docs/monitors/haproxy.md
@@ -73,16 +73,20 @@ Configuration](../monitor-config.md#common-configuration).**
 
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
+| `httpTimeout` | no | `int64` | HTTP timeout duration for both read and writes. This should be a duration string that is accepted by https://golang.org/pkg/time/#ParseDuration (**default:** `10s`) |
+| `username` | no | `string` | Basic Auth username to use on each request, if any. |
+| `password` | no | `string` | Basic Auth password to use on each request, if any. |
+| `useHTTPS` | no | `bool` | If true, the agent will connect to the exporter using HTTPS instead of plain HTTP. (**default:** `false`) |
+| `skipVerify` | no | `bool` | If useHTTPS is true and this option is also true, the exporter's TLS cert will not be verified. (**default:** `false`) |
+| `caCertPath` | no | `string` | Path to the CA cert that has signed the TLS cert, unnecessary if `skipVerify` is set to false. |
+| `clientCertPath` | no | `string` | Path to the client TLS cert to use for TLS required connections |
+| `clientKeyPath` | no | `string` | Path to the client TLS key to use for TLS required connections |
 | `host` | no | `string` | The host/ip address of the HAProxy instance. This is used to construct the `url` option if not provided. |
 | `port` | no | `integer` | The port of the HAProxy instance's stats endpoint (if using HTTP). This is used to construct the `url` option if not provided. (**default:** `0`) |
 | `path` | no | `string` | The path to HAProxy stats. The default is `stats?stats;csv`. This is used to construct the `url` option if not provided. (**default:** `stats?stats;csv`) |
-| `useHTTPS` | no | `bool` | Whether to connect on HTTPS or HTTP. If you want to use a UNIX socket, then specify the `url` config option with the format `unix://...` and omit `host`, `port` and `useHTTPS`. (**default:** `false`) |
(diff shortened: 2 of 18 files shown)


pkg/monitors/haproxy/helpers.go
```diff
@@ -262,10 +261,12 @@ func parseStatusField(v string) int64 {
...
 
 func httpReader(conf *Config, method string) (io.ReadCloser, error) {
-	client := http.Client{
-		Timeout:   conf.Timeout,
-		Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: !conf.SSLVerify}},
+	client, err := conf.Build()
+	if err != nil {
+		return nil, err
 	}
+	defer client.CloseIdleConnections()
```
jaysfx: This should really be refactored to reuse http.Client. I started but haven't finished. It's no worse than it was before though.

pkg/monitors/haproxy/monitor.go
```diff
@@ -60,35 +59,22 @@ func (m *Monitor) Configure(conf *Config) (err error) {
...
-			ch := make(chan []*datapoint.Datapoint, 1)
 			wg.Add(1)
 			go func() {
-				defer close(ch)
-				defer wg.Done()
-				select {
-				case <-ctx.Done():
-					logger.Error(ctx.Err())
-					return
-				case ch <- fn(conf, pxs):
```
jaysfx: I removed this because I don't think the context cancelling actually worked. My understanding is that the call to `fn` would always block until finished since it doesn't take in `ctx`. Having the select with `ch <- fn(conf, pxs)` I believe is for doing non-blocking send calls and `ch` would never block since it had size `1`.

Anyway it didn't seem necessary overall so I simplified things a bit.

docs/monitors/logstash.md
```diff
@@ -31,10 +31,16 @@ Configuration](../monitor-config.md#common-configuration).**
 
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
+| `httpTimeout` | no | `int64` | HTTP timeout duration for both read and writes. This should be a duration string that is accepted by https://golang.org/pkg/time/#ParseDuration (**default:** `10s`) |
```
keitwb: If somebody changes `timeoutSeconds: 5` to `httpTimeout: 5`, what will it be interpreted as?  Or will it error?

pkg/monitors/haproxy/monitor.go
keitwb: @bjsignalfx please confirm.

bjsignalfx: @jaysfx No, I think the select statement blocks on both cases until there's stuff to read.  So, assuming fn() is long running and context cancellation is issued when timeout is reached, then the `case <-ctx.Done()` case will execute and the goroutine will return.

jaysfx: I don't think that's the case. Here's a test case I came up with, see if you think it's correctly modeling it:

https://gist.github.com/jaysfx/284e87b34bc8fb8e0f7228a71c8859e8

You get the output 

```
before sleep
after sleep
slow returned
```

I don't think the select statement runs until _after_ the `slow` function returns. I think to have the intended effect ctx needs to be passed into the slow function so that it can be cancelled.

jaysfx: To make it work I think you would have to call `fn` inside another separate goroutine and then do the select on a channel.

docs/monitors/logstash.md
jaysfx: Unfortunately it gets interpreted as nanoseconds since time.Duration is an `int64`.

pkg/monitors/haproxy/monitor.go
jaysfx: To make it work I think you would have to call fn inside another separate goroutine and then do the select on a channel.

jaysfx:
> Put in the commit message how to migrate existing integrations that use any changed fields to the new form. Something that could be copy/pasted into a 5.0 upgrade guide. Basically all of the breaking changes should be documented in git commits explicitly.

See updated PR description (I'll include it when squashing)

docs/monitors/logstash.md
jaysfx: I'm adding a custom time.Duration for it so we can enforce it being a string. I can easily see people (or me..) forgetting to specify the unit and have it silently using things like 5ns as the timeout.

keitwb: Ok, yeah I think having the default interpretation of integers as seconds is the best way to go.

pkg/monitors/haproxy/monitor.go
bjsignalfx: Yes @jaysfx, you are right. If you can, please make this change instead.

docs/monitors/logstash.md
jaysfx: I've done the PR for this in https://github.com/signalfx/signalfx-agent/pull/1168 as we have existing time.Duration usage in master that this would benefit from.

jaysfx:
I think this is ready. chef tests are failing but unrelated.

docs/monitors/conviva.md
```diff
@@ -149,9 +149,14 @@ Configuration](../monitor-config.md#common-configuration).**
 
 | Config option | Required | Type | Description |
 | --- | --- | --- | --- |
-| `pulseUsername` | **yes** | `string` | Conviva Pulse username required with each API request. |
```
keitwb: This breaking change is not mentioned in the description.

docs/monitors/haproxy.md
```diff
@@ -73,16 +73,20 @@ Configuration](../monitor-config.md#common-configuration).**
...
 | `port` | no | `integer` | The port of the HAProxy instance's stats endpoint (if using HTTP). This is used to construct the `url` option if not provided. (**default:** `0`) |
 | `path` | no | `string` | The path to HAProxy stats. The default is `stats?stats;csv`. This is used to construct the `url` option if not provided. (**default:** `stats?stats;csv`) |
-| `useHTTPS` | no | `bool` | Whether to connect on HTTPS or HTTP. If you want to use a UNIX socket, then specify the `url` config option with the format `unix://...` and omit `host`, `port` and `useHTTPS`. (**default:** `false`) |
-| `url` | no | `string` | URL on which to scrape HAProxy. Scheme `http://` for http-type and `unix://` socket-type urls. If this is not provided, it will be derive from the `host`, `port`, `path`, and `useHTTPS` options. |
-| `username` | no | `string` | Basic Auth username to use on each request, if any. |
-| `password` | no | `string` | Basic Auth password to use on each request, if any. |
-| `sslVerify` | no | `bool` | Flag that enables SSL certificate verification for the scrape URL. (**default:** `true`) |
```
keitwb: This is a breaking change.

docs/monitors/haproxy.md
```diff
@@ -73,16 +73,20 @@ Configuration](../monitor-config.md#common-configuration).**
...
-| `timeout` | no | `int64` | Timeout for trying to get stats from HAProxy. This should be a duration string that is accepted by https://golang.org/pkg/time/#ParseDuration (**default:** `5s`) |
+| `url` | no | `string` | Whether to connect on HTTPS or HTTP. If you want to use a UNIX socket, then specify the `url` config option with the format `unix://...` and omit `host`, `port` and `useHTTPS`. |
```
keitwb: This description seems off, it makes it sound like it is a bool.

PR #1162, opened 2020-01-23
Remove old-style filtering and whitelist.json

keitwb:
This forces enableBuiltInFiltering to always be true (the default since 4.7.0)

It will break anybody who was still using the old whitelist.json filtering.  The
migration steps are in docs/legacy-filtering.md


pkg/core/config/config.go
```diff
@@ -101,21 +101,12 @@ type Config struct {
...
-	// `metricsToInclude` as before.  If you set this option to `true`, the
-	// `whitelist.json` entry under `metricToExclude` should be removed, if it
-	// is present -- otherwise custom metrics won't be emitted.
-	EnableBuiltInFiltering bool `yaml:"enableBuiltInFiltering" default:"false"`
+	// This must be unset or explicitly set to true. In prior versions of the
+	// agent, there was a filtering mechanism that relied heavily on an
+	// external whitelist.json file to determine which metrics were sent by
+	// default.  This is all inherent to the agent now and the old style of
+	// filtering is no longer available.
+	EnableBuiltInFiltering *bool `yaml:"enableBuiltInFiltering" default:"true"`
```
jaysfx: Don't think it technically needs to be a pointer since you're always setting a default value.

pkg/core/config/config.go
```diff
@@ -228,21 +219,25 @@ func (c *Config) validate() error {
 		return err
 	}
 
+	if !*c.EnableBuiltInFiltering {
+		return errors.New("enableBuiltInFiltering must be true, false is no longer supported")
```
jaysfx: Could say "or removed from the config" as well so people know they can delete it from their config files.

pkg/core/config/config.go
keitwb: It must be a pointer to make it default to `true` since the default value of bool is `false`.  Otherwise the default library doesn't know whether it was explicitly set to `false` or unset.

PR #1165, opened 2020-01-23
support tokens that start with -

dloucasfx:

The current logic in `parse_args_and_install` marks any argument that starts with `-` and not equal to `-h` (the help option) as invalid.
This causes tokens that start with `-` to fail the argument parser.

We have few options to address this behavior:
1) add a key option for `access_token` (--access_token) and either keep or remove the old way of passing the token value without any key 
2) use `--` key option which signals the end of options and disables further option processing

After discussing it with @jaysfx , it made sense to go with 2) as it's a standard way used by many shell commands and we can keep the current behavior intact. 

example from rm help:
```
To remove a file whose name starts with a '-', for example '-foo', use one of these commands:

rm -- -foo
```

I also changed the usage to inform user of this option. 

Note:
My test only covered the install.sh script and made sure that the token gets stored in the appropriate file

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ and use it on your hosts by running:
 
 ```sh
 curl -sSL https://dl.signalfx.com/signalfx-agent.sh > /tmp/signalfx-agent.sh
-sudo sh /tmp/signalfx-agent.sh YOUR_SIGNALFX_API_TOKEN --realm YOUR_SIGNALFX_REALM
+sudo sh /tmp/signalfx-agent.sh --realm YOUR_SIGNALFX_REALM -- YOUR_SIGNALFX_API_TOKEN
 ```
 
 ##### Windows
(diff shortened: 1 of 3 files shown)


dloucasfx:
> I would update the docs on how to use the installer script so that it always uses the `--` form so that it is never a question of whether it would work or not.

@keitwb docs are now updated as you suggested

README.md
```diff
@@ -101,7 +101,7 @@ and use it on your hosts by running:
 
 ```sh
 curl -sSL https://dl.signalfx.com/signalfx-agent.sh > /tmp/signalfx-agent.sh
-sudo sh /tmp/signalfx-agent.sh YOUR_SIGNALFX_API_TOKEN --realm YOUR_SIGNALFX_REALM
+sudo sh /tmp/signalfx-agent.sh -- YOUR_SIGNALFX_API_TOKEN --realm YOUR_SIGNALFX_REALM
```
keitwb: Make this like the quick-install.md one.  Normally the `--` comes _after_ all other arguments, not in the middle.

dloucasfx: done, PR is updated

PR #1166, opened 2020-01-23
Unknown Metric Filtering Enhancements

keitwb:
Adds a field to metadata.yaml called `sendUnknown`, which, if true, will
cause the monitor built-in filtering logic to send any metrics that are
not specified in the metadata.yaml metrics section.  This is useful for
monitors like JMX where the metric names are defined by complex
user-provided configuration and cannot be predicted ahead of time.  This
replaces the half-implemented `metricsExhaustive` field that never
really did anything.

Now all unknown metrics for monitors which do not have `sendUnknown:
true` will be dropped by default.  Previously all unknown metrics from
all monitors would be let through.

Also enhances the built-in filtering to automatically send everything if
there are no metrics defined.  This is largely redundant with the
existing `sendAll` flag but it provides a good safeguard to avoid
confusing filtering behavior if the `sendAll` flag is omitted on
accident.

### DIFF
--- a/cmd/monitorcodegen/genmetadata.tmpl
+++ b/cmd/monitorcodegen/genmetadata.tmpl
@@ -77,7 +77,7 @@ var {{if gt (len $.monitors) 1 -}}
 	MonitorType: "{{.MonitorType}}",
 	DefaultMetrics: {{"defaultMetrics" | formatVariable}},
 	Metrics: {{"metricSet" | formatVariable}},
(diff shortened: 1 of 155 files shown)


keitwb:
> I may have missed it but do we need to add some special text in the docs for monitors that are `sendUnknown: true`? Similar to the text for sendAll?

Added a little note for the monitors where sendUnknown is set.

PR #1168, opened 2020-01-24
Make time.Duration safer in monitor configs

jaysfx:
This makes current usage of time.Duration in monitor configs safer as before
if the user specified an integer it'd be interpreted as an nanoseconds which
is unlikely what people want.

Instead of using time.Duration in a monitor config the developer should use
timeutil.Duration. An integer will be interpreted as seconds (instead of
nanoseconds) and if it's a string it'll be parsed via ParseDuration. Also
if it's a string but it is able to be parsed as an integer we also interpret
it as seconds.

### DIFF
--- a/pkg/monitors/telegraf/monitors/winperfcounters/win_perf_counters_windows.go
+++ b/pkg/monitors/telegraf/monitors/winperfcounters/win_perf_counters_windows.go
@@ -31,7 +31,7 @@ func GetPlugin(conf *Config) (*telegrafPlugin.Win_PerfCounters, error) {
 
 	// Telegraf has a struct wrapper around time.Duration, but it's defined
 	// in an internal package which the gocomplier won't compile from
-	plugin.CountersRefreshInterval.Duration = conf.CountersRefreshInterval.Get()
+	plugin.CountersRefreshInterval.Duration = conf.CountersRefreshInterval.AsDuration()
 
 	// copy nested perf objects
 	for _, perfobj := range conf.Object {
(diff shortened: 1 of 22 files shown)


pkg/monitors/telegraf/monitors/winperfcounters/win_perf_counters_windows.go
```diff
@@ -31,7 +31,7 @@ func GetPlugin(conf *Config) (*telegrafPlugin.Win_PerfCounters, error) {
 
 	// Telegraf has a struct wrapper around time.Duration, but it's defined
 	// in an internal package which the gocomplier won't compile from
-	plugin.CountersRefreshInterval.Duration = conf.CountersRefreshInterval * time.Millisecond
+	plugin.CountersRefreshInterval.Duration = conf.CountersRefreshInterval.Get()
```
jaysfx: @codesmith14 any idea why it was being multiplied by time.Millisecond here before?

pkg/utils/timeutil/duration.go
```diff
@@ -0,0 +1,68 @@
...
+// Duration is a wrapped time.Duration that supports durations as integers
+// or a ParseDuration string. If it's an integer it is interpreted in seconds instead
+// of being directly cast to time.Duration which would normally make it in nanoseconds.
+type Duration time.Duration
+
+// ErrInvalidDuration is returned when the duration can't be interpreted
+var ErrInvalidDuration = errors.New("the duration must be a string with time unit specified or an integer as seconds")
+
+// Get returns the underlying time.Duration type
+func (d Duration) Get() time.Duration {
```
keitwb: `Get` implies this is some kind of container for a duration, when it is actually just a duration.  Normally for conversion funcs I have seen them named `As*`.

PR #1169, opened 2020-01-24
Add K8s v1.17.0 circleci test job

jchengsfx:


.circleci/config.yml
```diff
@@ -536,21 +542,9 @@ workflows:
           requires:
             - build
             - helm_check
-      - k8s_integration_tests:
```
keitwb: Let's keep 1.13

PR #1174, opened 2020-01-27
jvm groovy script

jaysfx:


docs/monitors/jmx.md
```diff
@@ -115,6 +115,37 @@ output.sendDatapoints([
 
 ```
 
+This script collects the same JMX metrics as emitted by default from the [genericjmx monitor](collectd-genericjmx.md).
+It can be used as a starting point for collecting additional JMX metrics.
+
+```groovy
+Map dims(Map args) {
+    // so that dashboard will import
+    d = ["plugin": "GenericJMX"]
```
keitwb: This is a lie.

docs/monitors/jmx.md
```diff
@@ -115,6 +115,37 @@ output.sendDatapoints([
...
+    d.putAll(args ?: [:])
+    d
+}
+dps = [
+    util.makeGauge("gauge.jvm.threads.count", util.queryJMX("java.lang:type=Threading").first().ThreadCount, dims()),
```
keitwb: Let's not encourage people to use the naming conventions of collectd.  Just drop the `gauge.`

docs/monitors/jmx.md
```diff
@@ -115,6 +115,37 @@ output.sendDatapoints([
...
+    util.makeGauge("gauge.loaded_classes", util.queryJMX("java.lang:type=ClassLoading").first().LoadedClassCount, dims()),
+]
+mem = util.queryJMX("java.lang:type=Memory").first()
+memoryPool = util.queryJMX("java.lang:type=MemoryPool,*")
+for (i in ["committed", "init", "max", "used"]) {
+    metric = "jmx_memory.$i"
+    dps << util.makeGauge(metric, mem.HeapMemoryUsage."$i", dims(plugin_instance: "memory-heap"))
```
keitwb: Please no more `plugin_instance`.  This is one of the worst things about collectd metrics in that we have no idea what the "plugin_instance" is actually referring to.  Use `pool_name` or something a million times more descriptive.

jaysfx:
> Let's not perpetuate so many bad practices from collectd. We can make new content if that is such an issue.

Yeah we can fix it we just need to update builtin content to have an additional import selector and categorize the new metric names appropriately.

jaysfx:
So while I agree on the above I think having this even in its current state is a net positive. The only alternative is to abandon the change since updating the builtin content isn't a priority right now. (This is just something I had laying around and wanted to get off my machine). I can checkin just the file maybe?

keitwb:
Yeah if you want a reference to point people to as a migratory step to get off of `collectd/genericjmx`, then just check in the groovy script somewhere similar to the python scripts we have as examples.  But I would also check in another script with the "modernized" version of it along side the "legacy" one and make it clear that we don't encourage the collectd patterns.

PR #1175, opened 2020-01-28
Add elasticsearch query monitor

asuresh4:


pkg/monitors/elasticsearch/query/request.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		agg.AggregationProps[k] = v
+	}
+
+	return nil
+}
+
+// Returns a response for a given elasticsearch query
+func (es ESQueryHTTPClient) makeHTTPRequestFromConfig(mq *metricQuery) ([]byte, error) {
```
keitwb: This is very confusing going from declaring methods above here for the `aggregationInfo` type and all of the sudden switching to the `ESQueryHTTPClient` type.  Group methods together with their type declaration in the same files.

pkg/monitors/elasticsearch/query/response.go
```diff
@@ -0,0 +1,174 @@
...
+	Values interface{} `json:"values,omitempty"`
+	// Map of sub aggregations with aggregationInfo names, includes both bucket
+	// and metric aggregations
+	SubAggregations map[string]aggregationResponse `json:"-"`
+	// Non nil for multi-value bucket aggregations
+	Buckets []*bucketResponse `json:"buckets,omitempty"`
+	// This field only exists to handle cases where buckets are returned as a
+	// map from the key to the object itself. In such cases the "buckets" field
+	// will be post processed and put into the "Buckets" field of this struct
+	BucketsMap map[string]*bucketResponse `json:"buckets,omitempty"`
```
keitwb: What exactly does the json module do when you have two separate struct fields with the same json key `buckets`?  In yaml this will cause a panic, but not sure how json handles it.

pkg/monitors/elasticsearch/query/response.go
```diff
@@ -0,0 +1,174 @@
...
+
+	*agg = aggregationResponse(temp)
+	agg.SubAggregations = map[string]aggregationResponse{}
+
+	var m map[string]interface{}
+	if err := json.Unmarshal(b, &m); err != nil {
+		return err
+	}
+
+	t := map[string]interface{}{}
```
keitwb: `t` is pretty cryptic.

pkg/monitors/elasticsearch/query/response.go
```diff
@@ -0,0 +1,174 @@
...
+	OtherValues map[string]interface{} `json:"-"`
+}
+
+// Represents a slice ("bucket") of ES documents
+type bucketResponse struct {
+	// Value uniquely identifying a bucket for a given bucket aggregationInfo
+	Key interface{} `json:"key,omitempty"`
+	// Reported in cases where buckets are grouped by a non string key,
+	// for instance time
+	KeyAsString string `json:"key_as_string,omitempty"`
```
keitwb: Just make this a method called `KeyAsString` that does the conversion instead of storing it on the struct directly.  This doesn't actually come from a real ES response as far as I know.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+			return err
+		}
+
+		metricQueryInfos[i] = &metricQueryInfo{
+			interval: mq.QueryInterval.Seconds(),
+			aggsMeta: aggsMeta,
+		}
+	}
+
+	// Since metric queries can have their own interval, set them running on separate go routines
```
keitwb: Don't let them have their own interval -- it invalidates the meaning of `intervalSeconds` on the monitor itself.  If people want differing intervals, they can make multiple instances of this monitor.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+		}, time.Duration(metricQueryInfos[i].interval)*time.Second)
+	}
+	return nil
+}
+
+func (m *Monitor) sendDatapoints(dps []*datapoint.Datapoint) {
+	// This is the filtering in place trick from https://github.com/golang/go/wiki/SliceTricks#filter-in-place
+	n := 0
+	for i := range dps {
+		if dps[i] == nil {
```
keitwb: You should really fix the thing that is inserting nil datapoints so that it doesn't send nil datapoints.  Then you can get rid of this method and just use `m.Output.SendDatapoints` directly.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+// Conceptually the data field in search queries define  bucket aggregations
+// that determine "buckets" (some group of elasticsearch documents that match
+// the search query) and metric aggregations that compute metrics from each "bucket"
+type metricQuery struct {
+	// Index that's being queried
+	Index string `yaml:"index" validate:"required"`
+	// Lucene search query which the elasticsearch documents should match
+	LuceneSearchQuery string `yaml:"luceneSearchQuery" validate:"required"`
+	// Query Interval
+	QueryInterval time.Duration `yaml:"queryInterval" default:"30s"`
```
keitwb: Get rid of this field and rely on the monitor's `IntervalSeconds` config.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Conceptually the data field in search queries define  bucket aggregations
+// that determine "buckets" (some group of elasticsearch documents that match
+// the search query) and metric aggregations that compute metrics from each "bucket"
+type metricQuery struct {
+	// Index that's being queried
+	Index string `yaml:"index" validate:"required"`
+	// Lucene search query which the elasticsearch documents should match
+	LuceneSearchQuery string `yaml:"luceneSearchQuery" validate:"required"`
```
keitwb: Is this still needed?

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+	// Query Interval
+	QueryInterval time.Duration `yaml:"queryInterval" default:"30s"`
+	// HTTP Post data field for aggregations
+	HTTPPostData string `yaml:"httpPostData" validate:"required"`
```
keitwb: The fact that it is HTTP post data is an implementation detail of the protocol used by the agent to communicate with elasticsearch.  This is really the `ElasticsearchQuery`.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+
+	MetricQueries []*metricQuery `yaml:"metricQueries" validate:"required"`
+}
+
+// Conceptually the data field in search queries define  bucket aggregations
+// that determine "buckets" (some group of elasticsearch documents that match
+// the search query) and metric aggregations that compute metrics from each "bucket"
+type metricQuery struct {
+	// Index that's being queried
+	Index string `yaml:"index" validate:"required"`
```
keitwb: What if the user wants to run an aggregation over multiple indexes?  Can they put `_all` here and make that work?  If so, that should be documented.

pkg/monitors/elasticsearch/query/response.go
asuresh4: With `json` from what I've seen depending on the type of `buckets`, it intelligently does the right thing. But I noticed this is causing a lint failure (`structtag: struct field BucketsMap repeats json tag "buckets" also at response.go:25 (govet)`). I was thinking of handling `buckets` as a non standard field in the switch case to work around that. Do you think that's a good approach?

keitwb: You can also make `buckets` deserialized as a `interface{}` and then just have a special method that figures out the right form it has an converts it to the canonical slice format that you do the processing on.  You can go through json again or just pull out stuff directly from the map.

pkg/monitors/elasticsearch/query/response.go
asuresh4: `KeyAsString` is returned by ES in cases where the bucket is identified by a non-string value like a date. Since these keys are dimensions for us, I'm considering looking at the type of `Key` and proceed only if the key is of string type. What do you think?

pkg/monitors/elasticsearch/query/monitor.go
asuresh4: Ok will remove 👍

pkg/monitors/elasticsearch/query/monitor.go
asuresh4: I don't think there'll be nil values, I'll make sure and remove this logic from here

pkg/monitors/elasticsearch/query/monitor.go
asuresh4: This is actually redundant. A lucene query string can be provided to match documents on (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-uri-request.html) or looks like this query can be specified in the request body too along with the aggregations (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html).

pkg/monitors/elasticsearch/query/monitor.go
asuresh4: Yes, this is actually possible. Will handle that case.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+	monitors.Register(&monitorMetadata, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+type metricQueryInfo struct {
+	interval float64
+	aggsMeta map[string]*AggregationMeta
+}
+
+// Configure monitor
+func (m *Monitor) Configure(c *Config) error {
```
pmcollins: What do you think about renaming the config variable `conf`, `cfg`, or `config`.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,136 @@
...
+	m.logger = utils.NewThrottledLogger(log.WithFields(log.Fields{"monitorType": monitorType}), 20*time.Second)
+
+	httpClient, err := c.HTTPConfig.Build()
+	if err != nil {
+		return err
+	}
+	esClient := NewESQueryClient(c.Host, c.Port, c.HTTPConfig.Scheme(), httpClient)
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+
+	metricQueryInfos := make([]*metricQueryInfo, len(c.MetricQueries))
```
pmcollins: Looks like this code down to line 88 could be extracted into a function that takes a config and returns a `[]*metricQueryInfo`. A matter of taste so up to you.

asuresh4: Ended up removing this logic based on comments from Ben about supporting one query per instance of monitor

pkg/monitors/elasticsearch/query/monitor.go
asuresh4: Removed this config

pkg/monitors/elasticsearch/query/monitor.go
keitwb: Yeah, one query can support multiple aggregations so one should be enough.

pkg/monitors/elasticsearch/query/request.go
```diff
@@ -0,0 +1,117 @@
...
+// See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html
+// for details about various type of bucket and metric aggregations
+type aggregationInfo struct {
+	// Holds aggregationInfo type and properties specific to the aggregation
+	AggregationProps map[string]interface{} `json:"-"`
+	// Sub aggregations within the parent aggregation
+	SubAggregations map[string]aggregationInfo `json:"aggs,omitempty"`
+}
+
+type AggregationMeta struct {
```
pmcollins: Is it preferable to use a struct with one string field rather than just using a string

pkg/monitors/elasticsearch/query/request.go
```diff
@@ -0,0 +1,117 @@
...
+
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return getAggregationsMetaHelper(aggsToType), nil
+}
+
+func getAggregationsWithType(aggName string, agg aggregationInfo, out map[string]string) error {
```
pmcollins: I want to get sense of how this and other functions work. Unit tests would go a long way towards helping readers and reviewers, in addition to the other benefits they provide.

keitwb: This isn't getting merged without some pretty heavy unit tests of the parsing, as well as a few integration tests proving the whole process.

pkg/monitors/elasticsearch/query/response.go
asuresh4: I've updated to look at buckets only when the key is a string

pkg/monitors/elasticsearch/query/request.go
asuresh4: yes, those unit tests are on its way!

pkg/monitors/elasticsearch/query/datapoints.go
```diff
@@ -0,0 +1,253 @@
...
+
+			out = append(out, dp)
+		}
+	}
+
+	return out
+}
+
+// Collect datapoint from "percentiles" metric aggregation
+func getDatapointsFromPerciltes(aggRes *aggregationResponse, dims map[string]string) []*datapoint.Datapoint {
```
pmcollins: typo

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+        "avg_cpu_utilization": {
+          "avg": {
+            "field": "cpu_utilization"
+          }
+        }
+      }
+    }
+```
+
+The `elasticsearchRequest` take in a request in the format specified
```
keitwb: Point out that it takes the request as a **string** . If you remove the `|` from your example config above it will be valid YAML since JSON is a subset of YAML.  Of course, agent config validation would catch it.

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+   '@timestamp':1580321240579
+}
+```
+
+This query returns the average value of `cpu_utilization` across all documents with a `@timestamp`
+in the last five minutes. This value is metricized to the following form in SignalFx -
+
+```
+{
+metric_name: avg,
```
keitwb: Shouldn't this be `avg_cpu_utilization` since that is the aggregation name?  `avg` is the metric aggregation type -- it is way too generic and meaningless to be the metric name.

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+dimensions:
+  index: <name_of_index>
+  metric_aggregation_name: avg_cpu_utilization
```
keitwb: This should be something like `metric_aggregation_type: avg` and the agg name be the metric name.

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+            "field": "cpu_utilization"
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+The above query will result in a metric called `avg` (aggrregation type defined by Elasticsearch query syntax)
```
keitwb: Yeah swap metric name and the dimension as described above.

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+      }
+    ]
+  }
+}
+...
+```
+
+In this case, each bucket will result 5 metrics -
+
+  1. stats.count
```
keitwb: Should be `cpu_usage_stats.count`, etc.

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+| `useHTTPS` | no | `bool` | If true, the agent will connect to the exporter using HTTPS instead of plain HTTP. (**default:** `false`) |
+| `skipVerify` | no | `bool` | If useHTTPS is true and this option is also true, the exporter's TLS cert will not be verified. (**default:** `false`) |
+| `caCertPath` | no | `string` | Path to the CA cert that has signed the TLS cert, unnecessary if `skipVerify` is set to false. |
+| `clientCertPath` | no | `string` | Path to the client TLS cert to use for TLS required connections |
+| `clientKeyPath` | no | `string` | Path to the client TLS key to use for TLS required connections |
+| `host` | **yes** | `string` |  |
+| `port` | **yes** | `string` |  |
+| `index` | no | `string` | Index that's being queried. If none is provided, given query will be applied across all indexes. To apply the search query to multiple indices, provide a comma separated list of indices (**default:** `_all`) |
+| `elasticsearchRequest` | **yes** | `string` | Takes in an elasticsearch request body search request. See [here] (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html) for details. |
```
keitwb:
```suggestion
| `elasticsearchRequest` | **yes** | `string` | Takes in an Elasticsearch request body search request. See [here] (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html) for details. |
```

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+          "value" : 49.99789885523837
+        }
+      }
+    ]
+  }
+}
+...
+```
+
+2. Extended Stats Metric aggregation as a sub aggregation of Terms aggregation
```
keitwb:
```suggestion
2. `extended_stats` Metric aggregation as a sub aggregation of Terms aggregation
```

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+3. Bucket aggregations that do not have any Metric aggregations as sub aggregations will be
+transformed to a metric called `doc_count` and will have `bucket_aggregation_name` dimension
+apart from the `key` of each bucket.
+
+**Note**: Since Bucket aggregations determine dimensions in SignalFx, in most cases Bucket aggregations
+should be performed on `string` fields that represent a slice of the data from Elasticsearch.
+
+## Examples
+
+1. Average Metric aggregation as a sub aggregation of Terms aggregation
```
keitwb:
```suggestion
1. `avg` metric aggregation as a sub aggregation of `terms` aggregation
```

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,282 @@
...
+   'memory_utilization':94,
+   'host':'helsniki',
+   'service':'android',
+   'container_id':'macbook',
+   '@timestamp':1580321240579
+}
+```
+
+This query returns the average value of `cpu_utilization` across all documents with a `@timestamp`
+in the last five minutes. This value is metricized to the following form in SignalFx -
```
keitwb:
```suggestion
in the last five minutes. This value is metricized to the following form in SignalFx:
```

pkg/monitors/elasticsearch/query/request.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return nil
+}
+
+func getAggregationType(agg aggregationInfo) (string, error) {
+	for k := range agg.AggregationProps {
+		return k, nil
+	}
+
+	return "", errors.Errorf("Unable to determine type for %v", agg)
```
keitwb: Use `fmt.Errorf`.

tests/monitors/elasticsearch_query/elasticsearch_query_test.py
```diff
@@ -0,0 +1,283 @@
...
+
+from datetime import datetime
+from elasticsearch import Elasticsearch
+from random import randint
+
+from tests.helpers.agent import Agent
+from tests.helpers.assertions import has_datapoint_with_all_dims, has_datapoint_with_metric_name, http_status
+from tests.helpers.util import container_ip, run_service, wait_for
+
+VERSIONS = ["5.0.0", "5.5.0", "6.2.0", "6.4.2", "7.0.0", "7.5.1"]
```
keitwb: Let's just test the oldest supported and the newest.  Testing 6 versions is going to bog down the CI time and is kind of overkill.

tests/monitors/elasticsearch_query/elasticsearch_query_test.py
```diff
@@ -0,0 +1,283 @@
...
+        check_service_status(host)
+        write_data(host, version)
+
+        query = {"query": {"match_all": {}}, "aggs": {"host_name": {"terms": {"field": "host"}}}}
+
+        agent_config = AGENT_CONFIG_TEMPLATE.format(host=host, index="metrics", query=json.dumps(query))
+
+        with Agent.run(agent_config) as agent:
+            assert wait_for(
+                p(has_datapoint_with_metric_name, agent.fake_services, "elasticsearch_query.doc_count")
```
keitwb: Use `p(has_datapoint, agent.fake_services, metric_name="elasticsearch_query.doc_count", dimensions={"index": "metrics", "bucket......})

docs/monitors/elasticsearch-query.md
asuresh4: Updated as suggested

pkg/monitors/elasticsearch/query/datapoints.go
```diff
@@ -0,0 +1,253 @@
...
+
+	for k, v := range aggsResult {
+		// each aggregation at the highest level starts with an empty set of dimensions
+		out = append(out, collectDatapointsHelper(k, *v, aggsMeta, sfxDimensions)...)
+	}
+
+	return out
+}
+
+func collectDatapointsHelper(aggName string, aggRes aggregationResponse,
```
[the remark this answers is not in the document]
asuresh4: Yes, that's a good point. Having it like this, in a way keeps the context of each level of recursion to an instance of `dpCollector`. Updated based on your suggestion.

pkg/monitors/elasticsearch/query/response.go
```diff
@@ -0,0 +1,186 @@
...
+type aggregationResponse struct {
+	// Number of documents backing the aggregationInfo
+	DocCount *int64 `json:"doc_count,omitempty"`
+	// Non nil for single value metric aggregations
+	Value interface{} `json:"value,omitempty"`
+	// Non nil for multi-value metrics aggregations
+	Values interface{} `json:"values,omitempty"`
+	// Map of sub aggregations with aggregationInfo names, includes both bucket
+	// and metric aggregations
+	SubAggregations map[string]*aggregationResponse `json:"-"`
```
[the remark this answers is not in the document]
asuresh4: Incorporated your suggestion about having types for buckets and aggregations. For the purposes in this monitor I don't know if it would have made a difference but I think in general mapping to references allow for a easier way to modify entries in a map.

tests/monitors/elasticsearch_query/elasticsearch_query_test.py
```diff
@@ -0,0 +1,265 @@
...
+                    ), "Didn't get elasticsearch-query datapoints"
+
+
+def write_data(host, version, num_docs=10):
+    """
+    Populates ES with mock data
+    """
+    es = Elasticsearch(hosts=[host])
+
+    doc_type = "doc"
```
keitwb: I think this needs to be `_doc` or just remove the `doc_type` shell object from the mapping definition.

asuresh4: `_doc` is not a supported in 5.0.0 since it starts with an underscore. That's how I ended up with this. I'll try a different mapping to work around the issue.

asuresh4: or actually this might be because the ``elasticsearch7.0.0` client is incompatible with ES 5.0.0. I'll try just using the REST API

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,285 @@
...
+  index: <name_of_index>
+  metric_aggregation_type: avg
+}
+```
+
+## Data Model Transformation
+
+Understanding how this monitor transforms Elasticsearch responses to SignalFx datapoints.
+
+At high level this monitor metricizes responses of following types -
```
keitwb:
```suggestion
At high level this monitor metricizes responses of the following types:
```

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,285 @@
...
+metric_name: elasticsearch_query.avg_cpu_utilization,
+dimensions:
+  index: <name_of_index>
+  metric_aggregation_type: avg
+}
+```
+
+## Data Model Transformation
+
+Understanding how this monitor transforms Elasticsearch responses to SignalFx datapoints.
```
keitwb: This is not a complete sentence.

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,285 @@
...
+   'cpu_utilization':87,
+   'memory_utilization':94,
+   'host':'helsniki',
+   'service':'android',
+   'container_id':'macbook',
+   '@timestamp':1580321240579
+}
+```
+
+This query returns the average value of `cpu_utilization` across all documents with a `@timestamp`
```
keitwb:
```suggestion
The query specified in `elasticsearchRequest` returns the average value of `cpu_utilization` across all documents with a `@timestamp`
```

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,285 @@
...
+      }
+    ]
+  }
+}
+...
+```
+
+In this case, each bucket will result 5 metrics -
+
+  1. elasticsearch_query.cpu_usage_stats.count
```
keitwb: Use backticks.

docs/monitors/elasticsearch-query.md
```diff
@@ -0,0 +1,285 @@
...
+
+An simple configuration looks like the following:
+
+```yaml
+monitors:
+- type: elasticsearch-query
+ host: localhost
+ port: 9200
+ index: <name_of_index>
+ elasticsearchRequest: |
```
keitwb: You would probably want to set `intervalSeconds: 300` on this example, otherwise the CPU utilization will be overlapping.  Or change 5m to 20s.

pkg/monitors/elasticsearch/query/datapoints.go
```diff
@@ -0,0 +1,286 @@
...
+		return dpC.collectDatapointsFromMetricAggregation()
+	}
+
+	// Recursively collect all datapoints from buckets at this level
+	for _, b := range dpC.aggRes.Buckets {
+		key, ok := b.Key.(string)
+
+		if !ok {
+			log.WithFields(
+				log.Fields{
```
keitwb: Set a logger on `dpCollector` that has a default field of `monitorType: elasticsearch-query`.  See other monitors for an example.

keitwb: This makes it easier to identify which monitor is emitting the warning if you have a lot configured.

pkg/monitors/elasticsearch/query/monitor.go
```diff
@@ -0,0 +1,95 @@
...
+	logger *utils.ThrottledLogger
+}
+
+func init() {
+	monitors.Register(&monitorMetadata, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Configure monitor
+func (m *Monitor) Configure(config *Config) error {
+	m.logger = utils.NewThrottledLogger(log.WithFields(log.Fields{"monitorType": monitorType}), 20*time.Second)
```
keitwb: Yeah pass down this logger per the comment above.

PR #1179, opened 2020-01-29
Install windows python with nuget

jchengsfx:
Use nuget to install a stand-alone version of python since the windows
installer fails if python is already installed on the build system.


keitwb:
Can you confirm that the bundled Python is still capable of doing `pip installs`?

keitwb:
Ok, I cherry-picked this commit into the 5.0-staging branch's Python 3.8 upgrade commit, so this PR can be closed.

keitwb:
gg

PR #1180, opened 2020-01-31
CloudFoundry Firehose Nozzle Monitor

keitwb:
Supports reading from the RLP Gateway and sending gauge and counter
metrics.


pkg/monitors/cloudfoundry/client.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"code.cloudfoundry.org/go-loggregator/rpc/loggregator_v2"
+	"github.com/cloudfoundry/go-loggregator"
+	"github.com/signalfx/golib/v3/datapoint"
+	"github.com/signalfx/signalfx-agent/pkg/core/common/dpmeta"
+	"github.com/signalfx/signalfx-agent/pkg/utils"
+	"github.com/sirupsen/logrus"
+)
+
+const defaultShardID = "signalfx-nozzle"
```
asuresh4: Looks like this default does not match the default config, `signalfx_nozzle`. Is that intentional?

pkg/monitors/cloudfoundry/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+  doc: |
+    This is a CloudFoundry [firehose
+    nozzle](https://docs.pivotal.io/tiledev/2-8/nozzle.html) by connecting to
+    the Cloud Foundry Reverse Log Proxy (RLP) Gateway that feeds metrics from
+    the Loggregator. This uses the new RLP Gateway model that was [introduced
+    in Pivotal Cloud Foundry (PCF)
+    2.4](https://docs.pivotal.io/tiledev/2-4/release-notes.html#log-api), so it
+    will not work with older releases.
+
+    It supports `gauge` and `counter` metrics at this time.  All of the `tags` in
```
asuresh4: I think we should clarify counter here is a cumulative counter

pkg/monitors/cloudfoundry/client.go
asuresh4: Do we actually need it, seems like it's always overwritten by the value from the config

pkg/monitors/cloudfoundry/conversion.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	case *loggregator_v2.Envelope_Timer:
+	case *loggregator_v2.Envelope_Event:
+	default:
+		return nil, fmt.Errorf("cannot convert envelope %v to SignalFx datapoints", spew.Sdump(env))
+	}
+
+	var dps []*datapoint.Datapoint
+	for name, val := range namesToValues {
+		dps = append(dps, datapoint.New(cleanupName(name), dims, datapoint.NewFloatValue(val), metricType, time.Unix(0, env.Timestamp)))
```
asuresh4: What do you think about adding a prefix name to all metrics (something like `pcf_firehose`) to make it easier to search?

keitwb: The metrics will be prefixed with the name of the source/origin within CloudFoundry that emitted them.  We generally should avoid metric prefixes that are something really generic like the platform or mechanism by which they were derived (e.g. the firehose).

If you want to make them easy to search, you could add a global dimension in your agent config that gave all of them a common fixed dimension to search on.

pkg/monitors/cloudfoundry/client.go
keitwb: It's good to have to decouple the client from the code that uses it (e.g. in case that other code gets refactored and that assumption is no longer true that it is always overwritten).

deployments/cloudfoundry/runtime-config-tile/tile.yml
```diff
@@ -0,0 +1,66 @@
...
+    optional: true
+
+packages: []
+
+runtime_configs:
+  - name: signalfx-agent
+    runtime_config:
+      releases:
+        - name: signalfx-agent
+          version: 4.19.0-dev18
```
asuresh4: is this the right version we want to point to?

deployments/cloudfoundry/bosh/tile.yml
```diff
@@ -0,0 +1,159 @@
+---
+name: signalfx-monitoring
+icon_file: resources/icon.png
```
asuresh4: the icon being used here is the old one, we should update it.

deployments/cloudfoundry/runtime-config-tile/tile.yml
keitwb: I have to merge this first and do a release of the agent with the firehose nozzle, and then I can fix up all of the release versions.

pkg/monitors/cloudfoundry/metadata.yaml
keitwb: Ok, fixed.

deployments/cloudfoundry/bosh/tile.yml
keitwb: Updated to the latest used in the Grafana plugin.

PR #1183, opened 2020-02-03
Updating dimensions on kubernetes-events monitor

mstumpfx:
Trying to get parity with the `kubernetes-cluster` monitor, we
should update our dimensions emitted by the events monitor.
- **Break change**: No longer send host specfic dimensions along with the event (eg Host
dimension)
- If the event is about a node, `kubernetes_node` and
`kubernetes_node_uid` will be emitted

### DIFF
--- a/pkg/monitors/kubernetes/events/events.go
+++ b/pkg/monitors/kubernetes/events/events.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/signalfx/golib/v3/event"
+	"github.com/signalfx/signalfx-agent/pkg/core/common/dpmeta"
 	"github.com/signalfx/signalfx-agent/pkg/core/common/kubernetes"
 	"github.com/signalfx/signalfx-agent/pkg/core/config"
 	"github.com/signalfx/signalfx-agent/pkg/monitors"
@@ -151,6 +152,7 @@ func (m *Monitor) shouldSendEvent(ev *v1.Event) bool {
 func (m *Monitor) handleNewEvent(ev *v1.Event) {
 	if m.shouldSendEvent(ev) {
 		sfxEvent := k8sEventToSignalFxEvent(ev)
+		sfxEvent.Properties[dpmeta.NotHostSpecificMeta] = true
 		m.Output.SendEvent(sfxEvent)
 	}
 }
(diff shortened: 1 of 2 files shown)


mstumpfx:
Should we add `kubernetes_workload_name` as well?

pkg/monitors/kubernetes/events/events.go
```diff
@@ -151,22 +152,27 @@ func (m *Monitor) shouldSendEvent(ev *v1.Event) bool {
...
 		sfxEvent := k8sEventToSignalFxEvent(ev)
+		sfxEvent.Properties[dpmeta.NotHostSpecificMeta] = true
 		m.Output.SendEvent(sfxEvent)
 	}
 }
 
 func k8sEventToSignalFxEvent(ev *v1.Event) *event.Event {
 	dims := map[string]string{
-		"kubernetes_kind":      ev.InvolvedObject.Kind,
+		"kubernetes_workload":  ev.InvolvedObject.Kind,
```
keitwb: Isn't this TitleCase whereas our agent's property is lowercase?

keitwb:
> Should we add `kubernetes_workload_name` as well?

No, that seems unnecessary.

pkg/monitors/kubernetes/events/events.go
keitwb: This seems semantically questionable since our agent uses this property to describe the workload of a pod, whereas here it would be describing the type of the workload that the event pertains to.

mstumpfx: In the `kubernetes-cluster` monitor though we are talking about metrics, so we wanted the `workload` to be set to whatever object is running the pod. 
For events, they are always about the object themselves. StatefulSet's will have a "Created" event, which will create a pod and we will then get another "Created" event for a pod.

If we want events to have the same "workload" meaning as our metrics, we would need to extend the cluster cache to this monitor too. 
Maybe we could have an external "Cluster State" sort of singleton running in of the elected master node that could be used for these types of linking objects around issues.

mstumpfx:
After speaking with RT, we decided we don't need this linking of objects for now. I reverted the change to make `kubernetes_kind` --> `kubernetes_workload` because of the semantic issue with using workload.

PR #1188, opened 2020-02-06
Add support for custom HTTP headers

dloucasfx:
This PR adds support to custom HTTP Headers.
The client will pick up users defined custom headers.

example YAML:
```
httpHeaders:
  Proxy-Authenticate: basic
  Accept: application/vnd.api+json, image/png
```

### DIFF
--- a/docs/monitors/coredns.md
+++ b/docs/monitors/coredns.md
@@ -45,6 +45,7 @@ Configuration](../monitor-config.md#common-configuration).**
 | `username` | no | `string` | Basic Auth username to use on each request, if any. |
 | `password` | no | `string` | Basic Auth password to use on each request, if any. |
 | `useHTTPS` | no | `bool` | If true, the agent will connect to the exporter using HTTPS instead of plain HTTP. (**default:** `false`) |
+| `httpHeaders` | no | `map of strings` | A map of key=message-header and value=header-value. Comma separated multiple values for the same message-header is supported. |
 | `skipVerify` | no | `bool` | If useHTTPS is true and this option is also true, the exporter's TLS cert will not be verified. (**default:** `false`) |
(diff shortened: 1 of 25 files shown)


jaysfx:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

Given that 99% of the time people are probably going to be setting the header with a single value _and_ that it's possible to represent multiple values with a comma (can verify this in the tests) I think it might be more user-friendly to just take a `map[string]string`.

jaysfx:
@keitwb what do you think about ^?

keitwb:
> https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
> 
> > Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.
> 
> Given that 99% of the time people are probably going to be setting the header with a single value _and_ that it's possible to represent multiple values with a comma (can verify this in the tests) I think it might be more user-friendly to just take a `map[string]string`.

Yes, let's just take `map[string]string` since the values can always be combined into the comma separated form per the standard.

dloucasfx:
PR updated with the above recommendation

PR #1191, opened 2020-02-10
Jenkins improvements - WIP

dloucasfx:
This is still Work in progress cc: @jaysfx 

This change also depends on improvements done to our fork of the gojenkins librariry
https://github.com/signalfx/gojenkins/commits/SFEnhancements

Still need to:
1- create integration test
2- add pipeline support


pkg/monitors/jenkins/client/client.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return err
+	}
+	defer res.Close()
+
+	responseData, err := ioutil.ReadAll(res)
+	if err != nil {
+		return err
+	}
+	*result = string(responseData)
```
keitwb: If you have a fixed non-interface type of the result object (string), then you don't really need to do the trick where you pass it by pointer like with the JSON decoding, you can just return `(string, error)`.

pkg/monitors/jenkins/jobmetrics.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, jobBuild := range buildsResp.Builds {
+		if jobBuild.Building {
+			continue
+		}
+		buildCompletion := jobBuild.Timestamp + jobBuild.Duration
+		if buildCompletion > lastBuild {
+			if buildCompletion > m.lastBuildSent[jobName] {
+				m.lastBuildSent[jobName] = buildCompletion
+			}
+			dimensions["Result"] = jobBuild.Result
```
keitwb: What are the possible values of `jobBuild.Result`?

pkg/monitors/jenkins/jobmetrics.go
```diff
@@ -0,0 +1,106 @@
...
+	err := job.GetBuildsFields([]string{"timestamp", "result", "duration", "building"}, &buildsResp)
+	if err != nil {
+		return nil, err
+	}
+	if len(buildsResp.Builds) == 0 {
+		return nil, fmt.Errorf("no builds for job %s", jobName)
+	}
+
+	var dimensions = make(map[string]string)
+	dimensions["Job"] = jobName
```
keitwb: dimension names should be snake_case, lowercase.

pkg/monitors/jenkins/jobmetrics.go
```diff
@@ -0,0 +1,106 @@
...
+
+	lastBuild := m.lastBuildSent[jobName]
+	for _, jobBuild := range buildsResp.Builds {
```
keitwb: It isn't super clear what this logic is doing.  It is getting the last build and generating a metric about how long it took?  Maybe split stuff out to well-named functions to make it clearer.

pkg/monitors/jenkins/nodemetrics.go
```diff
@@ -0,0 +1,72 @@
...
+	err := jkClient.FetchText(fmt.Sprintf(pingEndPoint, conf.MetricsKey), &result)
+	if err != nil {
+		return nil, err
+	}
+	var val int64
+	if strings.TrimSpace(result) == "pong" {
+		val = 1
+	}
+	var dps []*datapoint.Datapoint
+	dps = append(dps, datapoint.New(jenkinsNodeOnlineStatus, nil, datapoint.NewIntValue(val), metricSet[jenkinsNodeOnlineStatus].Type, time.Time{}))
```
keitwb: Seems like we could have a simple helper function in the genmetadata.go that makes a datapoint with the signature `(metricName string, val int64, dims map[string]string)` and could look up the type within.  Would clear some of these dp creation stuff up a lot.

pkg/monitors/jenkins/slavestatus.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	metric := groupMetricsMap[groupSlaveStatus][0]
+
+	for _, node := range nodes {
+		if node.Raw.Class == "hudson.slaves.SlaveComputer" {
+			val, err := datapoint.CastMetricValueWithBool(!node.Raw.Offline)
+			if err != nil {
+				logger.Warnf("Skipping metric: failed to cast value for key %s and value %v : %v", metric, val, err)
+			}
+			dps = append(dps, datapoint.New(metric, map[string]string{"Slave": node.Raw.DisplayName}, val, metricSet[metric].Type, time.Time{}))
```
keitwb: Same thing about lowercase, snake_case dimension names.

pkg/monitors/jenkins/client/client.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, fmt.Errorf("could not build request for url %s: %v", url, err)
+	}
+	res, err := c.Do(req)
+
+	if err != nil {
+		return nil, fmt.Errorf("could not get url %s: %v", url, err)
+	}
+	if res.StatusCode != 200 {
+		res.Body.Close()
+		return nil, fmt.Errorf("received status code that's not 200: %s, url: %s", res.Status, url)
```
keitwb: Show the response body in the error message. This can save a lot of time debugging.  Also I think you have to fully read the `resp.Body` reader for it to reuse connections properly.

pkg/monitors/jenkins/client/client.go
```diff
@@ -0,0 +1,91 @@
...
+
+// fetchResponse takes a URL and HTTP client and returns a reader
+// caller should always close the reader
+func fetchResponse(url string, c *http.Client) (io.ReadCloser, error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not build request for url %s: %v", url, err)
+	}
+	res, err := c.Do(req)
+
```
keitwb: No blank line here but a blank line above the `c.Do(req)` call.

pkg/monitors/jenkins/jenkins.go
```diff
@@ -0,0 +1,121 @@
...
+	httpClient, err := conf.HTTPConfig.Build()
+	if err != nil {
+		return err
+	}
+	jkClient, err := jc.NewJenkinsClient(conf.Host, conf.MetricsKey, conf.Port, conf.Scheme(), httpClient)
+	if err != nil {
+		return err
+	}
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+	m.lastBuildSent = make(map[string]int64)
```
keitwb: This makes me wary.  See if the metrics can be sent in such a way (e.g. a cumulative counter or histogram-like set of metrics) that don't require keeping state like this.

pkg/monitors/jenkins/nodemetrics.go
```diff
@@ -0,0 +1,72 @@
...
+	"jenkins.node.vm.memory.heap.usage":    "vm.memory.non-heap.used",
+	"jenkins.node.vm.memory.non-heap.used": "vm.memory.non-heap.used",
+	"jenkins.node.queue.size":              "jenkins.queue.size.value",
+	"jenkins.node.health-check.score":      "jenkins.health-check.score",
+	"jenkins.node.executor.count":          "jenkins.executor.count.value",
+	"jenkins.node.executor.in-use":         "jenkins.executor.in-use.value",
+}
+
+type NodeMetrics struct {
+	Version string                            `json:"version"`
```
keitwb: Node version could make a good dimension property.

pkg/monitors/jenkins/client/client.go
dloucasfx: yes, makes sense, I'll change that

pkg/monitors/jenkins/jobmetrics.go
dloucasfx: SUCCESS, UNSTABLE, FAILURE, NOT_BUILT. ABORTED and empty string (Null) while still running. 
https://javadoc.jenkins-ci.org/hudson/model/Result.html
Aside from changing the dimension name, anything else is needed based on the above result state?

pkg/monitors/jenkins/jobmetrics.go
dloucasfx: roger that, I will change it

pkg/monitors/jenkins/jobmetrics.go
dloucasfx: Assumption: builds can be completed out of order

The logic iterates through the last 100 builds, which is the max `GetBuildsFields` returns.
- for each build:
          * skip if it's in running state 
          * check if the build completion time (ie: start + duration) , is larger than the job stored build time from the previous monitor call. If it is, then this build has not been processed. 1) update the time placeholder if it's larger than the stored value and 2) send the datapoint

pkg/monitors/jenkins/nodemetrics.go
dloucasfx: yeah, chatted with @jaysfx about it, he was planning to implement it.
I can come back and update this code once the method is created, for now, you ok marking this as resolved?

pkg/monitors/jenkins/slavestatus.go
dloucasfx: done

pkg/monitors/jenkins/client/client.go
dloucasfx: done

pkg/monitors/jenkins/client/client.go
dloucasfx: good point, done

pkg/monitors/jenkins/jenkins.go
dloucasfx: we'll discuss in person to why this approach is followed

pkg/monitors/jenkins/nodemetrics.go
dloucasfx: This is not the Jenkins version, I think it might be the metric plugin version.
```
version: "4.0.0"
```
but again, checking the UI, the plugin version is 4.0.2.6

pkg/monitors/jenkins/jobmetrics.go
```diff
@@ -0,0 +1,106 @@
...
+	var dimensions = make(map[string]string)
+	dimensions["job_name"] = jobName
+
+	lastBuild := m.lastBuildSent[jobName]
+	for _, jobBuild := range buildsResp.Builds {
+		if jobBuild.Building {
+			continue
+		}
+		buildCompletion := jobBuild.Timestamp + jobBuild.Duration
+		if buildCompletion > lastBuild {
```
asuresh4: does this imply that this metric could potentially be emitted at irregular intervals?

dloucasfx: @asuresh4 
That's correct. But after further internal discussion, I will be using the cumulative metric instead , 2 main metrics to keep track of (build count and build total time) and each will be broken down to 5 different results `SUCCESS, UNSTABLE, FAILURE, NOT_BUILT. ABORTED`, so a total of 10 MTS for each job max.

docs/monitors/jenkins.md
```diff
@@ -0,0 +1,160 @@
...
+| `useHTTPS` | no | `bool` | If true, the agent will connect to the exporter using HTTPS instead of plain HTTP. (**default:** `false`) |
+| `httpHeaders` | no | `map of strings` | A map of key=message-header and value=header-value. Comma separated multiple values for the same message-header is supported. |
+| `skipVerify` | no | `bool` | If useHTTPS is true and this option is also true, the exporter's TLS cert will not be verified. (**default:** `false`) |
+| `caCertPath` | no | `string` | Path to the CA cert that has signed the TLS cert, unnecessary if `skipVerify` is set to false. |
+| `clientCertPath` | no | `string` | Path to the client TLS cert to use for TLS required connections |
+| `clientKeyPath` | no | `string` | Path to the client TLS key to use for TLS required connections |
+| `host` | **yes** | `string` | Host of the jenkins instance |
+| `port` | **yes** | `string` | Port of the jenkins instance |
+| `metricsKey` | **yes** | `string` | Api key to authenticate to the jenkins metrics plugin |
+| `jenkins_cluster` | **yes** | `string` | Dimension with a value that uniquely identifies the jenkins cluster |
```
keitwb: Explain what this is used for and why it cannot be derived automatically.

docs/monitors/jenkins.md
```diff
@@ -0,0 +1,160 @@
...
+| `dimUpdateInterval` | no | `int64` | The interval (in minutes) at which to send a dimension update (default is 5 minutes). (**default:** `5m`) |
```
keitwb: This is unnecessary, the agent writer will handle deduping so just send the updates every interval that you send metrics.

docs/monitors/jenkins.md
```diff
@@ -0,0 +1,160 @@
...
+copy the whitelist.json, modify it, and reference that in `metricsToExclude`.
+
+## Dimensions
+
+The following dimensions may occur on metrics emitted by this monitor.  Some
+dimensions may be specific to certain metrics.
+
+| Name | Description |
+| ---  | ---         |
+| `Job` | Name of the job. |
```
keitwb: Lowercase dimension names.

pkg/monitors/jenkins/jenkins.go
```diff
@@ -0,0 +1,149 @@
...
+
+	uniqueDimeMap := map[string]string{"jenkins_cluster": conf.JenkinsCluster}
+	// anonymous function used to inject the unique dimension
+	updateAndSendDPS := func(dps ...*datapoint.Datapoint) {
+		for _, dp := range dps {
+			if dp.Dimensions == nil {
+				dp.Dimensions = uniqueDimeMap
```
keitwb: `Dime` --> `Dim`

keitwb: Also you don't want to set it to the uniqueDimeMap because logic in `Output.SendDatapoints` might modify the map and mess up the map used in this montior.  Use `utils.MergeStringMaps` which will simplify this and remove the need for `if/else`.

pkg/monitors/jenkins/jenkins.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	jkClient, err := jc.NewJenkinsClient(conf.Host, conf.MetricsKey, conf.Port, conf.Scheme(), httpClient)
+	if err != nil {
+		return err
+	}
+	m.ctx, m.cancel = context.WithCancel(context.Background())
+	m.lastBuildSent = make(map[string]int64)
+
+	uniqueDimeMap := map[string]string{"jenkins_cluster": conf.JenkinsCluster}
+	// anonymous function used to inject the unique dimension
```
keitwb: Get rid of this comment.  It is pretty obvious what it does.

docs/monitors/jenkins.md
dloucasfx: @keitwb 
While discussing this yesterday, there were two concerns doing this with regular interval:
- frequently calling jenkins unnecessarily to fetch the labels
- the worry of the channel buffer filling up in case of throttling (@jaysfx  has more context), not sure if this is still an issue if we retry frequently

docs/monitors/jenkins.md
dloucasfx: Copied this from collectd metadata , this whole list is inaccurate and needs to be update.
I noticed other monitors don't populate this list, is there a reason for that?

pkg/monitors/jenkins/jenkins.go
dloucasfx: done

pkg/monitors/jenkins/nodemetrics.go
```diff
@@ -0,0 +1,99 @@
...
+func getUniqueDimension(jkClient jc.JenkinsClient, conf *Config) (*types.Dimension, error) {
+	nodeInfo, err := jkClient.GoJenkins.Info()
+	if err != nil {
+		return nil, err
+	}
+
+	dim := &types.Dimension{
+		Name:              "jenkins_cluster",
+		Value:             conf.JenkinsCluster,
+		Tags:              map[string]bool{},
```
keitwb: See if there are any other useful values that could be sent as properties.

keitwb: e.g. jenkins version.

dloucasfx: @keitwb I couldn't find the jenkins version through api, I will keep digging https://github.com/signalfx/signalfx-agent/pull/1191#discussion_r377748412

docs/monitors/jenkins.md
keitwb: There is a JIRA issue created to go back and fill in the other monitors as well.

pkg/monitors/jenkins/nodemetrics.go
dloucasfx: @keitwb found the version, it's in the header ("X-Jenkins"), I will add it as properties

pkg/monitors/jenkins/nodemetrics.go
```diff
@@ -0,0 +1,99 @@
...
+	"jenkins.node.executor.count":          "jenkins.executor.count.value",
+	"jenkins.node.executor.in-use":         "jenkins.executor.in-use.value",
+}
+
+type NodeMetrics struct {
+	Version string                            `json:"version"`
+	Gauges  map[string]map[string]interface{} `json:"gauges"`
+}
+
+func nodeMetrics(jkClient jc.JenkinsClient, conf *Config) ([]*datapoint.Datapoint, error) {
```
asuresh4: seems like this method does not really need the whole config, just the `metricsKey`. I think we should just pass the required field(s) from the config. same thing with other methods like `liveness`, `getUniqueDimension`.

Also, what do you think about adding methods like `nodeMetrics`, `liveness` etc on to a struct and initializing that struct in the `Configure` method, rather than passing the client around?

pkg/monitors/jenkins/nodemetrics.go
dloucasfx:
```
Detecting Jenkins version
To check the version of Jenkins, load the top page (or, as of 1.483, any .../api/* page too) and check for the X-Jenkins response header. This contains the version number of Jenkins, like "1.404" This is also a good way to check if an URL is a Jenkins URL.
```

version will be added a property over `jenkins_cluster`

pkg/monitors/jenkins/jenkins.go
dloucasfx: fixed

pkg/monitors/jenkins/jenkins.go
dloucasfx: done

docs/monitors/jenkins.md
dloucasfx: change is made

docs/monitors/jenkins.md
dloucasfx: change is made

docs/monitors/jenkins.md
dloucasfx: done

pkg/monitors/jenkins/nodemetrics.go
dloucasfx: done

pkg/monitors/jenkins/nodemetrics.go
dloucasfx: As we discussed, made the change to only pass metricsKey 
As for the changing the functions into method with a receiver containing the jenkinsClient, with the last push of the code, it makes no sense to do that as each metric call will have its own jenkins client

pkg/monitors/jenkins/jobbuildstate_test.go
```diff
@@ -0,0 +1,260 @@
...
+	thirdDPS := monitor.getJobDPS(job1, thirdBuilds)
+	assert.EqualValues(t, thirdMetricStateResult, monitor.JobsMetricsState[job1FullName])
+	assert.ElementsMatch(t, expectedDPS3, thirdDPS)
+}
+
+func TestJobsBuildsStateCleanup(t *testing.T) {
+	monitor.JobsMetricsState[job1FullName] = thirdMetricStateResult
+	monitor.JobsMetricsState[job1FullName] = secondMetricStateResult
+	assert.Contains(t, monitor.JobsMetricsState, job1FullName)
+	assert.Contains(t, monitor.JobsMetricsState, job1FullName)
```
jaysfx: extra assert?

pkg/monitors/jenkins/jobmetrics.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+func (m *Monitor) jobMetrics(jkClient jc.JenkinsClient) ([]*datapoint.Datapoint, error) {
+	// GetAllJobs returns the first level jobs
+	firstLevelJobs, err := jkClient.GoJenkins.GetAllJobs()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get firstLevelJobs %v", err)
+	}
+
+	allJobs := getAllJobs(firstLevelJobs)
```
jaysfx: Wonder if we should cache this and only run periodically?

pkg/monitors/jenkins/nodemetrics.go
```diff
@@ -0,0 +1,100 @@
...
+					// Could be a string
+					logger.Warnf("Skipping metric: failed to cast value for key %s and value %v : %v", metric, val, err)
+					continue
+				}
+				dps = append(dps, datapoint.New(metric, nil, v, metricSet[metric].Type, time.Time{}))
+			} else {
+				logger.Warnf("Failed to find json key for metric: %s", metric)
+			}
+		}
+	}
```
jaysfx: Same here, I think it can be unnested.

pkg/monitors/jenkins/jobmetrics.go
keitwb: Dani and I talked about this before and we decided that it should run everytime to be the most responsive, but only if it was not a huge burden on the Jenkins server.

pkg/monitors/jenkins/metadata.yaml
```diff
@@ -0,0 +1,123 @@
...
+      description: Slave display name.
+    jenkins_cluster:
+      description: A user defined value that uniquely identifies the current jenkins instance.
+    jenkins_version:
+      description: The Jenkins version.
+  doc: |
+    This monitor collects metrics from the Jenkins Metrics Plugin by hitting these endpoints:
+    [../api/json](https://wiki.jenkins.io/display/jenkins/remote+access+api)
+    and the plugin [metrics/&lt;MetricsKey&gt;/..](https://wiki.jenkins.io/display/JENKINS/Metrics+Plugin)
+    (default and optional Codahale/Dropwizard JVM metrics).
```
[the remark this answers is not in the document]
dloucasfx: done

keitwb:
@dloucasfx this seems reasonably complete.  What is remaining to make it at least ready for an initial beta release?

keitwb:
@dloucasfx Can you either close this or rebase it with the latest master and fix up for a beta release?

dloucasfx:
@keitwb 
I haven't got a chance to migrate and update the canned dashboards and charts.

emaderer:
@dloucasfx  can we close this PR?

PR #1192, opened 2020-02-11
1095 smart agent 5 updates

galpizarsfx:
Added blurbs to make it clear which version of the agent pertains to which filtering document.


docs/legacy-filtering.md
```diff
@@ -1,10 +1,13 @@
 # Legacy Filtering
-This page describes the old style of filtering and is deprecated and removed in
-agent version 5.0+. See [Filtering](filtering.md) for how to configure
-filtering in SignalFx Smart Agent 4.7.0+.
+
+## Before you begin
+This document only applies to users who are running a version of the Smart Agent previous to 4.7.0.
```
keitwb: Not true, it applies to people using agent < 5.0 who have never switched to the new style filtering.

galpizarsfx: Does this make better sense? The nuance between the updated filtering option versus agent version is a bit confusing. 

---

For **Legacy Filtering** document, I updated with:

This document only applies to users who: 
  * Are running a version of the Smart Agent lower than 5.0 and
  * Have not updated their filtering preferences. 
    * Specifically, this document applies to users who have **not** set `enableBuiltInFiltering: true` in the config file for the `whitelist.json` file. 

If you are running Smart Agent version 5.0 and have updated this filtering preference, see [Filtering](filtering.md).

----

For **Filtering**, I updated with:

This document only applies to users who: 
   * Are running Smart Agent version 5.0, and
   * Have updated their filtering preference.
      * Specifically, this document applies to users who have set `enableBuiltInFiltering: true` in the config file for the `whitelist.json` file. 

If you are running a previous version of the Smart Agent and have not updated this filtering preference, see [Legacy Filtering](legacy-filtering.md).

keitwb:
Closing as this is stale and very out of date.

PR #1194, opened 2020-02-11
Helm improvements for upgrade

jaysfx:
* Add nodeSelector as option
* Add v5 to configmap name so that when upgrading existing pods will use the old
  configmap (even though it's been deleted) and the new configmap will be used
  by the newly deployed pods.

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
+++ b/deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
@@ -55,3 +55,15 @@ Create the name of the secret holding the token
 {{- define "signalfx-agent.secretName" -}}
     {{ default (include "signalfx-agent.fullname" .) .Values.signalFxAccessTokenSecretName }}
 {{- end -}}
+
+{{/*
+Create the configmap name. It will have -v<major version> appended to it from v5 and onward.
+*/}}
+{{- define "configmap-name" -}}
+{{- $major_version := .Values.agentVersion | default "0.0.0" | splitList "." | first | atoi -}}
+{{- if ge $major_version 5 -}}
+{{- template "signalfx-agent.fullname" . -}}-v{{- $major_version -}}
+{{- else -}}
+{{- template "signalfx-agent.fullname" . -}}
+{{- end -}}
+{{- end -}}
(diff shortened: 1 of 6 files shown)


keitwb:
So if somebody does a `helm upgrade` with this new version of the agent chart, it would replace the agents and assign the new configmap to only the new agent instances while it was rolling out?

I wonder if there is some way to do a canary-type test with just one node with Helm?

deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
```diff
@@ -55,3 +55,15 @@ Create the name of the secret holding the token
 {{- define "signalfx-agent.secretName" -}}
     {{ default (include "signalfx-agent.fullname" .) .Values.signalFxAccessTokenSecretName }}
 {{- end -}}
+
+{{/*
+Create the configmap name. It will have -v<major version> appended to it from v5 and onward.
+*/}}
+{{- define "configmap-name" -}}
+{{- $major_version := .Values.agentVersion | splitList "." | first | atoi -}}
```
keitwb: Since you don't have to set `agentVersion` (if using custom image), will this blow up in that case?

deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
```diff
@@ -55,3 +55,15 @@ Create the name of the secret holding the token
...
+{{- if ge $major_version 5 -}}
```
keitwb: Nice, this seems like a good backwards-compatible solution.

deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
jaysfx: Hmm yeah it's going to fail in that case. Should we make `agentVersion` required then since we're depending on its value to control the update? (ie if they're updating a custom image from v4 to v5, it'd work this time if it goes to v0 but next breaking change it'd still be v0). We also use the `agentVersion` as a pod label. So I think just making it required would be fine, I don't think many people use a custom image anyway.

keitwb: Just use the [default helper](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) and make the default `0.0.0` and then we don't have to worry about all of that.

keitwb: If somebody is using a custom image they don't need us holding their hand on upgrades.

PR #1227, opened 2020-02-26
Add PCF tile build to circleci

jchengsfx:


.circleci/config.yml
```diff
@@ -591,6 +613,9 @@ workflows:
       - dependency_check:
           requires:
             - build
+      - pcf_tile_build:
+          requires:
+            - build
```
keitwb: Let's do this only on master and tag pushes since there isn't really anything in the rest of the project that this depends on to work successfully.

jchengsfx: > Let's do this only on master and tag pushes since there isn't really anything in the rest of the project that this depends on to work successfully.

Alternatively, should we run for all branches and just exit fast unless there is something in `deployments/cloudfoundry` that has changes?

keitwb: Sure, if you can do that and make it also always run on master/release tags.

PR #1230, opened 2020-02-27
Make Helm monitor config easier to override

keitwb:
This adds a config option called 'configureStandardMonitoring' that defaults to
'true' that keeps the standard set of host + docker + K8s monitors.  If set to
'false', then the only monitors that will be added are the ones specified in the
'monitors' config option.

Removes all of the 'built-in' monitor configuration from the default value of
'montiors'.


keitwb:
Helm check is going to fail due to the fact that we haven't actually released 5.0.0 yet, but it will just be temporary.

PR #1233, opened 2020-03-02
Update PCF tile agent config for 5.0.0

jchengsfx:


deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
```diff
@@ -62,16 +62,14 @@ monitors:
 <% else %>
  - type: host-metadata
  - type: processlist
- - type: collectd/cpu
  - type: collectd/cpufreq
```
keitwb: Remove cpufreq.

PR #1241, opened 2020-03-09
Validate default false

keitwb:


keitwb:
Note that this will still cause an error if the discovery rule is blatantly malformed.

PR #1247, opened 2020-03-10
Resolve direct go mod download issues

jchengsfx:
- Update github.com/openshift/api with commit ID since all tags were
removed
- Replace launchpad.net/gocheck with github.com/go-check/check to avoid
bzr dependency

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,8 @@ replace code.cloudfoundry.org/go-loggregator => github.com/signalfx/go-loggregat
 
 replace github.com/influxdata/telegraf => github.com/signalfx/telegraf v0.10.2-0.20200220234948-7794d95dad35
 
+replace launchpad.net/gocheck => github.com/go-check/check v0.0.0-20200227125254-8fa46927fb4f
+
 require (
 	code.cloudfoundry.org/go-diodes v0.0.0-20190809170250-f77fb823c7ee // indirect
 	code.cloudfoundry.org/go-loggregator v7.4.0+incompatible
@@ -117,7 +119,7 @@ require (
 	github.com/olekukonko/tablewriter v0.0.1
 	github.com/onsi/ginkgo v1.10.1
 	github.com/onsi/gomega v1.7.0
-	github.com/openshift/api v3.9.0+incompatible
+	github.com/openshift/api v0.0.0-20191213091414-3fbf6bcf78e8
 	github.com/openshift/client-go v0.0.0-20191216194936-57f413491e9e
 	github.com/opentracing/opentracing-go v1.1.0
 	github.com/ory-am/common v0.4.0 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -8,6 +8,8 @@ replace code.cloudfoundry.org/go-loggregator => github.com/signalfx/go-loggregat
 
 replace github.com/influxdata/telegraf => github.com/signalfx/telegraf v0.10.2-0.20200220234948-7794d95dad35
 
+replace launchpad.net/gocheck => github.com/go-check/check v0.0.0-20200227125254-8fa46927fb4f
```
keitwb: Let's avoid these if possible -- they have caused a lot of headaches in golib when using that module in other places and now we kind of support importing the agent as a library (see #1116).  If we are replacing the module below with the same version, just pinned to a commit hash instead of a tag, why does the bzr dependency differ?

jchengsfx: > Let's avoid these if possible -- they have caused a lot of headaches in golib when using that module in other places and now we kind of support importing the agent as a library (see #1116). If we are replacing the module below with the same version, just pinned to a commit hash instead of a tag, why does the bzr dependency differ?

`launchpad.net/gocheck` is a transitive dependency of `code.cloudfoundry.org/go-loggregator` and downloading the module directly will fail if `bzr` is not installed since go will try to download directly from https://launchpad.net.  The project has been migrated to https://github.com/go-check/check.

keitwb: If you are using the standard GOPROXY server, does it still require using bzr or will it work without it installed?  I don't have bzr installed on my dev VM and I've never had any issues.  Are you sure this isn't only a problem when using GOPROXY=off?

jchengsfx: The default GOPROXY server has the modules cached, so `bzr` is not required; go will just download everything over https from the proxy server instead of trying to checkout the project from the origin.

Running with `GOPROXY=direct` and without the replacement fails:
>docker run --rm -e GO111MODULE=on -e GOPROXY=direct -v $(pwd):/code -w /code golang:1.13.1 go mod download
go: code.cloudfoundry.org/go-loggregator@v7.4.0+incompatible requires
	launchpad.net/gocheck@v0.0.0-20140225173054-000000000087: bzr branch --use-existing-dir -- https://launchpad.net/~niemeyer/gocheck/trunk . in /go/pkg/mod/cache/vcs/f46ce2ae80d31f9b0a29099baa203e3b6d269dace4e5357a2cf74bd109e13339: exec: "bzr": executable file not found in $PATH

Running with default `GOPROXY` and without the replacement is successful:
>docker run --rm -e GO111MODULE=on -e GOPROXY= -v $(pwd):/code -w /code golang:1.13.1 go mod download

Also, see https://github.com/golang/go/issues/29790#issuecomment-457304976 and https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy/.

keitwb:
If this isn't an issue anymore then lets avoid doing replace statements in the go.mod file and forget about this.

PR #1250, opened 2020-03-12
add infra correlation for spans

codesmith14:


pkg/core/writer/tracetracker/tracker.go
```diff
@@ -98,68 +99,136 @@ func (a *ActiveServiceTracker) processSpan(span *trace.Span, now time.Time) {
...
 		return
 	}
-
 	service := *span.LocalEndpoint.ServiceName
-	a.ensureServiceActive(service, now)
+
+	// TODO: check on what the tag name should be in order to fetch environment
+	environment, environmentFound := span.Tags["environment"]
+	if !environmentFound {
+		// TODO set to default environment
```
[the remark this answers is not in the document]
codesmith14: Yes, we are going to set a default environment for both pod/container and host-level.  For host dims we want to send that default always because there are multiple environments.  But for k8s/container dims, we have a plan to send the default and then overwrite it if a real value comes in on a span tag.

pkg/core/writer/tracetracker/tracker.go
```diff
@@ -98,68 +99,136 @@ func (a *ActiveServiceTracker) processSpan(span *trace.Span, now time.Time) {
...
+	}
+
+	// Handle host level service and environment correlation
+	if dimValue, ok := span.Tags["host"]; ok {
```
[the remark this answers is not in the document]
codesmith14: @keitwb Why does it matter?  Host is guaranteed to be on the span, and we want to associate the service/environment with all of those applicable ids AWSUniqueId/gcp_id/azure_resource_id/etc.  We do that inside inside this if statement with the for loop over hostIDDims.  

As for the original correlation metrics that behavior should be untouched here.

pkg/core/writer/correlations/client.go
```diff
@@ -0,0 +1,172 @@
...
+	logUpdates bool
+
+	TotalClientError4xxResponses int64
+	TotalRetriedUpdates          int64
+	TotalInvalidDimensions       int64
+}
+
+// NewCorrelationClient returns a new Client
+func NewCorrelationClient(ctx context.Context, conf *config.WriterConfig) (CorrelationClient, error) {
+	// TODO: someday... adapt property filters to work with correlation objects
```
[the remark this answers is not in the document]
codesmith14: cool, will remove

pkg/core/writer/tracetracker/tracker.go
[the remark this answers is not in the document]
codesmith14: If service isn't "", we could just always add the service to ensureServiceActive, and then iterate overall hostIDDims regardless.  That removes the if "host" in span.Tags, but it doesn't change the correlation behavior.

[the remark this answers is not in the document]
keitwb: What does it mean for a span to have a `host` tag?  And conversely, what kind of span would _not_ have a `host` tag?  Is that supposed to be set by the application or is this dependent on the agent having already set it before this method is called?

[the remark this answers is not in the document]
codesmith14: actually yeah it makes more sense to do that, but it doesn't address your comment.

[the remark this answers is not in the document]
codesmith14: @keitwb, All hostIDDims are added to the span prior to entering the ActiveServiceTracker.  We were leveraging this.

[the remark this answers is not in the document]
keitwb: ok, yeah this if condition seems unnecessary and confusing.

codesmith14:
I think load wise, this shouldn't be significantly different from what active service tracker was already doing.  The active service tracker was already reading every span as it came through.  My main concern is around requests backing up on startup as the service/environment caches fill out.  We only send requests when we see new services/environments and for new k8s/container ids.  The get we're going to add at start up should also help alleviate some of this.

docs/config-schema.md
```diff
@@ -48,7 +48,8 @@ if not set.
...
 | `globalDimensions` | no | map of strings | Dimensions (key:value pairs) that will be added to every datapoint emitted by the agent. To specify that all metrics should be high-resolution, add the dimension `sf_hires: 1` |
-| `cluster` | no | string | The logical environment/cluster that this agent instance is running in. All of the services that this instance monitors should be in the same environment as well. This value, if provided, will be synced as a property onto the `host` dimension, or onto any cloud-provided specific dimensions (`AWSUniqueId`, `gcp_id`, and `azure_resource_id`) when available. Example values: "prod-usa", "dev" |
+| `environment` | no | string | The logical environment/cluster that this agent instance is running in. All of the services that this instance monitors should be in the same environment as well. This value, if provided, will be synced as a property onto the `host` dimension, or onto any cloud-provided specific dimensions (`AWSUniqueId`, `gcp_id`, and `azure_resource_id`) when available. Example values: "prod-usa", "dev" |
```
keitwb: @mpetazzoni is proposing having this as a span tag config option on the forwarder itself.  That seems like a better idea than adding another top-level config option that duplicates `cluster`.

docs/config-schema.md
```diff
@@ -149,7 +150,8 @@ The **nested** `writer` config object has the following fields:
 | `logDimensionUpdates` | no | bool | If `true`, dimension updates will be logged at the INFO level. (**default:** `false`) |
 | `logDroppedDatapoints` | no | bool | If true, and the log level is `debug`, filtered out datapoints will be logged. (**default:** `false`) |
 | `addGlobalDimensionsAsSpanTags` | no | bool | If true, the dimensions specified in the top-level `globalDimensions` configuration will be added to the tag set of all spans that are emitted by the writer.  If this is false, only the "host id" dimensions such as `host`, `AwsUniqueId`, etc. are added to the span tags. (**default:** `false`) |
-| `sendTraceHostCorrelationMetrics` | no | bool | Whether to send host correlation metrics to correlation traced services with the underlying host (**default:** `true`) |
+| `traceInfrastructureCorrelation` | no | bool | Whether to send host correlation properties to correlate traced services with the infrastructure (**default:** `"false"`) |
```
keitwb: Just make it always send them if somebody is sending traces through the agent and take out this config option.  There are too many config options and this is way too confusing what should be on or off.  Just make it always on.

pkg/core/config/config.go
```diff
@@ -161,6 +165,19 @@ type Config struct {
...
 
+// remapValues is used to move deprecated config values to newer keys.
+// a warning should be logged for each migrated value
+func (c *Config) remapValues() {
+	// if environment is empty, then set it to conf.cluster for legacy behavior
+	if c.Environment == "" && c.Cluster != "" {
+		log.WithFields(log.Fields{
+			"cluster":     c.Cluster,
+			"environment": c.Environment,
+		}).Warn("Config value 'cluster' is deprecated, please use `environment` instead")
```
keitwb: Yeah let's definitely take this out and use the extra span tag config on the forwarder instead.

pkg/core/writer/correlations/client.go
```diff
@@ -0,0 +1,227 @@
...
+// CorrelationClient is an interface for correlations.Client
+type CorrelationClient interface {
+	Correlate(correlation *Correlation)
+	Delete(correlation *Correlation)
+	Get(dimName string, dimValue string, callback func(map[string][]string, error))
+	Start()
+}
+
+// NOOPClient implements CorrelationClient interface but doesn't do anything with the correlation
+type NOOPClient struct{}
```
keitwb: Put this in a separate file to declutter it from the core logic.

pkg/core/writer/correlations/client.go
```diff
@@ -0,0 +1,227 @@
...
+			}
+
+			log.WithError(err).WithFields(log.Fields{
+				"url":         req.URL.String(),
+				"correlation": r,
+			}).Error("Unable to update dimension, retrying")
+			atomic.AddInt64(&cc.TotalRetriedUpdates, int64(1))
+			// The retry is meant to provide some measure of robustness against
+			// temporary API failures.  If the API is down for significant
+			// periods of time, dimension updates will probably eventually back
```
keitwb: Comment is out of date.

pkg/core/writer/correlations/client.go
```diff
@@ -0,0 +1,227 @@
...
+	return nil
+}
+
+func (cc *Client) processChan() {
+	for {
+		select {
+		case <-cc.ctx.Done():
+			return
+		case req := <-cc.requestChan:
+			_ = cc.makeRequest(req)
```
keitwb: Shouldn't any errors be at least logged?

pkg/core/writer/dimensions/client.go
```diff
@@ -175,7 +176,7 @@ func (dc *DimensionClient) processQueue() {
 			dc.Unlock()
 
 			if err := dc.setPropertiesOnDimension(delayedDim.Dimension); err != nil {
-				log.WithError(err).WithField("dim", delayedDim.Key()).Error("Could not send dimension update")
+				log.WithError(err).WithField("dim", delayedDim.Key()).Error("Could not Send dimension update")
```
keitwb: What's up with the random capitalization?

keitwb: Oh, looks like you did a s/send/Send/ and it did the string literals as well.

codesmith14: Honestly, I have no idea what happened there?  Will fix it though.

tests/helpers/assertions.py
```diff
@@ -395,6 +395,33 @@ def has_dim_prop(fake_services, dim_name, dim_value, prop_name, prop_value=None)
...
+    tests for the presence of the property regardless of value. Only tests if the given properties are present, does
+    not test that they are the only values in the given property set
+    """
+    dim = fake_services.dims[dim_name].get(dim_value, {})
+    props = dim.get(prop_name, [])
+    has_props = False
+    if props is not None:
+        has_props = True
+        for v in prop_values:
+            has_props = True and v in props
```
keitwb: Is the `True and` part really necessary?

tests/helpers/assertions.py
```diff
@@ -395,6 +395,33 @@ def has_dim_prop(fake_services, dim_name, dim_value, prop_name, prop_value=None)
...
+    return has_props
+
+def not_has_dim_set_prop(fake_services, dim_name, dim_value, prop_name, prop_values):
```
keitwb: Name should be `has_no_dim_prop`.

keitwb: Also it should be able to handle a scalar value for `prop_values`, as well as a list.

keitwb: And then call it `prop_value`.

pkg/core/writer/correlations/client.go
codesmith14: Removed because we don't need now that we're removing the traceHostCorrelation config option

tests/helpers/assertions.py
sstewart-signalfx: should be `has_props and`, updated

tests/helpers/assertions.py
sstewart-signalfx: Updated to use `*prop_value` as the last parameter, but that also means that we can't use named arguments when this gets used in `wait_for`

keitwb: I would make it `prop_value: Union[str, List[str]]` (you don't have to include Py3 typing annotations, but they are nice to have) and then have `if isinstance(prop_value, str): prop_value = [prop_value]` or something like that to make it work nicely with partials.

pkg/core/writer/tracetracker/tracker.go
```diff
@@ -20,44 +20,88 @@ const spanCorrelationMetricName = "sf.int.service.heartbeat"
...
+func (a *ActiveServiceTracker) LoadCorrelations() {
+	// asynchronously fetch all services and environments for each hostIDDim at startup
+	for dimName, dimValue := range a.hostIDDims {
+		dimName := dimName
+		dimValue := dimValue
+		a.correlationClient.Get(dimName, dimValue, func(correlations map[string][]string, err error) {
+			// if there was an error unmarshalling the response,
+			// then there's not much we can do.  Just move on.
+			if err != nil {
+				return
```
keitwb: I would make sure this is logged either here or higher up.

keitwb: Also if this is just an error about unmarshalling, what happens if the server can't be reached at all?

pkg/core/writer/tracetracker/tracker.go
```diff
@@ -20,44 +20,88 @@ const spanCorrelationMetricName = "sf.int.service.heartbeat"
...
-type serviceStatus struct {
-	LastSeen    time.Time
-	ServiceName string
+// LoadCorrelations asynchronously retrieves all known correlations from the backend
+// for all known hostIDDims.  This allows the agent to timeout and manage correlation
+// deletions on restart.
+func (a *ActiveServiceTracker) LoadCorrelations() {
+	// asynchronously fetch all services and environments for each hostIDDim at startup
+	for dimName, dimValue := range a.hostIDDims {
+		dimName := dimName
```
keitwb: This is really unnecessary since you aren't passing by pointer.

pkg/core/writer/tracetracker/tracker.go
```diff
@@ -20,44 +20,88 @@ const spanCorrelationMetricName = "sf.int.service.heartbeat"
...
+		dimValue := dimValue
+		a.correlationClient.Get(dimName, dimValue, func(correlations map[string][]string, err error) {
+			// if there was an error unmarshalling the response,
+			// then there's not much we can do.  Just move on.
+			if err != nil {
+				return
+			}
+			a.Lock()
```
keitwb: This lock is hard to understand, I'm guessing it protects some map mutated by `a.ensureServiceActive`?  If so, that function should just grab the lock right before it's needed instead of it being an implied dependency.  If you want to get multiple services in under the same lock cycle, then just make a method `ensureServicesActive` that takes multiple services.

pkg/core/writer/tracetracker/tracker.go
codesmith14: If the server can't be reached the error is handled by the client.  This call back will just get an empty payload and the error.

pkg/core/writer/tracetracker/tracker.go
codesmith14: i made the lock more granular for the GET call back, but i can't move it into ensureServiceActive with out more significant refactoring of all of the locking in the service tracker.  In general the original behavior of the tracker was to lock at the beginning of any externally accessed method.  This was what I continued doing.

pkg/core/writer/tracetracker/tracker.go
codesmith14: I've added logging for failures to unmarshal the payload

pkg/core/writer/tracetracker/tracker.go
codesmith14: I went ahead and did the more extensive refactoring of locking.  The caches themselves have mutex's.  And I've made the mutex in the service tracker specific to the dpCache.  Once we remove the dpCache we can remove the the mutex on the service tracker entirely.

PR #1254, opened 2020-03-18
add authentication with SASL to jmx monitor

bjsignalfx:
Implemented support for SASL/PLAIN, SASL/DIGEST-MD5 and SASL/CRAM-MD5 profiles which can be combined with profile TLS to encrypt connections.

Issue: https://signalfuse.atlassian.net/browse/INT-1813


mstumpfx:
let's get some integration tests for this if possible

PR #1279, opened 2020-04-06
k8s-apiserver: make apiserver_request_duration_seconds_bucket non-default

mstumpfx:
The metric `apiserver_request_duration_seconds_bucket` is a high cardinality metric, as well as reports by default 40 different buckets, making it generate at least 40 MTSs per resource.

We do not use this metric in the SignalFx default content, so we should not enable it by default for users.


jrcamp:
> We do not use this metric in the SignalFx default content

Did you check if there are any others marked as default true that are not used in builtin content?

jrcamp:
Also is categorization changing for this or was it not bundled to start with?

mstumpfx:
> Also is categorization changing for this or was it not bundled to start with?

No, it is staying the same, it is bundled.

mstumpfx:
> Did you check if there are any others marked as default true that are not used in builtin content?
Yes, from this monitor the rest of the metrics enabled are in our content

PR #1282, opened 2020-04-08
Update mongodb and openstack collectd releases

dloucasfx:
cc: @jaysfx 


mstumpfx:
@dloucasfx looks like you will need to update mongodb tests as well

dloucasfx:
> @dloucasfx looks like you will need to update mongodb tests as well

@mstumpfx yeah working on it, I am not sure how the tests were even passing with this bug :(

PR #1286, opened 2020-04-10
1293 agent python requirement

galpizarsfx:
You must run Visual C++ Compiler for Python 2.7.  ==> (Optional) If you want to invoke a Python script for non-default monitors, specifically exec, then you must have Python 3.8.


molner:
Unless others feel strongly here, I wouldn't specifically list a Python version for the "optional" piece. Just instead say that Python needs to be installed in that particular case.

galpizarsfx:
I went ahead and removed the specific python version.

molner:
I don't like the phrase "Before proceeding, you must install .NET Frame 3.5..."

Most people will already have this installed. This should be changed to something along the lines of "Before proceeding, you should ensure that .NET Framework 3.5 or later is installed".

Issue #1299, opened 2020-04-24
Add native monitor for NTP


xp-1000:
Hello,

Be able to monitor time synchronization on servers where signalfx agent is installed seems important enough to propose a native monitor for ntp.

I checked https://collectd.org/wiki/index.php/Plugin:NTPd but failed to configure on some ntp servers because of https://github.com/collectd/collectd/issues/932

keitwb:
Fixed by #1300

PR #1300, opened 2020-04-24
#1299 add ntp monitor

xp-1000:
hello here is a PR to address https://github.com/signalfx/signalfx-agent/issues/1299
this is the first monitor I implement so please guide me to improve and validate it.
notice it will overrides the default time interval to 30m while this does not need a short interval and should not load default ntp pool.
if signalfx has a public ntp server we should probably use it.
the doc for monitor seems to be auto generated so I did not create id.
I tested building a dev agent and metrics is right sent to signalfx


xp-1000:
I don't know how to fix "check_links" step while the doc is autogenerated on master github branch where ntp does not exist yet obviously

pkg/monitors/ntp/ntp.go
```diff
@@ -0,0 +1,76 @@
...
+
+func init() {
+	monitors.Register(&monitorMetadata, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `singleInstance:"false" acceptsEndpoints:"false"`
+	// The host/ip address of the NTP server.
+	Host string `yaml:"host" default:"pool.ntp.org"`
```
keitwb, line 27: There should not be any default for this, it should be explicit every time this monitor is configured.

pkg/monitors/ntp/ntp.go
```diff
@@ -0,0 +1,76 @@
...
+	utils.RunOnInterval(ctx, func() {
+		options := ntp.QueryOptions{Version: conf.Version, Port: conf.Port, Timeout: time.Duration(conf.Timeout) * time.Second}
+		response, err := ntp.QueryWithOptions(conf.Host, options)
+		if err != nil {
+			m.logger.WithError(err).Error("unable to get ntp statistics")
+			return
+		}
+		clockOffset := response.ClockOffset.Seconds()
+		m.Output.SendDatapoints([]*datapoint.Datapoint{
+			datapoint.New("ntp.offset", map[string]string{"ntp": conf.Host}, datapoint.NewFloatValue(clockOffset), datapoint.Gauge, time.Time{}),
```
keitwb, line 63: I would put the unit in the metric name, e.g. `ntp.offset_seconds`.  A bit more verbose, but makes it easy to understand without having to consult code/docs when making/editing/viewing charts/alerts.

keitwb:
> I don't know how to fix "check_links" step while the doc is autogenerated on master github branch where ntp does not exist yet obviously

Don't worry about this one for new monitors (it always fails), it isn't a required check, just informational.

pkg/monitors/ntp/ntp.go
```diff
@@ -0,0 +1,76 @@
...
+type Config struct {
+	config.MonitorConfig `singleInstance:"false" acceptsEndpoints:"false"`
+	// The host/ip address of the NTP server.
+	Host string `yaml:"host" default:"pool.ntp.org"`
+	// The port of the NTP server.
+	Port int `yaml:"port" default:"123"`
+	// NTP protocol version to.
+	Version int `yaml:"version" default:"4"`
+	// Timeout in seconds for the request.
+	Timeout uint16 `yaml:"timeout" default:"5"`
```
jrcamp, line 33:
```suggestion
	Timeout timeutil.Duration `yaml:"timeout" default:"5s"`
```

pkg/monitors/ntp/ntp.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/signalfx/signalfx-agent/pkg/core/config"
+	"github.com/signalfx/signalfx-agent/pkg/monitors"
+	"github.com/signalfx/signalfx-agent/pkg/monitors/types"
+	"github.com/signalfx/signalfx-agent/pkg/utils"
+	"github.com/sirupsen/logrus"
+
+	"github.com/beevik/ntp"
+)
+
+const minIntervalSeconds = 1800
```
jrcamp, line 17:
```suggestion
const minIntervalSeconds = 30 * time.Minutes
```

pkg/monitors/ntp/ntp.go
jrcamp: Also use constants in genmetadata.go

pkg/monitors/ntp/ntp.go
keitwb: It would no longer be `Seconds` at that point, but the base time.Time unit (nanoseconds), so the var should just be called `minInterval`.

pkg/monitors/ntp/ntp.go
xp-1000: ok I updated the code and seems good for me but I don't resolve this as I don't understand why the "required" value remains "no" in the doc https://github.com/xp-1000/signalfx-agent/blob/1299_ntp_monitor/docs/monitors/ntp.md may be do you have an idea?

pkg/monitors/ntp/ntp.go
xp-1000: both done

pkg/monitors/ntp/ntp.go
keitwb: To make it required you just need to put a struct tag `validate:"required"`.

xp-1000: oh my bad I already knew it .. sorry I just learned go in addition to signalfx integration process so I am not comfortable yet

pkg/monitors/ntp/metadata.yaml
```diff
@@ -0,0 +1,17 @@
+monitors:
+- doc: |
+    This monitor will retrieve clock offset from ntp server.
+
+    It enforces a minimum interval of 15m.
```
keitwb, line 5:
```suggestion
    It enforces a minimum interval of 30m.
```

keitwb: Ok, this is my last request and then I'm good with it.

xp-1000: oh good catch! 
It is true I choose 15m first without any reason then I checked any best practice and see the pool ntp org one so I changed.

keitwb:
Don't forget to run `make docs`.

keitwb:
Thanks for contributing this @xp-1000.

Issue #1301, opened 2020-04-28
Add native monitor for webcheck


xp-1000:
Hello,

a simple webcheck to retrieve http and ssl metrics will be a good feature even if signalfx does not aim to provide synthetics features this could help to check internal or simple web services out the box.

- https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-health-checker.html
- https://collectd.org/wiki/index.php/Plugin:cURL
Both are able to achieve webcheck but it miss some feature and metrics in my opinion.

keitwb:
Fixed by #1302

PR #1302, opened 2020-04-28
 #1301 add http monitor

xp-1000:
a very simple solution for https://github.com/signalfx/signalfx-agent/issues/1301


pkg/monitors/http/http.go
```diff
@@ -0,0 +1,179 @@
...
+func init() {
+	monitors.Register(&monitorMetadata, func() interface{} { return &Monitor{} }, &Config{})
+}
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig `singleInstance:"false" acceptsEndpoints:"false"`
+	Body                 string            `yaml:"body"`
+	FollowRedirects      bool              `yaml:"followRedirects" default:"true"`
+	Headers              map[string]string `yaml:"headers"`
```
jrcamp, line 33: Headers and Timeout can be removed after including https://github.com/signalfx/signalfx-agent/blob/master/pkg/core/common/httpclient/http.go. It provides common HTTP client and TLS configuration. Can check other code for how it's used.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,179 @@
...
+	Method               string            `yaml:"method" default:"GET"`
+	Timeout              int               `yaml:"timeout" default:"5"`
+	URLs                 []string          `yaml:"urls"`
+	Regex                string            `yaml:"regex"`
```
jrcamp, line 37: Add comments for all of the config options (they get included into the user docs).

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,179 @@
...
+		tlsCfg := &tls.Config{InsecureSkipVerify: true}
+		timeoutDuration := time.Duration(conf.Timeout) * time.Second
+		// add data
+		var body io.Reader
+		if conf.Body != "" {
+			body = strings.NewReader(conf.Body)
+		}
+		// get stats for each website
+		for _, site := range conf.URLs {
+			m.logger = m.logger.WithFields(logrus.Fields{"url": site})
```
jrcamp, line 63: Feels a bit odd to be mutating m.logger each time. Could see wanting to run multiple URLs in parallel so probably best to avoid mutating shared state. Would make it a local variable.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,179 @@
...
+			// starts timer
+			now := time.Now()
+			resp, err := client.Do(req)
+			if err != nil {
+				m.logger.WithError(err).Error("could not do the request")
+				continue
+			}
+			dps = append(dps, datapoint.New("http.response_time", dimensions, datapoint.NewFloatValue(time.Since(now).Seconds()), datapoint.Gauge, time.Time{}))
+			dps = append(dps, datapoint.New("http.status_code", dimensions, datapoint.NewIntValue(int64(resp.StatusCode)), datapoint.Gauge, time.Time{}))
+			defer resp.Body.Close()
```
jrcamp, line 102: Would either move Close up to after Do() or maybe close it after reading the response body since it won't close until the function returns. Though this would be less of an issue with a broken up function. See below.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,179 @@
...
+				}
+			}
+			// starts timer
+			now := time.Now()
+			resp, err := client.Do(req)
+			if err != nil {
+				m.logger.WithError(err).Error("could not do the request")
+				continue
+			}
+			dps = append(dps, datapoint.New("http.response_time", dimensions, datapoint.NewFloatValue(time.Since(now).Seconds()), datapoint.Gauge, time.Time{}))
```
jrcamp, line 100: Use constants from genmetadata.go. Applies throughout.

jrcamp, line 100: Can pass multiple items to append so don't need to call more than once.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,179 @@
...
+			defer resp.Body.Close()
+			bodyBytes, err := ioutil.ReadAll(resp.Body)
+			dps = append(dps, datapoint.New("http.content_length", dimensions, datapoint.NewIntValue(int64(len(bodyBytes))), datapoint.Gauge, time.Time{}))
+			if err != nil {
+				m.logger.WithError(err).Error("could not parse body")
+			} else if conf.Regex != "" {
+				var match int64 = 0
+				regex, err := regexp.Compile(conf.Regex)
+				if err != nil {
+					m.logger.WithError(err).Error("failed to compile regular expression")
```
jrcamp, line 111: Consider maybe precompiling on start to cache and check they're all valid.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,179 @@
...
+				if regex.Match(bodyBytes) {
+					match = 1
+				}
+				dps = append(dps, datapoint.New("http.regex_match", dimensions, datapoint.NewIntValue(match), datapoint.Gauge, time.Time{}))
+			}
+			parsedURL, err := url.Parse(resp.Request.URL.String())
+			if err != nil {
+				m.logger.WithError(err).Error("could not parse url")
+				m.Output.SendDatapoints(dps...)
+				continue
```
jrcamp, line 122: If all the time datapoints that were gathered should still be sent (appears to be all the time?) then should refactor sending to a central place instead of relying on multiple call sites to.

xp-1000:
> Looks like a useful monitor, thanks for the contribution!
> 
> I think it'd help to refactor into separate checks for at least http and tls. Each of the checks would take a single url and return the datapoints it collected. You can then have centralized logic that does any common behavior (e.g. apply dimensions). It would also make it easy to parallelize in the future if desired.

Hello,
Thanks for your review.
I should confess I don't be aware of everything is possible and don't fully understand the way agent works yet.
what I want to keep is the ability to retrieve every of these metrics from only one config:
indeed, this could make it easy to configure webchecks like:

```
  - type: http
    intervalSeconds: 60
    urls:
      - "http://google.com"
```

will be enough to check website uptime AND ssl validity without to configure duplicate monitors.
So if you think it is possible to separate checks keeping this ability to configure only one monitor this will be perfect.

may be do you have an example to point ?

thanks

jrcamp:
> So if you think it is possible to separate checks keeping this ability to configure only one monitor this will be perfect.

Yeah I think just one monitor is fine, I'm just talking about breaking the code up into a couple other functions. Maybe something like:

```go
func (m *Monitor) getHTTPStats(url string) ([]*datapoint.Datapoint, error)
func (m *Monitor) getTLSStats(url string) ([]*datapoint.Datapoint, error)
```

pkg/monitors/http/http.go
xp-1000: I am not sure how to reuse for tls but this should be good now

pkg/monitors/http/http.go
xp-1000: fixed

pkg/monitors/http/http.go
xp-1000: ok

pkg/monitors/http/http.go
xp-1000: done

pkg/monitors/http/http.go
xp-1000: I updated the doc. at the origin, there are dedicated metrics like tls valid or regex match and should be reported only if applicable but now it is in dimensions so this is only true for tls expiration metric which will be reported only if https could connect

pkg/monitors/http/http.go
xp-1000: done

xp-1000:
> > So if you think it is possible to separate checks keeping this ability to configure only one monitor this will be perfect.
> 
> Yeah I think just one monitor is fine, I'm just talking about breaking the code up into a couple other functions. Maybe something like:
> 
> ```go
> func (m *Monitor) getHTTPStats(url string) ([]*datapoint.Datapoint, error)
> func (m *Monitor) getTLSStats(url string) ([]*datapoint.Datapoint, error)
> ```

Hello @jrcamp 
I tried to apply your suggestions.
hope it is better

pkg/monitors/http/metadata.yaml
```diff
@@ -0,0 +1,53 @@
+monitors:
+- doc: |
+    This monitor will retrieve metrics (see below) from a list of configured URLs.
+
+    TLS information will automatically be fetched if applicable (from base url or redirection).
```
keitwb, line 5:
```suggestion
    TLS information will automatically be fetched if applicable (from base URL or redirection).
```

pkg/monitors/http/metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+    The configuration will be applied to every requests URLs from configured list
+    so you need to instance multiple times this monitor for different behavior.
+  
+    It could also affect the dimensions available on reported metrics.
+    Some dimensions are optionanly added to its corresponding metric depending of the configuration.
```
keitwb, line 11:
```suggestion
    Some dimensions are optionally added to its corresponding metric depending of the configuration.
```

pkg/monitors/http/metadata.yaml
```diff
@@ -0,0 +1,53 @@
+monitors:
+- doc: |
+    This monitor will retrieve metrics (see below) from a list of configured URLs.
```
keitwb, line 3:
```suggestion
    This monitor will generate metrics based on whether the HTTP responses from the configured URLs match expectations (e.g. correct body, status code, etc).
```

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,254 @@
...
+		}), datapoint.NewIntValue(int64(statusCode)), datapoint.Gauge, time.Time{}),
+	)
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		logger.WithError(err).Error("could not parse body response")
+	} else {
+		contentDimensions := dimensions
+		if m.conf.Regex != "" {
+			matchRegex := false
+			if m.regex.Match(bodyBytes) {
```
keitwb, line 233: This can simplify to just `matchRegex := m.regex.Match(bodyBytes)`.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,254 @@
...
+
+	monitors.Register(&monitorMetadata, func() interface{} { return &Monitor{monitorName: monitorMetadata.MonitorType} }, &Config{})
+}
+
+// Config for this monitor
+type Config struct {
+	config.MonitorConfig  `yaml:",inline" singleInstance:"false" acceptsEndpoints:"true"`
+	httpclient.HTTPConfig `yaml:",inline"`
+	// Optional HTTP request body as string like '{"foo":"bar"}'
+	Body string `yaml:"body"`
```
keitwb, line 37: Call this `RequestBody` as it is very easy to assume this means the expected response body to test against.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,254 @@
...
+		_, err := cert.Verify(opts)
+		if err != nil {
+			logger.WithError(err).Info("failed verify certificate")
+			valid = false
+		}
+	}
+	dimensions := map[string]string{
+		"url":        site.String(),
+		"serverName": host,
+		"isValid":    strconv.FormatBool(valid),
```
keitwb, line 174: `isValid` is a metric value, not a dimension.  A dimension should only be uniquely identifying attributes of a timeseries, not an accidental property of it.  You should just have another metric called something like `http.cert_valid`.

pkg/monitors/http/http.go
```diff
@@ -0,0 +1,254 @@
...
+		logger.WithError(err).Error("could not parse body response")
+	} else {
+		contentDimensions := dimensions
+		if m.conf.Regex != "" {
+			matchRegex := false
+			if m.regex.Match(bodyBytes) {
+				matchRegex = true
+			}
+			contentDimensions = addDimensions(contentDimensions, map[string]string{
+				"matchRegex": strconv.FormatBool(matchRegex),
```
keitwb, line 237: Same issue as for `isValid` with TLS above.  There should be a separate metric called something like`http.content_matched`.

pkg/monitors/http/http.go
xp-1000: ok it was my first implementation and it seems to be better like this in my mind.
I tried to explain why in the documentation, using dimensions will ease a generic detector implementation.

I am not sure there is a demarcated border between what info we can push as dimension and what can push as metrics. 

take the aws metric as example we can create a detector which will raise alert on terminated ec2 with something like a count on any ec2 metric and `filter=filter('aws_state', 'terminated')`.
in this case this is the same, the valuable info we need and use to alert is the value of a dimension, not the metric itself.

I understand this is not the common way to do but it does not seem to be an incompatible or problematic way to go.
That being said, I don't have any strong constraint against coming back to my first implementation with the 2 additional metrics if you think I should but:
- I just would like to explain what
- I am curious if I forget anything about this way which could cause problem

keitwb: That `aws_state` value is a _property_ on the `AWSUniqueId` dimension, and not a _dimension_.  See [Metric Metadata](https://docs.signalfx.com/en/latest/metrics-metadata/metrics-metadata.html), especially about a page into that doc and down where it has the table.  The key distinction is that the `aws_state` can change without creating a distinct timeseries for each unique value of the `aws_state`.  Whereas, with the way you have `isValid`, it will make new, non-overlapping timeseries when its value changes.

The agent does have the ability to set properties on dimensions, but I think the more appropriate thing to do here would be to just have separate timeseries.  If it were a property and you wanted to make a detector that fired whenever a cert was found to be invalid, regardless of how much time was remaining until expiry, you could do it, but it would be hacky because you would have to just do a count on the number of timeseries for `http.cert_expiry` with a filter for `isValid == "false"`.  If it is a separate timeseries you can just have a detector for `http.cert_valid` with analytics that excludes `value != 0`.  For your `http.cert_expiry` detector, I wouldn't worry about whether the cert is valid or not there, just have two detectors: one that checks expiry time and another that checks validity.  Especially since expiry is really only one aspect of whether a cert is valid; it could be invalid due to many other factors.

Does this make sense?  Is there any other aspect of this in your use case that I have not understood?

xp-1000: it makes sens, thank you for the explanation.
it is little unfortunate we cannot set property from monitor this is exactly what I look for to "log" some contextual / changing information. (i.e. the SSL error for cert_valid metric).

in any case, I just only need to be careful to report those new separate metrics only when it makes sens and this should be easy to create generic detectors (what we do not want is to have regex not matched alert when we do not configure regex for example).

and I prefer to follow your advices so here the changes for code and doc but I need to update tests now.

xp-1000: this time I resolve, everything should be good now @keitwb

keitwb:
Ok, @xp-1000 I pushed a commit to add some more whitespace which I find easier to read, and also changed the `serverName` dimension to `server_name` since our naming convention is to use snake_case instead of camelCase.

Also I tried out the monitor locally and everything seems good.  Let me know if you are good with my changes and I'll merge it after tests pass.

xp-1000:
> Ok, @xp-1000 I pushed a commit to add some more whitespace which I find easier to read, and also changed the `serverName` dimension to `server_name` since our naming convention is to use snake_case instead of camelCase.
> 
> Also I tried out the monitor locally and everything seems good. Let me know if you are good with my changes and I'll merge it after tests pass.

hello @keitwb thanks again for your review.
everything is good for me I will keep in mind both changes for any future PR ;)

PR #1304, opened 2020-04-29
Rework vSphere inventory traversal

pmcollins:
The previous inventory traversal code was not tested against
inventory trees with arbitrary directory placement. This
change generalizes directory traversal with support for more
complex tree structures.

Fixes: INT-1872

### DIFF
--- a/pkg/monitors/vsphere/service/gateway.go
+++ b/pkg/monitors/vsphere/service/gateway.go
@@ -14,7 +14,7 @@ import (
 // A thin wrapper around the vmomi SDK so that callers don't have to use it directly.
 type IGateway interface {
 	retrievePerformanceManager() (*mo.PerformanceManager, error)
-	retrieveTopLevelFolder() (*mo.Folder, error)
+	topLevelFolderRef() types.ManagedObjectReference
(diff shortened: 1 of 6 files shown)


pkg/monitors/vsphere/service/gateway.go
```diff
@@ -89,7 +81,14 @@ func (g *Gateway) queryPerfProviderSummary(ref types.ManagedObjectReference) (*t
 }
 
 func (g *Gateway) queryPerf(invObjs []*model.InventoryObject, maxSample int32) (*types.QueryPerfResponse, error) {
-	specs := make([]types.PerfQuerySpec, 0, len(invObjs))
+	numObjs := len(invObjs)
+	if numObjs == 0 {
+		// empty inventory, return empty response
+		// passing an empty spec to the api causes an error
+		return &types.QueryPerfResponse{}, nil
```
dloucasfx, line 88: Can you add a debug log statement informing the user that inventory is empty and we won't query vsphere ?

pmcollins: Definitely.

PR #1307, opened 2020-05-06
Add pagination to perf queries

pmcollins:
Larger vSphere deployments can fail when trying to
retrieve performance metrics for all inventory objects.
This change adds a `perfBatchSize` config, which
defaults to 10, and splits up requests for performance
data into batches of a number of inventory objects of
that size or smaller.

### DIFF
--- a/pkg/monitors/vsphere/service/gateway.go
+++ b/pkg/monitors/vsphere/service/gateway.go
@@ -14,14 +14,18 @@ import (
 
 // A thin wrapper around the vmomi SDK so that callers don't have to use it directly.
 type IGateway interface {
+	IPerfQuerier
 	retrievePerformanceManager() (*mo.PerformanceManager, error)
 	topLevelFolderRef() types.ManagedObjectReference
 	retrieveRefProperties(mor types.ManagedObjectReference, dst interface{}) error
 	queryAvailablePerfMetric(ref types.ManagedObjectReference) (*types.QueryAvailablePerfMetricResponse, error)
 	queryPerfProviderSummary(mor types.ManagedObjectReference) (*types.QueryPerfProviderSummaryResponse, error)
-	queryPerf(invObjs []*model.InventoryObject, maxSample int32) (*types.QueryPerfResponse, error)
 	retrieveCurrentTime() (*time.Time, error)
 	vcenterName() string
+}
+
+type IPerfQuerier interface {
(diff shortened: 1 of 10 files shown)


pkg/monitors/vsphere/service/paginator.go
```diff
@@ -0,0 +1,57 @@
...
+func (p *queryPerfPaginator) paginate(
+	objs []*model.InventoryObject,
+	maxSample int32,
+) (*types.QueryPerfResponse, error) {
+	numObjs := len(objs)
+	pages := p.numPages(numObjs)
+	start := time.Now()
+	var metrics []types.BasePerfEntityMetricBase
+	for i := 0; i < pages; i++ {
+		startIdx := i * p.pageSize
```
dloucasfx, line 33: I might be missing something, but If pageSize is 0, won't we end up iterating n times (where n is the number of objects in inventory) and end up calling `queryPerf` against the same slice n times?

pmcollins: Yeah I think you're right. That was dumb. Will fix.

pmcollins: Good catch. Should be fixed.

pkg/monitors/vsphere/service/paginator.go
```diff
@@ -0,0 +1,57 @@
...
+		metrics = append(metrics, perf.Returnval...)
+	}
+	end := time.Now()
+	duration := end.Sub(start)
+	p.log.Debugf("Paginator: %d pages took %v", pages, duration)
+	return &types.QueryPerfResponse{Returnval: metrics}, nil
+}
+
+func (p *queryPerfPaginator) numPages(n int) int {
+	// no pagination if page size is zero
```
pmcollins, line 56: I'm going to make another change. All of this special case handling in this file feels wrong to me. It should be handled by the caller.

pkg/monitors/vsphere/service/gateway.go
```diff
@@ -14,16 +14,20 @@ import (
 
 // A thin wrapper around the vmomi SDK so that callers don't have to use it directly.
 type IGateway interface {
+	IPerfQuerier
```
keitwb, line 17: In the future, don't use the Java `I` prefix for interfaces.  Just call it a `PerfQuerier`.

pmcollins: Done.

PR #1312, opened 2020-05-08
Add additional k8s volumes metrics

mstumpfx:


pkg/monitors/kubernetes/volumes/volumes.go
```diff
@@ -110,6 +110,15 @@ func (m *Monitor) getVolumeMetrics() ([]*datapoint.Datapoint, error) {
 			if v.CapacityBytes != nil {
 				dps = append(dps, sfxclient.Gauge("kubernetes.volume_capacity_bytes", dims, int64(*v.CapacityBytes)))
 			}
+			if v.Inodes != nil {
+				dps = append(dps, sfxclient.Gauge("kubernetes.volume_inodes", dims, int64(*v.Inodes)))
```
keitwb, line 114: Use the constants in genmetadata.go (e.g. `kubernetesVolumeInodes`) instead of string literals.  Also replace the existing metrics with constants for consistency.

tests/monitors/kubernetes_volumes/kubernetes_volumes_test.py
```diff
@@ -15,7 +15,10 @@ def test_kubernetes_volumes_in_k8s(k8s_cluster):
          kubeletAPI:
            skipVerify: true
            authType: serviceAccount
+         extraMetrics:
```
keitwb, line 18: Why do you need to put anything in `extraMetrics` since this monitor is configured to send all metrics?

mstumpfx: I changed sendAll to false and made the original metrics default, the new ones will be non-default

Issue #1313, opened 2020-05-08
Add native monitor for DNS


xp-1000:
Hello,

dns resolution check will be useful.

the telegraf plugin provides everything we need and could be imported instead of implement a new one : https://github.com/influxdata/telegraf/tree/master/plugins/inputs/dns_query

keitwb:
This was merged and released in 5.2.1.

PR #1314, opened 2020-05-08
#1313 add dns monitor from telegraf dns_query

xp-1000:
telegraf based plugin for dns. tested it works fine.

https://github.com/signalfx/signalfx-agent/issues/1313


pkg/monitors/telegraf/monitors/dns/metadata.yaml
```diff
@@ -0,0 +1,33 @@
...
+    plugin:
+      description: The plugin name "telegraf/dns".
+    domain:
+      description: The DNS domain queried.
+    server:
+      description: The DNS server queried.
+    record_type:
+      description: The DNS record type queried.
+    result:
+      description: Resolution result string (success = 0, timeout = 1, error = 2).
```
keitwb, line 16: This should be a metric value, not a dimension.

pkg/monitors/telegraf/monitors/dns/metadata.yaml
```diff
@@ -0,0 +1,33 @@
...
+  metrics:
+    dns.query_time_ms:
+      description: Reponse time of the query in ms.
+      default: true
+      type: gauge
+    dns.result_code:
+      description: Resolution result code (success = 0, timeout = 1, error = 2).
+      default: true
+      type: gauge
+    dns.rcode_value:
```
keitwb, line 28: This metric makes me wary because of the same things I was saying before about dimensions and properties.  Because of the `rcode` dimension, this will make a new timeseries for every distinct error that happens in every unique {server, domain, query type} set.

Can you just remove the `rcode` dimension for now (use `emitter.OmitTag("rcode")` I think is enough to do it)?

Make sure the same thing applies to other metrics, where there will be one unique continuous timeseries for each unique {dns server, domain, query type} set for each metric.  This will make alerting much easier overall.

xp-1000: hello @keitwb 

no problem, I think to this myself during my test but I was not sure about a telegraf based monitor if we should keep original implementation or transform to respect signalfx best practices.

Now I know.

xp-1000: Sorry for the delay, I did my changes and tested but forgot to push.

I dropped the `result` and `rcode` dimensions and also the `telegraf_type` one which is always = `untyped`.

I preserved the `record_type` dimension which could change depending of the type of request (`NS` by default). seems useful to separate 2 metrics from the same `server` and `domain` but a different record type in case of multi instantiation.

I rebased the branch, updated doc and tested again.

it this ok for you @keitwb ?

pkg/monitors/telegraf/monitors/dns/metadata.yaml
xp-1000: done

xp-1000:
> Hey @xp-1000, I tried it out and everything seems good. I added one commit to suppress error logs whenever there was an invalid query since it was spammy and we usually reserve agent error log entries for things that are actually wrong with the agent itself, and not with the monitored service (in this case dns server/entry).
> 
> Once all the tests pass I will merge.

@keitwb I seen you approved, everything is ok ? I see the integration tests ci is broken but seems to not come from my monitor, should I do anything or is it ok?

keitwb:
Yeah that failing test is just a bit flaky.  I introduced a lint error so I fixed that and will merge once tests look good.

PR #1316, opened 2020-05-11
Add support for EKS on Fargate

keitwb:
- Add node discovery to the k8s-api observer
 - Upgrade expr to get bugfix that prevented expressions in map values
 - Add config to Helm chart to allow for easy deployments in serverless environments (isServerless value)
 - Add `Getenv` and `Sprintf` functions to discovery rule built in function set (used in the discovery rule for kubelet-stats in the Helm chart).


mstumpfx:
> Add Getenv and Sprintf functions to discovery rule built in function set (used in the discovery rule for kubelet-stats in the Helm chart).

Probably worth noting these features the [auto-discovery](https://github.com/signalfx/signalfx-agent/blob/master/docs/auto-discovery.md) doc at some point. Not necessarily for this PR but definitely noteworthy.

docs/auto-discovery.md
```diff
@@ -16,10 +18,33 @@ container-based observer, you could use the following configuration:
...
+There are currently several target types that can be discovered by the various
+observers in the agent.  You can match these target types explicitly in your
+discovery rules with the expression `target == <type>`, where `<types>` are:
+
+ - `pod`: A Kubernetes pod as a whole.  The `host` field will be populated with
+   the Pod IP address, but no specific `port`.
+
+ - `hostport`: A host and port combination, i.e. a network endpoint.  This
+   endpoint could be on any type of runtime, e.g. container, remote host,
+   straigh on same host with no container.  This type of target will always
```
mstumpfx, line 32:
```suggestion
   running on same host with no container.  This type of target will always
```

PR #1319, opened 2020-05-14
Helm configmap checksum

jphilaine:
To force kubernetes pods replacement when there is only changes in the configmap, we added an checksum of this configmap into pod annotations.

Inspired by helm tips and tricks: https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/templates/daemonset.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/daemonset.yaml
@@ -32,7 +32,10 @@ spec:
         {{ toYaml . | indent 8 | trim }}
         {{- end }}
       annotations:
+        {{ if .Values.podAnnotations }}
         {{- toYaml .Values.podAnnotations | trim | nindent 8 }}
+        {{- end }}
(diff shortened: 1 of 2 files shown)


keitwb:
Hi @jphilaine, by default the agent does a watch (technically a poll) on the ConfigMap data file that K8s maps into the agent container.  When you update a ConfigMap, that file eventually updates and the agent reconfigures itself by diffing the new and old config and seeing what components in the agent need to change.  The intent with that is to minimize any downtime of monitoring and to keep a continuous flow of datapoints/spans with minimum disruption.  It also provides a bit of protection against bad config changes, as the agent will reject any new config that could have caused a newly created agent to fail to start up in the first place.

What is your reasoning for recreating pods when config changes as opposed to just letting the running agents reconfigure themselves?

jphilaine:
Hi @keitwb ,

After more tests, the file agent.yaml from configmap is updated after a configmap update, but it is not instant. That requires a few seconds to be updated. In that case, if the agent implements a watch on the configmap, I am agree with you : it is better to keep the agent to prevent downtime. You can reject my PR.

The delay between configmap update and real application by agent 
misled me.

Regards.

PR #1320, opened 2020-05-14
add supervisor monitor

xp-1000:
hello,

a very simple monitor for supervisor.

I see 2 different approach to check processes states:
- define a metric per status which count the number of processes for each one
- define only one metric state and report use the name of the process for its dimension

The first way has the advantage to always report only 3 metrics per monitor instance no matter the number of processes in supervisor.

The second way could report more metrics when there are lot of porcesses but will report less metrics for a simple supervisor config (1 or 2 processes).

I choose the second one because it brings more context allowing to know the process name affected directly from alert.


tests/monitors/supervisor/supervisor_test.py
```diff
@@ -0,0 +1,38 @@
...
+pytestmark = [pytest.mark.monitor_with_endpoints]
+
+
+METADATA = Metadata.from_package("supervisor")
+PORT = 9001
+PROCESS = "long_script"
+
+
+@contextmanager
+def run_php_fpm():
```
keitwb, line 20: Shouldn't this be `run_supervisor`?  Same for a few things below.

xp-1000: oh good catch sorry ^^ fixed

PR #1321, opened 2020-05-15
Splunk logging

atoulme:
Adds the ability to send data to Splunk HEC endpoints, SignalFX, or both.
Change the helm chart definition to use new configuration for Splunk HEC.


pkg/core/config/writer.go
```diff
@@ -110,6 +110,20 @@ type WriterConfig struct {
 	// handle the volume of trace spans and should be upgraded to more powerful
 	// hardware/networking.
 	MaxTraceSpansInFlight uint `yaml:"maxTraceSpansInFlight" default:"100000"`
+	// Enable logging to a Splunk Enterprise instance
+	LogSplunkEnabled bool `yaml:"splunkEnabled"`
```
keitwb, line 114: The fact that all of these config values start with `splunk` makes me think they should be put into a common config struct nested within this struct.

pkg/core/writer/splunk/splunk.go
```diff
@@ -0,0 +1,221 @@
...
+	SkipTLSVerify bool
+	Errors        chan error
+	once          sync.Once
+	events        chan splunkMessage
+	sendQueue     chan []splunkMessage
+}
+
+func (h *Handler) init() {
+	if h.HTTPClient == nil {
+		tr := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: h.SkipTLSVerify}}
```
jrcamp, line 37: Use https://github.com/signalfx/signalfx-agent/blob/master/pkg/core/common/httpclient/http.go which has a common config struct and building for HTTP clients.

Using transport without setting things like MaxIdleConns can cause issues so it's best to clone the default transport which the builder does.

atoulme: Thanks for the review! I have added a commit following your recommendation.

deployments/docker/examples/splunk/README.md
```diff
@@ -0,0 +1,17 @@
...
+The example runs as a Docker Compose deployment. The agent can be configured to send various metrics to Splunk Enterprise.
+
+Splunk is configured to receive data from the SignalFX Agent using the HTTP Event collector. To learn more about HEC, visit [our guide](https://dev.splunk.com/enterprise/docs/dataapps/httpeventcollector/).
+
+To deploy the example, open a terminal and in this directory type:
+```bash
+$> docker-compose up
+```
+:
+Splunk will become available on port 8000. You can login on [http://localhost:18000] with `admin` and `changeme`.
```
j4ys0n, line 14: change `Splunk will become available on port 8000.` to `Splunk will become available on port 18000.`

atoulme: Thanks good catch!

keitwb:
_SignalFx_ (with lowercase _x_) is the standard way to spell it.

atoulme:
Fixed the SignalFX->SignalFx typo.

pkg/core/writer/splunk/splunk.go
```diff
@@ -0,0 +1,240 @@
...
+					oldEvents := eventsToSend
+					eventsToSend = make([]interface{}, 0)
+					h.sendQueue <- oldEvents
+				}
+			}
+		}
+	}()
+	go func() {
+		for {
+			events := <-h.sendQueue
```
keitwb, line 90: This sending model is pretty similar to the original model we used for the agent (buffer a bunch of data, and send them out after a certain batch size is hit or after a certain period of time has passed).  We have since switched to something a bit more sophisticated since then due to limitations that we had with that at high volumes of data generation.

This for loop here can only handle sending one batch at a time since logEvents is synchronous.  That means you are primarily limited by the RTT of the request to Splunk, which is uncontrollable by the agent. Also since you are limiting the batch size to 50 that means you have a max event rate of `1/<RTT in seconds> * 50` events per second.

We can discuss what the other writer in the agent does and whether it is appropriate for this when we talk later today.

pkg/core/writer/splunk/splunk.go
```diff
@@ -0,0 +1,240 @@
...
+		Index:      h.index,
+		Event:      eventdata{Properties: props, Dimensions: e.Dimensions, Meta: meta, EventType: e.EventType, Category: e.Category},
+	}
+	h.events <- newEvent
+}
+
+func (h *Writer) logEvents(events []interface{}) {
+	buf := new(bytes.Buffer)
+	for _, e := range events {
+		b, err := json.Marshal(e)
```
keitwb, line 194: I think it might be more efficient to use https://golang.org/pkg/encoding/json/#Encoder and pass in the buffer so that it doesn't have to allocate the separate byte array on heap.

pkg/core/writer/splunk/splunk.go
```diff
@@ -0,0 +1,240 @@
...
+		if err != nil {
+			log.WithFields(log.Fields{
+				"error": err,
+			}).Error("Error marshaling data to Splunk")
+		} else {
```
keitwb, line 199: Usual pattern is:

```go
if err != nil {
    // log error
    return
}
//happy path logic
```

pkg/core/writer/splunk/splunk.go
atoulme: OK, changed.

pkg/core/writer/splunk/splunk.go
atoulme: OK, changed as well.

deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -2,14 +2,39 @@
...
+splunkSourceType: ""
+
+# Splunk index, optional name of the Splunk index to store the event in
+splunkIndex: ""
+
+# Skip verifying the certificate of the HTTP Event Collector
+splunkSkipTLSVerify: false
+
+# The hostname of the agent sent to Splunk
+splunkHost: "signalfx-agent"
```
keitwb, line 36: SignalFx datapoints and events generated by the agent usually have a `host` dimension that the 
Splunk writer could introspect and use.  Regardless, we probably shouldn't hard code this name in the Helm chart unless _host_ means something different in Splunk.

pkg/core/writer/spans.go
```diff
@@ -18,17 +18,18 @@ func (sw *SignalFxWriter) sendSpans(ctx context.Context, spans []*trace.Span) er
...
 
-	// This sends synchonously
-	err := sw.client.AddSpans(context.Background(), spans)
-	if err != nil {
-		log.WithFields(log.Fields{
-			"error": err,
-		}).Error("Error shipping spans to SignalFx")
-		// If there is an error sending spans then just forget about them.
-		return err
+	if sw.client != nil {
```
keitwb, line 21: Remove this conditional since the Splunk writer isn't going to deal with trace spans at this point.

pkg/core/writer/writer.go
```diff
@@ -97,17 +100,23 @@ func New(conf *config.WriterConfig, dpChan chan []*datapoint.Datapoint, eventCha
 
 	ctx, cancel := context.WithCancel(context.Background())
 
-	dimensionClient, err := dimensions.NewDimensionClient(ctx, conf)
-	if err != nil {
-		cancel()
-		return nil, err
+	var dimensionClient *dimensions.DimensionClient
+	var err error
+	if conf.SignalFxAccessToken != "" {
```
keitwb, line 105: There are many legitimate configs where the access token is blank but we still want to send (e.g. where the agent sends to a central gateway/proxy that adds the token).  Our Helm chart required it in K8s but what you did to make it optional was appropriate and should be done to make it consistent with the variety of deployment models.

You should conditionalize this based on the presence of `conf.APIUrl`.

pkg/core/writer/writer.go
```diff
@@ -97,17 +100,23 @@ func New(conf *config.WriterConfig, dpChan chan []*datapoint.Datapoint, eventCha
...
+			return nil, err
+		}
 	}
 
 	var correlationClient correlations.CorrelationClient
-	correlationClient, err = correlations.NewCorrelationClient(ctx, conf)
-	if err != nil {
-		cancel()
-		return nil, err
+	if conf.SignalFxAccessToken != "" {
```
keitwb, line 114: Same thing about `conf.APIUrl`.

pkg/core/writer/writer.go
```diff
@@ -135,72 +144,89 @@ func New(conf *config.WriterConfig, dpChan chan []*datapoint.Datapoint, eventCha
...
-	dpEndpointURL, err := conf.ParsedIngestURL().Parse("v2/datapoint")
-	if err != nil {
-		log.WithFields(log.Fields{
-			"error":     err,
-			"ingestURL": conf.ParsedIngestURL().String(),
-		}).Error("Could not construct datapoint ingest URL")
-		return nil, err
-	}
-	sw.client.DatapointEndpoint = dpEndpointURL.String()
+	if conf.SignalFxAccessToken != "" {
```
keitwb, line 148: Take out this conditional and make this more granular.

pkg/core/writer/writer.go
```diff
@@ -135,72 +144,89 @@ func New(conf *config.WriterConfig, dpChan chan []*datapoint.Datapoint, eventCha
...
 	sw.datapointFilters, err = sw.conf.DatapointFilters()
 	if err != nil {
 		return nil, err
 	}
 
-	sw.dimensionClient.Start()
-	sw.correlationClient.Start()
-
-	go sw.listenForEventsAndDimensionUpdates()
+	if conf.SignalFxAccessToken != "" {
```
keitwb, line 225: Change this conditional to be based on whether the sub clients are nil or not.

pkg/core/writer/writer.go
atoulme: The helm chart offers to set `APIUrl`, `IngestURL`, etc or rely on `realm`, which has a default value of `us0`. My impression from that is that `APIUrl` will never be empty as result. What's the appropriate move there?

pkg/core/writer/writer.go
```diff
@@ -230,10 +252,13 @@ func New(conf *config.WriterConfig, dpChan chan []*datapoint.Datapoint, eventCha
...
-	log.Infof("Sending datapoints to %s", sw.client.DatapointEndpoint)
-	log.Infof("Sending events to %s", sw.client.EventEndpoint)
-	log.Infof("Sending trace spans to %s", sw.client.TraceEndpoint)
-
+	if conf.SignalFxAccessToken != "" {
+		log.Infof("Sending datapoints to %s", sw.client.DatapointEndpoint)
+		log.Infof("Sending events to %s", sw.client.EventEndpoint)
+		log.Infof("Sending trace spans to %s", sw.client.TraceEndpoint)
+	} else {
+		log.Infof("SignalFx client is deactivated, provide a token to enable")
```
keitwb, line 260: Should be enabled by providing the realm and/or ingest/event/trace urls.

pkg/core/writer/spans.go
atoulme: That's fine. Note: I have another branch built on this one that exports spans too: https://github.com/atoulme/signalfx-agent/tree/export_traces_too

pkg/core/writer/writer.go
keitwb: They would need to override it to be null in the Helm values they provide to disable SignalFx sending.

pkg/core/writer/writer.go
atoulme: Done.

deployments/k8s/helm/signalfx-agent/values.yaml
atoulme: I have to check this. I do believe the host property is the host of the collector, not the host you pick values from. The host associated with the metric is added as dimension.

Checking that assumption in the docs and coming up empty, but will keep looking.

pkg/core/writer/writer.go
atoulme: Done

pkg/core/writer/writer.go
atoulme: Alright, message is fixed.

pkg/core/writer/writer.go
atoulme: OK I have changed the helm chart to disable signalfx sending.

deployments/k8s/helm/signalfx-agent/values.yaml
atoulme: Host is the origin of the event, and best passed in with hostIDDims looks like. I am going to use the "host" dimension for each event or data point for now to pass in. Let me know if that works.

pkg/core/writer/splunk/splunk.go
atoulme: Definitely open to anything - and it looks like you have a better model for a writer in hand, so let's use that.

deployments/docker/examples/splunk/README.md
```diff
@@ -0,0 +1,17 @@
+# SignalFX agent Splunk Docker-compose example
```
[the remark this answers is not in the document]
atoulme: Done.

pkg/core/config/writer.go
```diff
@@ -201,3 +237,32 @@ func (wc *WriterConfig) DefaultTraceEndpointPath() string {
...
+// SplunkConfig configures the writer specifically writing to Splunk.
+type SplunkConfig struct {
+	// Enable logging to a Splunk Enterprise instance
+	Enabled bool `yaml:"enabled"`
+	// url to Splunk HTTP Event Collector
+	URL string `yaml:"url"`
+	// Splunk HTTP Event Collector token
+	Token string `yaml:"token"`
+	// Splunk source field value, description of the source of the event
+	Source string `yaml:"source"`
```
keitwb, line 250: @atoulme should this field `source` be required or is it also optional?

atoulme: It can be optional. The user can set a default source.

PR #1324, opened 2020-05-15
Send vSphere points as they are created

pmcollins:
The previous implementation collected and accumulated all metrics for
the entire vSphere inventory before sending them out all at once. This
change sends out each datapoint right after it's created, reducing
memory usage and giving the system an opportunity to GC while getting
metrics.


pkg/monitors/vsphere/service/multi_perf_fetcher_test.go
```diff
@@ -29,18 +29,27 @@ func TestNumPages(t *testing.T) {
...
 }
 
-func TestPagination1(t *testing.T) {
-	p := queryPerfPaginator{pageSize: 2, gateway: &fakePaginatorGateway{}, log: getTestingLog()}
-	n := 100
-	resp, _ := p.queryPerf(invObjs(n), 1)
-	require.Equal(t, n, len(resp.Returnval))
+func TestMultiPage1(t *testing.T) {
+	p := multiPagePerfFetcher{pageSize: 4, gateway: &fakePaginatorGateway{}, log: testLog}
+	pInstance := p.invIterator(invObjs(4), 1)
```
pmcollins, line 34: oops, forgot to update the variable name

pkg/monitors/vsphere/runner.go
```diff
@@ -36,8 +34,7 @@ func (r *runner) run() {
 		r.log.WithError(err).Error("firstTimeSetup failed")
 		return
 	}
-	dps := r.vsm.retrieveDatapoints()
-	r.monitor.Output.SendDatapoints(dps...)
+	r.vsm.retrieveDatapoints()
```
keitwb, line 37: Call this something like `r.vsm.Run()` or something like that.  It isn't really retrieving datapoints, it is doing a bunch of stuff to generate them.

keitwb: The main thing is that this function is called entierly for its side effect so its name should reflect that.

pmcollins: Agree, but couldn't think of a name so I had left it. Changed it to `generateDatapoints`

PR #1330, opened 2020-05-26
[Helm] Enabling dnsConfig option in the Helm Chart

vladiceanu:
#### What this PR does / why we need it:
Adding possibility to set a custom [dnsConfig](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config) for the pods. Pod's DNS Config field enable users more control on the DNS settings for the pods.

#### Use case
SignalFx pods with `/etc/resolv.conf ndots:5` (default) and output `ingest.eu0.signalfx.com`,  syscall will try to resolve it sequentially going through all local search domains first and then resolve the absolute name only at last, which puts load on the DNS resolver and impacts the performance of the applications. Tuning `dnsConfig` for SignalFx pod allows us to avoid such scenarios and resolve the absolute name at first and cache the positive response.

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/Chart.yaml
+++ b/deployments/k8s/helm/signalfx-agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: The SignalFx Smart Agent
 name: signalfx-agent
 appVersion: 5.2.1
-version: 1.0.0
+version: 1.0.1
 keywords:
 - monitoring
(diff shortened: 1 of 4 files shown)


keitwb:
Thanks for this @vladiceanu.  I'll release the Helm chart to our repo in a bit.

keitwb:
Ok, was released to the repo as version 1.0.1.

PR #1335, opened 2020-05-28
add missing couchbase metrics to metadata.yaml

mstumpfx:


mstumpfx:
Hmm something seems to be wrong with the format of the `metadata.yaml` file that our code-gen doesn't like. @keitwb do you see anything obvious? The error message isn't too useful -
```FATA[0000] failed to format source: 111:25: expected ';', found '&' (and 13 more errors) ```

File: pkg/monitors/collectd/couchbase/metadata.yaml [hunk 1]
```diff
@@ -59,313 +161,1202 @@ monitors:
...
+      default: false
+      type: gauge
+      group: bucket
+    gauge.bucket.op.ep_dcp_views+2i_items_remaining:
+      description: Number of indexes items remaining to be sent.
+      default: false
+      type: gauge
+      group: bucket
+    gauge.bucket.op.ep_dcp_xdcr_total_backlog_size:
+      description: description
```
keitwb, line 374: ?

pkg/monitors/collectd/couchbase/metadata.yaml
```diff
@@ -59,313 +161,1202 @@ monitors:
...
+      default: false
+      type: gauge
+      group: bucket
+    gauge.nodes.mcdMemoryAllocated:
+      description: Amount of memcached memory allocated (bytes).
+      default: false
+      type: gauge
+      group: nodes
+    gauge.storage.hdd.total:
+      description: Total harddrive space available to cluster (bytes)
```
keitwb, line 498: The way you did this one is right, this is not a sentence and so it shouldn't end in a period.

keitwb:
I'm guessing it has something to do with the genmetadata.go file that wasn't committed here (but should be).

mstumpfx:
@keitwb the error was some invalid metric names containing `+` in them. I've cleaned it up now. 
Test failures look related the to k8s 1.18 issue, but unrelated to my changes.

PR #1336, opened 2020-05-28
1309 agent style updates

galpizarsfx:
Made style and content changes to the smart agent quick install guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,8 @@ You can [view the source for the installer script](./deployments/installer/insta
 and use it on your hosts by running:
 
 ```sh
-curl -sSL https://dl.signalfx.com/signalfx-agent.sh > /tmp/signalfx-agent.sh
-sudo sh /tmp/signalfx-agent.sh --realm YOUR_SIGNALFX_REALM -- YOUR_SIGNALFX_ACCESS_TOKEN
+curl -sSL https://dl.signalfx.com/signalfx-agent.sh > /tmp/signalfx-agent.sh;
+sudo sh /tmp/signalfx-agent.sh --realm YOUR_SIGNALFX_REALM -- YOUR_SIGNALFX_API_TOKEN
 ```
 
 ##### Windows
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# SignalFx Smart Agent 
+# SignalFx Smart Agent
 
 [![GoDoc](https://godoc.org/github.com/signalfx/signalfx-agent?status.svg)](https://godoc.org/github.com/signalfx/signalfx-agent)
 [![CircleCI](https://circleci.com/gh/signalfx/signalfx-agent.svg?style=shield)](https://circleci.com/gh/signalfx/signalfx-agent)
@@ -204,7 +204,7 @@ compatible.
 2) Ensure a valid configuration file is available somewhere on the target
 system.  The main thing that the distro packages provide -- but that you will
 have to provide manually with the bundle -- is a run directory for the agent to
-use.  Since you aren't installing from a package, there are three config 
+use.  Since you aren't installing from a package, there are three config
 options that you will especially want to consider:
 
  - `internalStatusHost` - This is the host name that
@@ -226,7 +226,7 @@ options that you will especially want to consider:
 	 pick up stale collectd config, so be sure that it is not used for anything
 	 else.  Also note that **these files could have sensitive information in
 	 them** if you have passwords configured for collectd monitors, so you
-	 might want to place this dir on a `tmpfs` mount to avoid credentials 
+	 might want to place this dir on a `tmpfs` mount to avoid credentials
 	 persisting on disk.
 
 See the section on [Privileges](#privileges) for information on the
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -100,8 +100,8 @@ You can [view the source for the installer script](./deployments/installer/insta
 and use it on your hosts by running:
 
 ```sh
-curl -sSL https://dl.signalfx.com/signalfx-agent.sh > /tmp/signalfx-agent.sh
-sudo sh /tmp/signalfx-agent.sh --realm YOUR_SIGNALFX_REALM -- YOUR_SIGNALFX_ACCESS_TOKEN
```
jchengsfx, line 104: I recently changed this to `YOUR_SIGNALFX_ACCESS_TOKEN` since we require the access token for sending datapoints, which is different than the API token.

docs/quick-install.md
```diff
@@ -1,82 +1,111 @@
...
 
-1. To confirm your installation, enter the following command on the Linux or Windows command line:
+**For Linux:** Use `terminal` or a similar app 
 
-    ```sh
-    sudo signalfx-agent status
-    ```
+**For Windows:** Use Windows PowerShell
+
+`sudo signalfx-agent status`
```
jchengsfx, line 103: This command is only applicable on Linux.  For Windows, the PowerShell command should be:
`& "\Program Files\SignalFx\SignalFxAgent\bin\signalfx-agent" status`

README.md
mstumpfx: Just FYI - `YOUR_SIGNALFX_API_TOKEN` is a keyword for the UI to auto-replace with the current user's API token. `YOUR_SIGNALFX_ACCESS_TOKEN` has no auto-replace functionality, so maybe there is a better option than using this text. Although it does get the point across.

docs/quick-install.md
```diff
@@ -1,82 +1,111 @@
...
+### Install SignalFx Smart Agent on Windows
 
+1. Run Windows PowerShell.
+
+2. To configure Windows execution policy, download the Smart Agent install script, and run it, copy the following commands and paste them into the Windows PowerShell window:
 
 ```sh
-& {Set-ExecutionPolicy Bypass -Scope Process -Force; $script = ((New-Object System.Net.WebClient).DownloadString('https://dl.signalfx.com/signalfx-agent.ps1')); $params = @{access_token = "YOUR_SIGNALFX_API_TOKEN"; ingest_url = "https://ingest.YOUR_SIGNALFX_REALM.signalfx.com"; api_url = "https://api.YOUR_SIGNALFX_REALM.signalfx.com"}; Invoke-Command -ScriptBlock ([scriptblock]::Create(". {$script} $(&{$args} @params)"))}
+& {Set-ExecutionPolicy Bypass -Scope Process -Force; $script = ((New-Object System.Net.WebClient).DownloadString(‘https://dl.signalfx.com/signalfx-agent.ps1’)); $params = @{access_token = "YOUR_SIGNALFX_API_TOKEN"; ingest_url = "https://ingest.YOUR_SIGNALFX_REALM.signalfx.com"; api_url = "https://api.YOUR_SIGNALFX_REALM.signalfx.com"}; Invoke-Command -ScriptBlock ([scriptblock]::Create(“. {$script} $(&{$args} @params)”))}
```
jchengsfx, line 88: Are the changes to the single and double quotes in the command necessary? Seems like it could cause issues if a user copies and pastes this command.

README.md
galpizarsfx: I'm a little confused. The following is what we currently have in the doc and UI, and it seems to populate content correctly. Does this need to change? 

`--realm YOUR_SIGNALFX_REALM -- YOUR_SIGNALFX_API_TOKEN`

` --realm us0 -- rb2yT1_ej4yXYkIdSQzp_w`

docs/quick-install.md
galpizarsfx: I'll make this change now.

docs/quick-install.md
galpizarsfx: We received a ticket that this was a bug, so we made the changes listed here: https://signalfuse.atlassian.net/browse/DOCS-1432

README.md
jchengsfx: > Just FYI - `YOUR_SIGNALFX_API_TOKEN` is a keyword for the UI to auto-replace with the current user's API token. `YOUR_SIGNALFX_ACCESS_TOKEN` has no auto-replace functionality, so maybe there is a better option than using this text. Although it does get the point across.

Thanks @mstumpfx! I was told to change it in my last PR and didn't realize it was a keyword for auto-replacement.

docs/quick-install.md
jchengsfx: Thanks @galpizarsfx, didn't realize the quotes were getting rendered differently in the UI.

README.md
jchengsfx: > I'm a little confused. The following is what we currently have in the doc and UI, and it seems to populate content correctly. Does this need to change?
> 
> `--realm YOUR_SIGNALFX_REALM -- YOUR_SIGNALFX_API_TOKEN`
> 
> ` --realm us0 -- XXXXXXXXXXX`

@galpizarsfx Keep your changes. It should be `YOUR_SIGNALFX_API_TOKEN`.

mstumpfx: `YOUR_SIGNALFX_API_TOKEN` should populate correctly with the current user's API token when logged in and viewing inside the application.

However, it sounds like when running the install script we need an _access_ token, not the api token. Which would mean we are incorrectly substituting here. But I think this is the part I'm unsure of - if the token that is displayed is valid or not for the install script. It sounds like @jchengsfx was saying we cannot use an API token here.

`YOUR_SIGNALFX_ACCESS_TOKEN` does not auto-replace with anything and would be displayed as is. But this could be valid way to display this message still, since we need the user to manually substitute with an access token.

mstumpfx: Okay just kidding, `YOUR_SIGNALFX_API_TOKEN` actually is being replaced by the orgs default access token, _not_ the user API token. My bad. Keep your changes, all good.

pohannigansfx:
merging based on approval from Brian Keith

PR #1338, opened 2020-05-29
Add kubelet-summary monitor to replace kubelet-stats

keitwb:
kubelet-stats uses endpoints that are now deprecated in K8s 1.18 and will
be disabled permanently in a future release.

Apparently the /stats/summary endpoint is being left and is what the metrics server in
K8s still relies on so it will probably be safe for a while.

Also updated the Helm chart to conditionally use the new monitor in
place of kubelet-stats if the agent is sufficiently new.


pkg/monitors/kubernetes/kubeletmetrics/metadata.yaml
```diff
@@ -0,0 +1,120 @@
+monitors:
+- monitorType: kubelet-metrics
+  doc: |
+    This monitor pulls container metrics from a Kubernetes kubelet instance via
+    the `/stats/summary` endpoint.
+
+    This montior is intended to replace the `kubelet-stats` monitor that
+    replied on now deprecated endpoints in the kubelet.  It does not get as
```
pmcollins, line 8: relied on

pkg/monitors/kubernetes/kubeletmetrics/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+				}
+				if i.RxErrors != nil {
+					dps = append(dps, sfxclient.Cumulative(podNetworkReceiveErrorsTotal, intfDims, int64(*i.RxErrors)))
+				}
+				if i.TxErrors != nil {
+					dps = append(dps, sfxclient.Cumulative(podNetworkTransmitErrorsTotal, intfDims, int64(*i.TxErrors)))
+				}
+			}
+		}
+
```
pmcollins, line 130: Do you think pod memory and pod cpu metrics could be worth sending?

pmcollins: Ditto for node-level metrics (cpu, fs, memory, network), or is that expected to be sent via another monitor?

keitwb: My understanding is that pod cpu/memory metrics are just aggregates of the individual container metrics and thus would be redundant -- you could just sum up the containers to get the pod.  Have you seen anything that says differently?

keitwb: > Ditto for node-level metrics (cpu, fs, memory, network), or is that expected to be sent via another monitor?

Those are all sent by the host monitors (e.g. `memory`, `cpu`, etc.).

PR #1339, opened 2020-05-30
allows host and port config on php and supervisor

xp-1000:
Currently, php fpm and supervisor montiors only accept url as config option.

Adding host and port will ease the config in discovery setups from:

```
  - type: collectd/php-fpm
    discoveryRule: container_image =~ "php"
    configEndpointMappings:
      url: 'Sprintf("http://%v:%v/fpm-status?json", host, port)'
```

to:

```
  - type: collectd/php-fpm
    discoveryRule: container_image =~ "php"
```

also improved dns configuration while domains could be optional

### DIFF
--- a/docs/monitors/collectd-php-fpm.md
+++ b/docs/monitors/collectd-php-fpm.md
@@ -51,7 +51,7 @@ monitors:
  - type: collectd/php-fpm
    host: localhost
    port: 80
-   url: "http://{{.Host}}:{{.Port}}/fpm-status?json"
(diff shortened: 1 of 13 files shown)


keitwb:
Hey @xp-1000, I meant to bring up these things in code review when you originally did these but forgot. Thanks for adding these.

I made a new changes to make the config template for collectd/php-fpm more consistent with the config names, and also added the ability to manipulate path/scheme in the supervisor monitor.  Take a look and let me know what you think.

PR #1347, opened 2020-06-08
Add checksum for ConfigMap to Pod Annotations

nagas:
This will change the Deployment/DaemonSet object on ConfigMap change which will in turn trigger deployment/restart of SignalFx Agent.


keitwb:
Hi @nagas, see https://github.com/signalfx/signalfx-agent/pull/1319 and the discussion there about why this isn't really necessary.  Additional questions/input is welcome.

nagas:
Thanks, @keitwb! Makes sense. I checked for similar PRs, but apparently my search skills are far from perfect :)

PR #1349, opened 2020-06-08
Add ability to export spans to Splunk

atoulme:
* Allow to specify an event index alongside the metrics index
* Send events and spans as events to the Splunk HEC.
* Add an example of use to the docker-compose example.

### DIFF
--- a/pkg/core/config/writer.go
+++ b/pkg/core/config/writer.go
@@ -249,11 +249,17 @@ type SplunkConfig struct {
 	// Splunk HTTP Event Collector token
 	Token string `yaml:"token"`
 	// Splunk source field value, description of the source of the event
-	Source string `yaml:"source"`
+	MetricsSource string `yaml:"source"`
 	// Splunk source type, optional name of a sourcetype field value
-	SourceType string `yaml:"sourceType"`
+	MetricsSourceType string `yaml:"sourceType"`
 	// Splunk index, optional name of the Splunk index to store the event in
-	Index string `yaml:"index"`
+	MetricsIndex string `yaml:"index"`
+	// Splunk index, specifically for traces (must be event type)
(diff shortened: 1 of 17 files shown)


keitwb:
Specific lint failures are at https://dev.azure.com/signalfx/signalfx-agent/_build/results?buildId=3627&view=logs&j=18a12b53-e25b-5b79-411d-9a29f3a83d7c&t=becae859-8d01-508d-6d9d-600a82ae380c&l=12.

pkg/core/config/writer.go
```diff
@@ -249,11 +249,17 @@ type SplunkConfig struct {
...
-	Source string `yaml:"source"`
+	MetricsSource string `yaml:"source"`
 	// Splunk source type, optional name of a sourcetype field value
-	SourceType string `yaml:"sourceType"`
+	MetricsSourceType string `yaml:"sourceType"`
 	// Splunk index, optional name of the Splunk index to store the event in
-	Index string `yaml:"index"`
+	MetricsIndex string `yaml:"index"`
+	// Splunk index, specifically for traces (must be event type)
+	EventsIndex string `yaml:"eventsIndex"`
```
keitwb, line 258: This name and description is pretty confusing because in SignalFx _events_ are a thing distinct from traces.  If this is just saying which index traces go into can't it just be called something like `SpanIndex`?

atoulme: Splunk has two types of indexes: metrics and events. This relates more to the type of data stored in Splunk, and that needs to be disambiguated before data is ingested.

Events and traces will go to an event-type index. Metrics will go to a metrics index.

atoulme:
Should be better now. Still failing on links, but that looks unrelated.

PR #1357, opened 2020-06-15
Add haproxy_session_limit and haproxy_session_max metrics

cvauvarin:
Hello,

A small PR to add two haproxy metrics :
* `haproxy_session_limit`
* `haproxy_session_max`

The limit metric is really useful and seems important to me to monitor and compare it with `haproxy_session_current`. It is defined by the `maxconn` directive (see https://cbonte.github.io/haproxy-dconv/2.1/configuration.html#maxconn%20(Server%20and%20default-server%20options)).

Thanks.


keitwb:
Thanks for submitting this @cvauvarin.  We will release it soon.

PR #1358, opened 2020-06-15
Docs - Change kubernetes_annotations type to map of string

trotman23:
Simple docs update to reflect the correct type of the `kubernetes_annotations` target variable for the `k8s-api` observer.


keitwb:
Hi @trotman23, the doc you modified is automatically generated from agent metadata so this won't work.  Let me look into how to fix it at the root problem and I'll update this PR.  Thanks for this.

keitwb:
Ok, I fixed a bunch of them.  Should be good to go once tests pass. Thanks for bringing this up.

PR #1361, opened 2020-06-16
processlist monitor: Change double quotes to single quotes in process…

keitwb:
… string fields

The backend cannot handle even escaped double quotes.

### DIFF
--- a/pkg/monitors/processlist/processlist.go
+++ b/pkg/monitors/processlist/processlist.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"math"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/signalfx/golib/v3/event"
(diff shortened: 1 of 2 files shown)


pkg/monitors/processlist/processlist.go
```diff
@@ -160,9 +161,9 @@ func (m *Monitor) encodeProcess(proc *TopProcess, sampleInterval time.Duration)
 		nice = strconv.Itoa(*proc.Nice)
 	}
 
-	return fmt.Sprintf("\"%d\":[\"%s\",%d,\"%s\",%d,%d,%d,\"%s\",%.2f,%.2f,\"%s\",\"%s\"]",
```
tedoc2000, line 163: Don't you need \\ to send a literal '\'?

keitwb: I replaced the outer quotes for the first are to Sprintf with backticks, which are the raw string delimiter in Go, so you don't need escapes.  The change on this specific line doesn't actually change anything functionally, it just changes the quoting style of the string so it is easier to read.

tedoc2000: Is there some reason we don't just string.ReplaceAll(..., `"`, `\\"`)?

keitwb: I tried escaping the `"` with both one and two backslashes and our backend still choked on it for some reason (it didn't seem to recognize the escapes).  I confirmed the output to the backend multiple times and it was definitely not liking any double quotes at all for some reason.

tedoc2000: Can you paste some of the stuff that failed?

PR #1363, opened 2020-06-18
Enable SlaveStats option on MySQL plugin

cvauvarin:
According to the [MySQL plugin documentation](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_mysql) the `SlaveStats` can be used to report information on the slave state and wether there is some delay or not compared to the master.

Since this PR https://github.com/collectd/collectd/pull/3463/files, slave threads informations are now reported as gauge and not only as notification, it seems that this PR is not yet included in https://github.com/signalfx/collectd/ and so not available in the agent (according to my tests) . Do you think it is possible to include those changes in the signalfx collectd process ?

This PR allows the reporting of those three new metrics once they will be available in the agent.

Thanks.

EDIT : also added the master stats option to have the full support of the plugin on replication.

### DIFF
--- a/docs/monitors/collectd-mysql.md
+++ b/docs/monitors/collectd-mysql.md
@@ -90,6 +90,7 @@ Configuration](../monitor-config.md#common-configuration).**
 | `password` | no | `string` |  |
 | `reportHost` | no | `bool` | A SignalFx extension to the plugin that allows us to disable the normal behavior of the MySQL collectd plugin where the `host` dimension is set to the hostname of the MySQL database server.  When `false` (the recommended and default setting), the globally configured `hostname` config is used instead. (**default:** `false`) |
 | `innodbStats` | no | `bool` |  (**default:** `false`) |
+| `slaveStats` | no | `bool` |  (**default:** `false`) |
 
 
 The **nested** `databases` config object has the following fields:
@@ -314,9 +315,12 @@ Metrics that are categorized as
  - `mysql_sort.scan` (*cumulative*)<br>    The number of sorts that were done by scanning the table.
  - `mysql_sort_merge_passes` (*cumulative*)<br>    The number of merge passes done by the sorting algorithm.
  - `mysql_sort_rows` (*cumulative*)<br>    The number of rows that were sorted.
+ - `slave-io-running` (*gauge*)<br>    Whether the I/O thread for reading the master's binary log is running.
+ - `slave-sql-running` (*gauge*)<br>    Whether the SQL thread for executing events in the relay log is running.
  - ***`threads.cached`*** (*gauge*)<br>    The number of threads cached by MySQL for re-use on a new client connection.  A MySQL thread corresponds to a single MySQL connection.
  - ***`threads.connected`*** (*gauge*)<br>    The number of currently open MySQL connections.  A MySQL thread corresponds to a single MySQL connection.
  - `threads.running` (*gauge*)<br>    The number of MySQL threads that are processing a query.  A MySQL thread corresponds to a single MySQL connection.
+ - `time-offset` (*gauge*)<br>    The number of seconds that the slave SQL thread is behind processing the master binary log.
  - `total_threads.created` (*cumulative*)<br>    The total number of threads created by MySQL for client connections.  A MySQL thread corresponds to a single MySQL connection.
 
 ### Non-default metrics (version 4.7.0+)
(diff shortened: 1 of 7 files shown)


keitwb:
Hi @cvauvarin, getting a new collectd release is not trivial given all the patches we have on it and we are moving away from the collectd side of the agent anyway.  I am going to implement an enhancement to the `sql` monitor that will allow doing exactly what you want and will probably get around to it later this week.  I'll keep you posted.

cvauvarin:
Ok thanks. Does it mean that any new metric than is not included in the `collectd` plugin will now be added into the `sql` monitor ? As I understand for now the `sql` monitor only allows to run query but does not report any metrics by default. Or they will be a full rewrite of the plugin MySQL without using `collectd` ?

keitwb:
> Ok thanks. Does it mean that any new metric than is not included in the `collectd` plugin will now be added into the `sql` monitor ? As I understand for now the `sql` monitor only allows to run query but does not report any metrics by default. Or they will be a full rewrite of the plugin MySQL without using `collectd` ?

It will allow you to arbitrary convert data from rows to datapoints, using the expression language that we currently use for discovery rules.  I will show an example of it with the MySQL stuff when I implement it.

xp-1000:
Hello @keitwb 

> getting a new collectd release is not trivial given all the patches we have on it and we are moving away from the collectd side of the agent anyway

I am little worried by this answer. what is the plan about the future of your agent so ? seems to be a huge change to drop collectd and not only for the collect of metrics. Indeed, lot of signalfx dashboards from integrations are based on collectd monitors which will implies new prebuilt dashboards.

> It will allow you to arbitrary convert data from rows to datapoints, using the expression language that we currently use for discovery rules. I will show an example of it with the MySQL stuff when I implement it.

I am also curious of this example. we could also create a telegraf monitor https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mysql but this will report lot of metrics and all of them as "custom" ..

keitwb:
> I am little worried by this answer. what is the plan about the future of your agent

We have been slowing adding functionality to the agent that replaces collectd plugins and generally make new built-in content that is separate for it.  We probably won't ever completely drop collectd from the agent but just don't have any plans to upgrade it, rather focusing efforts on making the non-collectd side of the agent better. 

Long-term the plan is to move to the [OpenTelemetry collector](https://opentelemetry.io/docs/collector/about/) as our standard agent but this is several months away at least as we need to get the monitors in this repo into the collector.  Most likely only the non-collectd stuff in the agent will be ported to the collector. See https://www.signalfx.com/blog/splunk-accelerates-opentelemetry-growth/ for the high-level picture.  We're going to keep supporting this Smart Agent for a while yet (prob at least a few years) but that is the general trajectory.

cvauvarin:
Hi @keitwb,

Any news on this ? Thanks.

keitwb:
Hey @cvauvarin, sorry it took me a bit but I got stuck making sure it stayed backwards compatible with the original `sql` monitor.  It is good now and I made a PR #1389 that should give you the ability to do this plus a whole lot more.  Let me know if that is acceptable for your use case.

xp-1000:
> Hey @cvauvarin, sorry it took me a bit but I got stuck making sure it stayed backwards compatible with the original `sql` monitor. It is good now and I made a PR #1389 that should give you the ability to do this plus a whole lot more. Let me know if that is acceptable for your use case.

Hello, just tested to retrieve replication status and log and it works like a charm:

```yaml
  - type: collectd/mysql
    host: &mysqlHost 127.0.0.1
    port: &mysqlPort 3306
    username: &mysqlUser root
    password: &mysqlPass master_root_password
    databases:
      - name: mysql
  - type: sql
    dbDriver: mysql
    # This is a YAML reference to avoid duplicating the above config.
    host: *mysqlHost
    port: *mysqlPort
    params:
      user: *mysqlUser
      password: *mysqlPass
    connectionString: '{{.user}}:{{.password}}@tcp({{.host}}:{{.port}})/mysql'
    queries:
      - query: 'SHOW SLAVE STATUS'
        datapointExpressions:
          - 'GAUGE("mysql_slave_running", {master_uuid: Master_UUID, channel: Channel_name}, Slave_SQL_Running == "Yes" ? 1 : 0)'
          - 'GAUGE("mysql_seconds_behind_master", {master_uuid: Master_UUID, channel: Channel_name}, Seconds_Behind_Master)'
```

cvauvarin:
Hey @keitwb sorry for the late reply but yes it seems that the new SQL feature is acceptable for us from what @xp-1000 has tested. We can close this PR so. Thanks !

PR #1364, opened 2020-06-18
ability to enable perCpu for load monitor in helm

xp-1000:
I understand we do not want to enable perCPU by default for retro compability purpose with collectd but, in my opinion, this metric is not usable withtout perCPU in a detector / alerting scope so we need, at least, an easy way override default value.

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/values.yaml
+++ b/deployments/k8s/helm/signalfx-agent/values.yaml
@@ -263,6 +263,11 @@ hostFSPath: /hostfs
 # monitors will be configured.
 configureStandardMonitors: true
 
+# If true, this will set the `perCPU` config option to `true` on `load`
+# standard monitor so the metric will be the average load per CPU.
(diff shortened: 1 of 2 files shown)


deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -263,6 +263,11 @@ hostFSPath: /hostfs
 # monitors will be configured.
 configureStandardMonitors: true
 
+# If true, this will set the `perCPU` config option to `true` on `load`
+# standard monitor so the metric will be the average load per CPU.
+# Has no effect if `configureStandardMonitors: false`
+configureLoadPerCPU: false
```
keitwb, line 269: Let's just call this `loadPerCPU` without the `configure` prefix.

xp-1000: done

PR #1365, opened 2020-06-18
Enable InnoDB stats on MySQL plugin

cvauvarin:
Add the possibility to use the `InnodbStats` option available in the [collectd plugin](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_mysql).

I decided to add [every available metrics](https://github.com/signalfx/collectd/blob/master/src/mysql.c#L805) gathered by the plugin when the stats are enabled. One metric is actually missing as I cannot find it in the [MySQL documentation](https://dev.mysql.com/doc/refman/5.6/en/server-status-variables.html) : `Innodb_dblwr_page_size`.


cvauvarin:
> Thanks for submitting this @cvauvarin. I'm assuming this works fine with the collectd version currently in the agent?

Yes it does unlike #1363. Tested it and it works.

PR #1376, opened 2020-06-29
Add ability to transform metric names per monitor

### LINKED ISSUE #1374: Ability to transform metric names from prometheus-exporter
It would be very useful for us to be able to specify a prefix to prepend to metrics ingested from a prometheus exporter.

For example, the exporter might present a metric like `request_count` but we'd like that to end up in signalfx as `service.subcomponent.request_count`

I don't *think* that this ability exists already, but would love to find that I just missed it. If not, is this something that would be considered for addition?

keitwb:
Fixes #1374.

I had had this work done a while back but it got put on the backburner and I forgot about it until now.


pkg/monitors/output.go
```diff
@@ -84,6 +86,20 @@ func (mo *monitorOutput) preprocessDP(dp *datapoint.Datapoint) bool {
...
+		re := reWithRepl.Regexp
+		repl := reWithRepl.Replacement
+
+		// An optimization for simple regexps (i.e. ones with no special
+		// matching syntax)
+		if prefix, complete := re.LiteralPrefix(); complete && prefix == dp.Metric {
+			dp.Metric = repl
+			continue
+		}
+		dp.Metric = re.ReplaceAllString(dp.Metric, repl)
```
rmfitzpatrick, line 100: What happens if this is the empty string?

keitwb: Then the metric name will be an empty string.

mstumpfx:
May want to add some sort of warning or note mentioning that changing default metric names could lead to increased custom metrics usage and/or break our pre-built content

PR #1378, opened 2020-06-30
Upgrade to cherry-picked collectd commit to get disk fix for newer ke…

keitwb:
…rnels


keitwb:
This is the latest from https://github.com/signalfx/collectd/tree/neoagent-5.8.0-sfx.

PR #1379, opened 2020-07-06
Change config file mode to 600 on Linux for Ansible, install script, and Salt

keitwb:


deployments/installer/install.sh
```diff
@@ -454,7 +458,7 @@ install() {
       ;;
   esac
 
-  configure_access_token "$access_token"
+  configure_access_token "$access_token" "$service_user" "$service_group"
```
jchengsfx, line 461: This should be moved to after `create_user_group` is called.

keitwb: Ok, I just moved all of these to after that.

PR #1381, opened 2020-07-07
Fix leak of tokens in processlist monitor on Windows

pjanotti:
The defer call of token.Close() can only be made after the value of the
token is set, otherwise, it "closes" an invalid handle.


keitwb:
@pjanotti  See https://dev.azure.com/signalfx/signalfx-agent/_build/results?buildId=3796&view=logs&j=18a12b53-e25b-5b79-411d-9a29f3a83d7c&t=becae859-8d01-508d-6d9d-600a82ae380c&l=14.  I would just ignore the error with `_ = func()` and it should make the linter happy.

pjanotti:
It is not straightforward to run the makefile on Windows since there is a bunch of bash scripts... perhaps I should try one of the *nix shells on Windows, for now relying on CI to run all validations.

PR #1389, opened 2020-07-10
sql monitor: Allow expressions to convert rows to metrics

keitwb:
This is a more powerful way of doing what #1363 is trying to achieve.


docs/monitors/sql.md
```diff
@@ -51,6 +51,60 @@ is the number of customers that belong to that combination of `country`
 and `status`.  You could also specify multiple `metrics` items to
 generate more than one metric from a single query.
 
+## Metric Expressions
+
+**Metric Expressions are a beta feature and may break in subsequent
+non-major releass.  The example documented will be maintained for backwards
```
mstumpfx, line 57:
```suggestion
non-major releases.  The example documented will be maintained for backwards
```

PR #1390, opened 2020-07-10
k8s: restrict configmap permissions to single namespace

mstumpfx:
Adds two new files configmap-role.yaml and configmap-rolebinding.yaml
and binds the role to the existing ServiceAccount. This allows
for the agent to only be permitted to view configmaps in the
namespace where the agent is running.


mstumpfx:
@keitwb It works as intended, the tests needed modified to use the new roles. I believe everything should be passing now.

PR #1397, opened 2020-07-24
Improvements to Correlation Client robustness

codesmith14:
This pr does several things around the correlation of services and environments to host id and tenant id dimensions.
* Retry failed requests on a delay configured by `propertiesSendDelaySeconds`
* Enforces a maximum number of retries per request configured by `traceHostCorrelationMaxRequestRetries`
* cancel conflicting pending delete and create requests
* deduplicate requests for creating and deleting correlations
* restructure correlation caches to be updated after an api request is successfully completed via callbacks
* Ensure that the configured access token is a valid http header at configuration time
* Enforces a maximum set size for host properties when the backend returns an http 418 status code with a maximum set size


pkg/core/writer/correlations/client.go
```diff
@@ -80,92 +108,163 @@ func NewCorrelationClient(ctx context.Context, conf *config.WriterConfig) (Corre
...
+	err := cc.putRequestOnChan(&request{
+		Correlation: cor,
+		operation:   http.MethodPut,
+		callback: func(body []byte, statuscode int, err error) {
+			switch statuscode {
+			case http.StatusOK:
+				if cc.logUpdates {
+					log.WithFields(log.Fields{"method": http.MethodPut, "correlation": cor}).Info("Updated dimension")
+				}
+			case http.StatusTeapot:
```
keitwb, line 186: Does our backend really return this or is this just some internal hack you are using?

codesmith14: It will in the not too distant future.  The scope of my bug creeped to include this now.

codesmith14:
@keitwb We do have some integration testing around this.  You can see them added with this commit: https://github.com/signalfx/signalfx-agent/commit/3d1c91ba28c7921fcec1dc264b39464bb69f5d37#diff-520820ace78ed9eefdc02d26fe2e7e43

codesmith14:
@keitwb I've added a performance test to put pressure on the correlation client with request retries and I hooked up the correlation client's internal metrics to the writer's internal metrics call back.  I also changed the data structures inside of the correlation client's deduper and added a purge interval to make sure it doesn't hold on to things it shouldn't.

keitwb:
@codesmith14 the perf test failures are very suspicious, can you see if they are related to your changes?

PR #1398, opened 2020-07-24
Add Windows Docker Support

keitwb:
This is just a simple Go compilation of the agent for Windows,
without any Python or Java subsystems.

In order to make this work, there need to be a special value `bypass` to
the `-service` cli flag that causes the Windows service framework to be
skipped and the agent executed directly.

The Helm chart is also updated to deploy the Windows agent container
more seamlessly.

### DIFF
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -1,0 +1,24 @@
+FROM golang:1.14-windowsservercore-1809 as builder
+
+WORKDIR C:\\SignalFxAgent
+
+COPY go.* ./
+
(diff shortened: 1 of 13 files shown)


rpille-sfx:
No customer facing doc changes needed for this, @keitwb ?

keitwb:
> No customer facing doc changes needed for this, @keitwb ?

I would consider everything in this repo "customer-facing", but I have made the updates that I think are appropriate for such an early stage of the Windows Docker Kubernetes support.

PR #1405, opened 2020-07-30
Add disk_ops.pending metric to diskio monitor

mstumpfx:
The `disk_ops.pending` metric is for parity with the collectd/disk metric `pending_operations`


pkg/monitors/diskio/diskio_others.go
```diff
@@ -36,6 +36,7 @@ func (m *Monitor) makeLinuxDatapoints(disk disk.IOCountersStat, dimensions map[s
 		datapoint.New("disk_merged.write", dimensions, datapoint.NewIntValue(int64(disk.MergedWriteCount)), datapoint.Counter, time.Time{}),
 		datapoint.New("disk_time.read", dimensions, datapoint.NewIntValue(int64(disk.ReadTime)), datapoint.Counter, time.Time{}),
 		datapoint.New("disk_time.write", dimensions, datapoint.NewIntValue(int64(disk.WriteTime)), datapoint.Counter, time.Time{}),
+		datapoint.New("pending_operations", dimensions, datapoint.NewIntValue(int64(disk.IopsInProgress)), datapoint.Gauge, time.Time{}),
```
keitwb, line 39: I would call this `disk_ops.pending` for consistency.

keitwb: Also for the new metric, reference the constant from genmetadata.go.  You don't need to change the existing metrics.

mstumpfx: Sounds good, was going for parity with the old collectd metric name but I like this format better.

PR #1406, opened 2020-07-30
Add telegraf/ntpq monitor

keitwb:
This has information that is very difficult to calculate in the
existing 'ntp' monitor, so the telegraf plugin was brought in to
provide it.


keitwb:
> Metric descriptions would be nice but LGTM

I have no idea what they mean and it's very hard to find clear docs for it.

PR #1408, opened 2020-08-06
Add comment to readme file

tmorrisonsfx:
Added comment that these README files should not be updated. All content should be updated in the specified file. This is related to https://github.com/signalfx/integrations/pull/866 and https://signalfuse.atlassian.net/browse/DOCS-1388.

### DIFF
--- a/deployments/ecs/README.md
+++ b/deployments/ecs/README.md
@@ -1,3 +1,5 @@
+<!--- This content should be updated in the aws-ecs/README.md.jinja file. DO NOT MODIFY HERE. --->
+
 # ECS Deployment
 
 ## Create Task Definition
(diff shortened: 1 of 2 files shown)


deployments/ecs/README.md
```diff
@@ -1,3 +1,5 @@
+<!--- This content should be updated in the aws-ecs/README.md.jinja file. DO NOT MODIFY HERE. --->
```
keitwb, line 1: Is there going to be some kind of automatic sync process?  I'm not aware of any.  I don't think this comment is accurate.

tmorrisonsfx: We request that any updates from Engineering be made in the Jinja file and not in this file. There will be no automatic sync process. I moved the information into the tile so that customers don’t have to go to GitHub for the setup information.

PR #1413, opened 2020-08-12
Add new 'mode' dimension to filesystems monitor

mstumpfx:
- Enable by setting the new config option `sendModeDimension` to true
- Added as a config option to not churn MTSs for existing users upon upgrade

### DIFF
--- a/pkg/monitors/filesystems/filesystems.go
+++ b/pkg/monitors/filesystems/filesystems.go
@@ -46,17 +46,22 @@ type Config struct {
 	// `/hostfs/` mount in the filter.  If both this and `fsTypes` is
 	// specified, the two filters combine in an AND relationship.
 	MountPoints []string `yaml:"mountPoints"`
+
+	// Set to true to emit the "mode" dimension, which represents
+	// whether the mount is "rw" or "ro".
+	SendModeDimension bool `yaml:"sendModeDimension" default:"false"`
 }
 
 // Monitor for Utilization
 type Monitor struct {
-	Output      types.FilteringOutput
-	cancel      func()
-	conf        *Config
-	hostFSPath  string
-	fsTypes     *filter.OverridableStringFilter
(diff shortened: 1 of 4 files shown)


pkg/monitors/filesystems/filesystems.go
```diff
@@ -66,6 +71,17 @@ func (m *Monitor) getCommonDimensions(partition *gopsutil.PartitionStat) map[str
 		"device":     strings.Replace(partition.Device, " ", "_", -1),
 		"fs_type":    strings.Replace(partition.Fstype, " ", "_", -1),
 	}
+	if m.sendModeDimension {
+		var mode string
+		if partition.Opts != "" {
+			mode = partition.Opts[:2]
```
keitwb, line 77: This seems really sketchy, are you sure it's always the first two chars of the options?

mstumpfx: I agree it's sketchy, but yeah after diving into how the field is set, it is always the first two characters.
I don't see another way to get the information other than this string.

Examples:
https://github.com/shirou/gopsutil/blob/master/disk/disk_darwin.go#L29-L32
https://github.com/shirou/gopsutil/blob/master/disk/disk_linux.go#L272-L284
https://github.com/shirou/gopsutil/blob/master/disk/disk_windows.go#L121-L127

keitwb: Given that options are always comma separated (I'm at least more confident of that than the fact that the mode option always comes first), just do a strings.Split(opts, ",") and see if one of the items is `ro` or `rw`.  That seems at least a bit more robust.

mstumpfx: That was the other way I was considering. Sounds good to me.

keitwb:
Don't worry abotu perf_test, I'm pretty confident it isn't related to your changes.

PR #1414, opened 2020-08-13
Revert "Add comment to readme file"

keitwb:
Reverts signalfx/signalfx-agent#1408.

@tmorrisonsfx This isn't going to work.  If we have the files in the agent repo, then they have to be maintainable. If we updated a file in the integrations repo and not here it would just be out of date here which would be very bad.

### DIFF
--- a/deployments/ecs/README.md
+++ b/deployments/ecs/README.md
@@ -1,5 +1,3 @@
-<!--- This content should be updated in the aws-ecs/README.md.jinja file. DO NOT MODIFY HERE. --->
-
 # ECS Deployment
 
 ## Create Task Definition
--- a/deployments/fargate/README.md
+++ b/deployments/fargate/README.md
@@ -1,5 +1,3 @@
-<!--- This content should be updated in the aws-ecs/README.md.jinja file. DO NOT MODIFY HERE. --->
-
 # Fargate Deployment
 
 ## Create Task Definition


tmorrisonsfx:
> Reverts #1408.
> 
> @tmorrisonsfx This isn't going to work. If we have the files in the agent repo, then they have to be maintainable. If we updated a file in the integrations repo and not here it would just be out of date here which would be very bad.

@keitwb OK, I was following the instructions in the Jira issue. Do I need to re-request membership for the integrations team on the SignalFx organization since I was removed? @rpille-sfx

keitwb:
> Do I need to re-request membership for the integrations team on the SignalFx organization since I was removed?

No, there is a team for docs that you are a part of that lets you make PRs to repos.

rpille-sfx:
Innocent question: Why do these files need to be in the agent repo? The goal is for us to be able to maintain content in a more streamlined fashion in the jinja files for each integration. Can't we remove them here and update the content in that new file?

keitwb:
Because these docs are inextricably linked to the agent itself and if the agent changes in a way that affects these docs, it is far more likely that they will get updated if they are in the repo where the changes are happening.  Also this is open source software so people are going to be comfortable interacting with the repo directly.  I would just direct people here from the other docs since you get all of the config files split out into separate files instead of inlining everything into a doc file.  Also the easiest way for developers to signal to the docs team that there are changes to things is to update docs like this and the docs team notices the changes and then does updates at the other places necessary, it is very difficult for us to keep track of where docs are once they escape from the code repos.

rpille-sfx:
Thanks for this extra detail @keitwb. We'll give some more thought to how to accommodate these needs in the new processes and repos being designed for the O11y suite.

PR #1422, opened 2020-08-18
Re-run failed tests if xdist worker crashed

jchengsfx:
xdist worker occasionally crashes due to bug in xdist or circleci load.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,6 +98,9 @@ commands:
       markers:
         type: string
         default: ""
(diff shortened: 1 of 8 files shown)


jchengsfx:
> See comments.

Re-worked the changes to separate xdist option from `PYTEST_OPTIONS` which should simplify first-run and re-run commands.

Also, ansible test failures fixed in https://github.com/signalfx/signalfx-agent/pull/1419.

PR #1425, opened 2020-08-20
add kubernetes-volumes monitor to helm chart

xp-1000:
Hello,

https://docs.signalfx.com/en/latest/integrations/agent/monitors/kubernetes-volumes.html monitor seems very valuable in my opinion and we need to be able to configure it from the helm chart.

Of course, the `monitors` variable already allows us to configure it but this is not enough while we need additional permissions in cluster role to allow the agent to gather these metrics.

The main goal of this PR is to make it possible to configure these permissions but for easier configuration this will also add this monitor to standard monitors configuration for kubernetes.

Still in work in progress while I did not tested enough (only GKE for now) and I could need to add changes but I will be happy to get your impression about that from now (may be I missed something which make this PR useless).


keitwb:
Well, the changes look good, but you do need to sync all of the k8s resource stuff.  The failing CircleCI job says what to do.

xp-1000:
> Well, the changes look good, but you do need to sync all of the k8s resource stuff. The failing CircleCI job says what to do.

I ran the script `./deployments/k8s/generate-from-helm` and everything seems good but it seems to have "downgraded" the agent version is it normal ?

xp-1000:
> @keitwb
> 
> > Well, the changes look good, but you do need to sync all of the k8s resource stuff. The failing CircleCI job says what to do.
> 
> I ran the script `./deployments/k8s/generate-from-helm` and everything seems good but it seems to have "downgraded" the agent version is it normal ?

ok, I confess I don't know why `./deployments/k8s/generate-from-helm` want to downgrade the agent version whereas my fork is synced to upstream master but I dismiss changes on `deployments/k8s/daemonset.yaml` and `deployments/k8s/serverless/deployment.yaml` and should be good

PR #1430, opened 2020-08-21
add some postgresql metrics

xp-1000:
Hello,

I create this PR as a draft hoping you could help me.

I would like to add some necessary metrics for our usage on `postgresql` monitor.

For now, I use this monitor with `sql` to add these queries but I am pretty sure it could be useful for others and will ease our configuration.

```
  - type: postgresql
    host: &psqlHost 127.0.0.1
    port: &psqlPort 5432
    params: &psqlParams
      username: {"#from": "vault:secret/my-database[username]"}
      password: {"#from": "vault:secret/my-database[password]"}
    masterDBName: &psqlDBName postgres
    # Do not add host, port and dbname
    connectionString: "user={{.username}} password={{.password}} sslmode=disable"
    # Do not change this, needed for detectors
    extraDimensions: &psqlDimensions
      postgres_host: *psqlHost
      postgres_port: *psqlPort
  # You should not have to change this while it uses postgresql monitor config
  - type: sql
    dbDriver: postgres
    extraDimensions: *psqlDimensions
    params:
      << : [ {host: *psqlHost}, {port: *psqlPort}, {dbname: *psqlDBName}, *psqlParams ]
    connectionString: "host={{.host}} port={{.port}} dbname={{.dbname}} user={{.username}} password={{.password}} sslmode=disable"
    queries:
      - query: >-
          WITH max_con AS (SELECT setting::float FROM pg_settings WHERE name = 'max_connections')
          SELECT COUNT(*)/MAX(setting) AS pct_connections FROM pg_stat_activity, max_con;
        metrics:
          - metricName: "postgres_pct_connections"
            valueColumn: "pct_connections"
      - query: >-
          SELECT GREATEST (0, (EXTRACT (EPOCH FROM now() - pg_last_xact_replay_timestamp()))) AS lag,
          CASE WHEN pg_is_in_recovery() THEN 'standby' ELSE 'master' END AS replication_role;
        metrics:
          - metricName: "postgres_replication_lag"
            valueColumn: "lag"
            dimensionColumns: ["replication_role"]
      - query: "SELECT slot_name, slot_type, database, case when active then 1 else 0 end AS active FROM pg_replication_slots;"
        metrics:
          - metricName: "postgres_replication_state"
            valueColumn: "active"
            dimensionColumns: ["slot_name", "slot_type", "database"]
      - query: "SELECT COUNT(*) AS locks FROM pg_locks WHERE NOT granted;"
        metrics:
          - metricName: "postgres_locks"
            valueColumn: "locks"
      - query: "SELECT datname AS database, xact_commit, xact_rollback, conflicts FROM pg_stat_database;"
        metrics:
          - metricName: "postgres_conflicts"
            valueColumn: "conflicts"
            dimensionColumns: ["database"]
            isCumulative: true
          - metricName: "postgres_xact_commits"
            valueColumn: "xact_commit"
            dimensionColumns: ["database"]
            isCumulative: true
          - metricName: "postgres_xact_rollbacks"
            valueColumn: "xact_rollback"
            dimensionColumns: ["database"]
            isCumulative: true
```

I use this configuration and every queries work fine, all metrics are collected properly this way.

Sadly, with my current changes on `postgresql` some metrics are not collected anymore:

```
  - type: postgresql
    host: &psqlHost 127.0.0.1
    port: &psqlPort 5432
    params: &psqlParams
      username:  {"#from": "vault:secret/my-database[username]"}
      password:  {"#from": "vault:secret/my-database[password]"}
    masterDBName: &psqlDBName postgres
    # Do not add host, port and dbname
    connectionString: "user={{.username}} password={{.password}} sslmode=disable"
    # Do not change this, needed for detectors
    extraDimensions: &psqlDimensions
      postgres_host: *psqlHost
      postgres_port: *psqlPort
    extraMetrics:
      - "postgres_conflicts"
      - "postgres_locks"
      - "postgres_replication_lag"
```

these metrics work:
- `postgres_deadlocks`
- `postgres_xact_rollbacks`
- `postgres_xact_commits`

but these metrics do not work;
- `postgres_pct_connections`
- `postgres_locks`

I run my 2 configurations (sql & postgres or new postgres only) on the same postgresql server and I don't see any log error about not collected metrics (or underlying queries).

do you have an idea ? may be @keitwb

### DIFF
--- a/pkg/monitors/postgresql/monitor.go
+++ b/pkg/monitors/postgresql/monitor.go
@@ -59,13 +59,11 @@ type Config struct {
 	TopQueryLimit int `default:"10" yaml:"topQueryLimit"`
 }
 
-func (c *Config) connStr() (template string, host string, port string, err error) {
+func (c *Config) connStr() (template string, port string, err error) {
 	connStr := c.ConnectionString
-	host = "localhost"
 	port = "5432"
 	if c.Host != "" {
 		connStr += " host=" + c.Host
-		host = c.Host
 	}
 	if c.Port != 0 {
 		connStr += fmt.Sprintf(" port=%d", c.Port)
@@ -98,11 +96,10 @@ func (m *Monitor) Configure(conf *Config) error {
 
 	queriesGroupEnabled := m.Output.HasEnabledMetricInGroup(groupQueries)
 
-	connStr, host, port, err := conf.connStr()
+	connStr, port, err := conf.connStr()
 	if err != nil {
 		return fmt.Errorf("could not render connectionString template: %v", err)
 	}
-	m.Output.AddExtraDimension("postgres_host", host)
 	m.Output.AddExtraDimension("postgres_port", port)
 
 	m.database, err = dbsql.Open("postgres", connStr+" dbname="+m.conf.MasterDBName)
@@ -196,7 +193,7 @@ func (m *Monitor) Configure(conf *Config) error {
 }
 
 func (m *Monitor) startMonitoringDatabase(name string) (*sql.Monitor, error) {
-	connStr, _, _, err := m.conf.connStr()
+	connStr, _, err := m.conf.connStr()
 	if err != nil {
 		return nil, err
 	}
@@ -246,7 +243,7 @@ func (m *Monitor) determineDatabases() ([]string, error) {
 func (m *Monitor) monitorServer() (*sql.Monitor, error) {
 	sqlMon := &sql.Monitor{Output: m.Output.Copy()}
 
-	connStr, _, _, err := m.conf.connStr()
+	connStr, _, err := m.conf.connStr()
 	if err != nil {
 		return nil, err
 	}
@@ -263,7 +260,7 @@ func (m *Monitor) monitorServer() (*sql.Monitor, error) {
 func (m *Monitor) monitorStatements() (*sql.Monitor, error) {
 	sqlMon := &sql.Monitor{Output: m.Output.Copy()}
 
-	connStr, _, _, err := m.conf.connStr()
+	connStr, _, err := m.conf.connStr()
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/monitors/postgresql/monitor.go
+++ b/pkg/monitors/postgresql/monitor.go
@@ -113,7 +113,7 @@ func (m *Monitor) Configure(conf *Config) error {
 	}
 
 	databaseDatapointFilter, err := dpfilters.NewOverridable(nil, map[string][]string{
-		"database": conf.Databases,
+		"database?": conf.Databases,
 	})
 	if err != nil {
 		m.database.Close()
@@ -249,6 +249,7 @@ func (m *Monitor) monitorServer() (*sql.Monitor, error) {
 		ConnectionString: connStr + " dbname=" + m.conf.MasterDBName,
 		DBDriver:         "postgres",
 		Queries:          defaultServerQueries,
+		LogQueries:       m.conf.LogQueries,
 	})
 }
 
@@ -265,6 +266,7 @@ func (m *Monitor) monitorStatements() (*sql.Monitor, error) {
 		ConnectionString: connStr + " dbname=" + m.conf.MasterDBName,
 		DBDriver:         "postgres",
 		Queries:          makeDefaultStatementsQueries(m.conf.TopQueryLimit),
+		LogQueries:       m.conf.LogQueries,
 	})
 }
 
--- a/pkg/monitors/postgresql/monitor.go
+++ b/pkg/monitors/postgresql/monitor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	dbsql "database/sql"
 	"fmt"
+	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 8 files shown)


keitwb:
@xp-1000 Yeah the issue is with the missing `database` dimension causing the metrics to get filtered out.  I added an enhancement to the map filtering logic to allow for missing keys with a special suffix of `?` when creating the filter, which the postgresql monitor now uses.  I could get both of those metrics after that.

Also I would make the new metrics that you added non-default, since our criterion for default vs non-default is whether they appear in our built-in content.

xp-1000:
> @xp-1000 Yeah the issue is with the missing `database` dimension causing the metrics to get filtered out. I added an enhancement to the map filtering logic to allow for missing keys with a special suffix of `?` when creating the filter, which the postgresql monitor now uses. I could get both of those metrics after that.
> 
> Also I would make the new metrics that you added non-default, since our criterion for default vs non-default is whether they appear in our built-in content.

Thank you very much for your help.

I did not know for this criteria it makes sens. is this possible for us to propose changes on built-in content ? may be submitting a PR here https://github.com/signalfx/integrations ?

the PR should be fine now but please notice I add 2 dimensions to every metrics `postgres_host` and `postgres_dimension` so tell me if you think it could be a problem.

The goal is to guarantee unique time series from different servers. Indeed for now there is only `host` dimension (except if disabled at agent level like for "serverless" deployment mode on helm chart) and this is not always enough. For example, suppose we monitor multiple postgresql servers from the same host (dedicated ec2 for signalfx pulling from multiple RDS endpoints).

pkg/monitors/postgresql/monitor.go
```diff
@@ -93,10 +98,12 @@ func (m *Monitor) Configure(conf *Config) error {
 
 	queriesGroupEnabled := m.Output.HasEnabledMetricInGroup(groupQueries)
 
-	connStr, err := conf.connStr()
+	connStr, host, port, err := conf.connStr()
 	if err != nil {
 		return fmt.Errorf("could not render connectionString template: %v", err)
 	}
+	m.Output.AddExtraDimension("postgres_host", host)
```
keitwb, line 105: Yeah I would get rid of these (host/port).  If you are using the docker or k8s observer, the observer will put dimensions on the output that will identify it based on certain attributes of the postgresql container.  If you are using a plain host without containers, the `host` dimension already added to datapoints at a higher level can be used to differentiate postgresql servers.  If you are targeting a postgresql server from an agent on a separate host, this could be helpful in combination with the `disableHostDimensions: true` config option on the monitor, but the standard deployment model is to run the agent on the same host.  So basically these dimensions will be redundant or useless in a lot of scenarios and shouldn't be added universally.

keitwb:
> For example, suppose we monitor multiple postgresql servers from the same host (dedicated ec2 for signalfx pulling from multiple RDS endpoints).

If you do that you need to use the `disableHostDimensions: true` config option on the monitor and use the extraDimensions config like you originally had.  The standard deployment model of the agent is same host as the service, so all the defaults cater to that for consistency.

pkg/monitors/postgresql/monitor.go
xp-1000: I understand and this is true that we do not need in most cases.

However this will allow to support every cases natively. Adding extra dimensions is not a problem in my opinion but this need is not obvious for everybody so it will avoid mishap for some users and ease the config for everyone.

I would say, this change should be fully retro compatible as it does not delete any dimension or change existing ones (so detectors using explicit aggregation will not be impacted) and just add more granularity (at worst redundant) so this will also work when aggregation is not used.

Also, some existing monitors work on this model (i.e. collectd/redis, elasticsearch with `plugin_instance` dimension). Some also provide an explicit option to set this field like apache or nginx with `name` which is more natural than extraDimension common monitor parameter.

I understand for most of them this is a collectd legacy but this make it straightforward for them to be used and this is sad to "lose" this feature in new "signalfx" monitors versions whereas I very like the default behavior of signalfx to apply signalflow to every single timeseries found contrary to datadog where aggregation is mandatory.

In my opinion, even if redondant in general it only brings good things so I would prefer to keep it if you do not have strong opinion on this. Else, tell me again and I will remove this change.

keitwb: I think the best thing is if you just always made the `host` dimension refer to the host of the postgresql server and refer to that in your charts/detectors.  If you are using the standard deployment model where the agent runs on the same host as Postgres, then it will already be the case.  If you are monitoring a remote server like RDS, then the best thing to do is to set `disableHostDimensions: true` on the monitor and then set `extraDimensions: {host: *host}` in your config.  Otherwise you'll get `host` dimension set to the host of the agent, which is probably irrelevant to monitoring purposes.  I admit it is not the most natural thing to have to configure it that way but it is because it is going against the standard deployment model of the agent, which always gets favored for consistency.

The other benefit of identifying postgres servers by `host` is that you can correlate other metrics from other monitors by that same dimension.  Nothing else is going to have `postgres_host` on it.

xp-1000: ok it is removed!

PR #1431, opened 2020-08-25
K8S deployment add new chart values parameter to split monitors configuration in multiple files

jphilaine:
Monitor parameters are automatically loaded from /etc/signalfx/monitors.d repository in container.

It's more easy to managed multiple small files than a one longer file.

To use this feature, you need add all files in helm arguments.

**Example**
Build helm parameters command:
```
files_args=""
for config_yaml in $(ls config/*.yaml)
do
    files_args="$files_args -f $config_yaml"
done
echo $files_args
```
and use `$files_args` in your helm command : 
```
helm_args="${release} ${chart_name} ${files_args} --namespace=${namespace}"
helm upgrade --install ${helm_args}
```

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/templates/configmap-monitorsd.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/configmap-monitorsd.yaml
@@ -1,0 +1,16 @@
+{{- if (.Values.monitorsd) }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "configmap-name" . }}-monitorsd
+  namespace: {{ template "signalfx-agent.namespace" . }}
+  labels:
+    app: {{ template "signalfx-agent.name" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    chart: {{ template "signalfx-agent.chart" . }}
+data:
+{{- if .Values.monitorsd }}
+{{ tpl (toYaml .Values.monitorsd) . | indent 2 }}
+{{- end }}
+{{- end -}}
--- a/deployments/k8s/helm/signalfx-agent/templates/configmap-monitorsd.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/configmap-monitorsd.yaml
@@ -10,7 +10,5 @@ metadata:
     release: {{ .Release.Name }}
     chart: {{ template "signalfx-agent.chart" . }}
 data:
-{{- if .Values.monitorsd }}
 {{ tpl (toYaml .Values.monitorsd) . | indent 2 }}
-{{- end }}
 {{- end -}}
--- a/deployments/k8s/helm/signalfx-agent/values.yaml
+++ b/deployments/k8s/helm/signalfx-agent/values.yaml
@@ -285,3 +285,12 @@ monitors:
 # included into a configmap. These files are automatically included in the previous
 # monitors section if it used.
 monitorsd:
+# All file names in the configmap will create a file in /etc/signalfx/monitors.d repository.
+# They must ended by '.yaml' to be valid.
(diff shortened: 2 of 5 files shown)


keitwb:
Is the bash script with the loop the only way to practically construct the args?  I'm a bit wary of having to do shell scripting stuff to use it, seems like Helm should be able to handle it natively.

jphilaine:
No, I just suggest a use case with a bash script. But you can defined all monitors definition in the same value file.

The goal of the request is to add the possibility to use multiple files for the complexe monitors configuration : in production is not easy to maintain a file with hundred of lines.

In practice, you need add all files one by one to helm install/upgrade command. To my knowledge, it's not possible to add path of files. But this part is an helm issue.

xp-1000:
> Is the bash script with the loop the only way to practically construct the args? I'm a bit wary of having to do shell scripting stuff to use it, seems like Helm should be able to handle it natively.

Hello @keitwb, in my knowledge there is no native way in helm to merge multiple files into only one (except by using multiple times the `-f` option which is the example provided by @jphilaine )

However, this is only an implementation suggestion in simple bash but everybody are free to use other implementation.

The goal of this PR is not to force them to use the base but only to allow them to separate configuration in different files for better readability and organization.

What is your feeling about that ? is there any blocker input to merge this ?

thanks

deployments/k8s/helm/signalfx-agent/templates/configmap-monitorsd.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: {{ include "configmap-name" . }}-monitorsd
+  namespace: {{ template "signalfx-agent.namespace" . }}
+  labels:
+    app: {{ template "signalfx-agent.name" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    chart: {{ template "signalfx-agent.chart" . }}
+data:
+{{- if .Values.monitorsd }}
```
keitwb, line 13: The whole file is wrapped in `if` for this value so this seems redundant.

deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -280,3 +280,8 @@ loadPerCPU: false
 # A list of monitor configurations to include in the agent config.  These
 # values correspond exactly to what goes under 'monitors' in the agent config.
 monitors:
+
+# For a complex monitor configurations, split into multiple files that will be
+# included into a configmap. These files are automatically included in the previous
+# monitors section if it used.
+monitorsd:
```
keitwb, line 287: Is this a boolean value or what?

keitwb:
@jphilaine what would the other .yaml files in the config/ dir look like?  Are they raw monitor config or would the monitor config be nested in something?

deployments/k8s/helm/signalfx-agent/templates/configmap-monitorsd.yaml
jphilaine: I removed this redundant condition.

deployments/k8s/helm/signalfx-agent/values.yaml
jphilaine: I added an example of usage of this parameter.

jphilaine:
> @jphilaine what would the other .yaml files in the config/ dir look like? Are they raw monitor config or would the monitor config be nested in something?

YAML files are directly included in monitors agent configuration part:

```
- {"#from": "/etc/signalfx/monitors.d/*.yaml", flatten: true, optional: true}
```

xp-1000:
Hello @keitwb 
what it miss to this PR to merge it ?
thanks

keitwb (approved):

PR #1436, opened 2020-08-26
Fix docker image for aarch64 images

squishykid:
While building the container, libs and binaries are patched so ld uses relative paths when loading binaries. It seems this is done because the final container is an amalgamation of different binaries with loading different shared libs.

The first change is to use `realpath` in a directory where it is not patched.

The second change is to skip patching `libc` and `libpthread` as these appear to get mangled by `patchelf`. They have no dependencies to be loaded, so leaving them unpatched is fine.

### DIFF
--- a/scripts/patch-rpath
+++ b/scripts/patch-rpath
@@ -11,7 +11,7 @@ echo "Using final root directory of $root"
 
 cd $root
 
-for f in $(find bin lib jre -type f -executable -or -name "*.so*"); do
+for f in $(find bin lib jre -type f -executable -or -name "*.so*" | grep -v libc.so.6 | grep -v libpthread.so.0); do
   existing_rpath=$(patchelf --print-rpath $f 2>/dev/null || true)
-  patchelf --set-rpath '$ORIGIN/'$($root/bin/realpath --relative-to $(dirname $f) ${root%/}/lib):$existing_rpath $f 2>/dev/null || true
+  patchelf --set-rpath '$ORIGIN/'$(/usr/bin/realpath --relative-to $(dirname $f) ${root%/}/lib):$existing_rpath $f 2>/dev/null || true
 done


scripts/patch-rpath
```diff
@@ -11,7 +11,7 @@ echo "Using final root directory of $root"
 
 cd $root
 
-for f in $(find bin lib jre -type f -executable -or -name "*.so*"); do
+for f in $(find bin lib jre -type f -executable -or -name "*.so*" | grep -v libc.so.6 | grep -v libpthread.so.0); do
```
MovieStoreGuy, line 14: Can I ask why we need to ignore these to shared objects? 
Furthermore, can they be turned into one grep statement that has a comment or such to help future developers understand why we care about these?

squishykid: These are skipped because `patchelf` mangles them. Luckily they do not have dependencies that need to be loaded, so they don't actually need to get patched.

scripts/patch-rpath
```diff
@@ -11,7 +11,7 @@ echo "Using final root directory of $root"
...
   existing_rpath=$(patchelf --print-rpath $f 2>/dev/null || true)
-  patchelf --set-rpath '$ORIGIN/'$($root/bin/realpath --relative-to $(dirname $f) ${root%/}/lib):$existing_rpath $f 2>/dev/null || true
+  patchelf --set-rpath '$ORIGIN/'$(/usr/bin/realpath --relative-to $(dirname $f) ${root%/}/lib):$existing_rpath $f 2>/dev/null || true
```
keitwb, line 16: The docker image filesystem also serves as the standalone tar bundle that we distribute in a non-containerized form.  That is why we bundle realpath and refer to the bundled version here so that we aren't assuming it exists on the target system (it quite often doesn't).

What problem was the bundled realpath causing on ARM?

keitwb:
Thanks for figuring that out.  The docker image filesystem also serves as the contents of our standalone tar.gz bundle that we distribute for non-containerized deployments, so I'm going to test this on RHEL 6 to be sure nothing is affected.  If all looks good we can merge this.

PR #1438, opened 2020-08-27
Cleanup patch for aarch64 support

squishykid:
PR #1436 was a bit of a hack- this achieves the same results without ignoring a magic list of shared objects.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,9 +127,9 @@ RUN wget -O /tmp/Python-${PYTHON_VERSION}.tgz https://www.python.org/ftp/python/
 
 # Compile patchelf statically from source
 RUN cd /tmp &&\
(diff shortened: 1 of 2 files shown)


MovieStoreGuy (approved):
This is a really neat way to ship support for multiple archs

keitwb:
Ok so basically patchelf 0.11 fixed the corruption issues in the prior version?

squishykid:
Yep, that's it.

squishykid:
I think it might have been fixed in this patch https://github.com/NixOS/patchelf/pull/171

MovieStoreGuy:
@keitwb is this okay to merge? I see the old PR as a code smell and this helps address the actual issue rolling forward without some tribal knowledge hack.

PR #1440, opened 2020-08-27
Build windows docker image

jchengsfx:
- Pushes image to quay.io if the branch is a release tag.

### DIFF
--- a/.azure-pipelines/config.yml
+++ b/.azure-pipelines/config.yml
@@ -145,6 +145,33 @@ jobs:
       pathtoPublish: '$(Build.ArtifactStagingDirectory)'
       artifactName: 'bundle'
 
+- job: 'win_docker_build'
+  pool:
+    vmImage: 'windows-2019'
+  steps:
+  - bash: |
+      set -euxo pipefail
+
+      IS_RELEASE_TAG="false"
+      AGENT_VERSION="latest"
+
+      if [[ "$BUILD_SOURCEBRANCH" =~ ^refs/tags/v\d+\.\d+\.\d+$ ]]; then
+          IS_RELEASE_TAG="true"
+          AGENT_VERSION="${BUILD_SOURCEBRANCHNAME#v}"
+      fi
+
+      docker build -t quay.io/signalfx/signalfx-agent:${AGENT_VERSION}-windows --build-arg AGENT_VERSION=${AGENT_VERSION} -f Dockerfile.windows .
+
+      # push release image
+      if [[ "$IS_RELEASE_TAG" = "true" ]]; then
+          docker login -u="$QUAY_USERNAME" -p="$QUAY_TOKEN" quay.io
+          docker push quay.io/signalfx/signalfx-agent:${AGENT_VERSION}-windows
+      fi
+    displayName: 'Build windows docker image'
+    env:
+      QUAY_USERNAME: $(quayUsername)
+      QUAY_TOKEN: $(quayToken)
+
 - job: 'win_2019_integration_tests'
   pool:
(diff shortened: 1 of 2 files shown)


.azure-pipelines/config.yml
```diff
@@ -145,6 +145,35 @@ jobs:
...
+
+      docker build -t quay.io/signalfx/signalfx-agent:${AGENT_VERSION}-windows --build-arg AGENT_VERSION=${AGENT_VERSION} -f Dockerfile.windows .
+
+      # push release image
+      if [[ "$AGENT_VERSION" != "latest" ]]; then
+          docker login -u="$QUAY_USERNAME" -p="$QUAY_TOKEN" quay.io
+          docker push quay.io/signalfx/signalfx-agent:${AGENT_VERSION}-windows
+      fi
+    displayName: 'Build windows docker image'
+    condition: and(ne(variables.hasChanges, 'false'), ne(variables.hasChanges, False))
```
keitwb, line 172: Just make it bool always and simplify this.

.azure-pipelines/config.yml
```diff
@@ -145,6 +145,35 @@ jobs:
...
+  pool:
+    vmImage: 'windows-2019'
+  steps:
+  - template: 'templates/changes-include.yml'
+    parameters:
+      paths: 'go.mod go.sum Dockerfile.windows .azure-pipelines'
+  - bash: |
+      set -euxo pipefail
+
+      if [[ "$BUILD_SOURCEBRANCH" =~ ^refs/tags/v\d+\.\d+\.\d+$ ]]; then
```
keitwb, line 158: If we are pushing final releases, this job should be limited to only tag pushes.

RELEASE.md
```diff
@@ -163,6 +163,11 @@ point but for now the process is manual.
    Server 2016 vagrant box. See the "Windows" section in
    [development.md](docs/development.md) for details.
 
+1. **Note**: The Windows docker image is automatically built and pushed to quay.io
+   in Azure Pipelines for release tags.  Check [Azure Pipelines](https://dev.azure.com/signalfx/signalfx-agent/_build?definitionId=1)
+   and [quay.io](https://quay.io/repository/signalfx/signalfx-agent?tab=tags) to
+   ensure that the image was built and pushed successfully for the release tag.
```
keitwb, line 169: The config above looks like it only gets pushed if there are changes to certain files, not release tags.

.azure-pipelines/config.yml
jchengsfx: The build has been taking over 20 minutes, so just wanted to limit unnecessary builds. But I can set to always build if desired.

.azure-pipelines/config.yml
jchengsfx: So don't build for PRs if there are changes to Dockerfile.windows and other related files?

.azure-pipelines/config.yml
keitwb: No I mean make it always `False` instead  of potentially `'false'` (string).  This condition is kind of irrelevant though if you want to build every time a release tag is pushed though.

.azure-pipelines/config.yml
keitwb: We should only push to the `signalfx/signalfx-agent` quay repo for production releases, and it has to build it every time because the Windows docker image includes the agent code, not just stuff in DOckerfile.windows/go.mod, etc.

RELEASE.md
jchengsfx: Simplified config to always build and only push for release tags.

keitwb (approved):

PR #1442, opened 2020-08-27
postgresql group for replication and auto disable for aurora

xp-1000:
Hello,

AWS Postgresql Aurora does not support `pg_last_xlog_receive_location()` and replication metrics (state or lag) could only be collected from cloudwatch.

This PR aims to auto disable replications metrics when aurora is detected. For that it uses group metrics.

Also, please note it also change the fix implementation from https://github.com/signalfx/signalfx-agent/pull/1429/files.
Indeed, call 2 times `rows.Close()` function when everything works is not efficient.

### DIFF
--- a/docs/monitors/postgresql.md
+++ b/docs/monitors/postgresql.md
@@ -35,6 +35,16 @@ Tested with PostgreSQL `9.2+`.
 
 If you want to collect additional metrics about PostgreSQL, use the [sql monitor](./sql.md).
 
+## Metrics about Replication
+
+Replication metrics could not be available on some PostgreSQL servers. For now, this monitor 
+automatically disable `replication` metrics group if it detects Aurora to avoid following error:
+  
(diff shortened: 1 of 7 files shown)


keitwb:
@xp-1000 Can you test the commit I pushed here against Aurora and make sure it still works?  I realized that both something the monitor was doing previously as well as the query you added to test for Aurora were not robust against transient database outages that happened when the agent starts up.  One of the principles of the agent is that the agent will eventually converge on monitoring a service regardless of whether that service is up or down at agent startup.

xp-1000:
> @xp-1000 Can you test the commit I pushed here against Aurora and make sure it still works? I realized that both something the monitor was doing previously as well as the query you added to test for Aurora were not robust against transient database outages that happened when the agent starts up. One of the principles of the agent is that the agent will eventually converge on monitoring a service regardless of whether that service is up or down at agent startup.

Tested on Aurora and no aurora and everything works fine @keitwb

Issue #1399, opened 2020-07-27
Varnish integration does not work


xp-1000:
Hello,

I try to setup varnish integration https://docs.signalfx.com/en/latest/integrations/integrations-reference/integrations.varnish.html without success, I always got ther following error:

```
Jul 27 08:33:43 ip-10-10-30-14 signalfx-agent[22451]: time="2020-07-27T08:33:43Z" level=error msg="varnish plugin: Unable to open connection." collectdInstance=monitor-10
Jul 27 08:33:43 ip-10-10-30-14 signalfx-agent[22451]: time="2020-07-27T08:33:43Z" level=info msg="read-function of plugin `varnish/localhost' failed. Will suspend it for 20.000 seconds." collectdInstance=monitor-10
```

I tried to use a custom collectd template:

```
  - type: collectd/custom
    template: |
     LoadPlugin varnish
     <Plugin "varnish">
       <Instance>
         CollectCache       true
         CollectConnections true
         CollectBackend     true
         CollectSHM         true
         CollectESI         true
         CollectFetch       true
         CollectHCB         true
         CollectSMA         true
         CollectSMS         true
         CollectTotals      true
         CollectWorkers     true
         CollectUptime      true
         CollectVCL         true
         CollectStruct      true
         CollectObjects     true
         CollectSession     true
         CollectVSM         true
       </Instance>
     </Plugin>
```

I tested varnish5 and varnish6 on debian: same result

I tested add `-n toto` to `varnishd` process and change `<Instance toto>` on collectd conf : same result.

Someone could help me?

keitwb:
Quite possibly it is because we link against libvarnish 4.x due to some issues we had with linking collectd with the newer varnish libraries a while back.  For that reason, we never made it a "first-class" fully-supported monitor in the agent.

It looks like Varnish 5+ has a JSON output for the varnishstats cli tool.  You could use the [python-monitor](https://github.com/signalfx/signalfx-agent/blob/master/docs/monitors/python-monitor.md) monitor to write a little python script to scrape the JSON output of a subprocess call to varnishstats and convert them to metrics and do it exactly how you want.  That seems to be what other agents do.  I think the actual metrics come from a shared memory segment on the Varnish process, so it's going to be pretty tough to access except through either libvarnish C library or varnishstats cli tool.

xp-1000:
I think https://github.com/signalfx/signalfx-agent/pull/1446 should resolve this issue

PR #1446, opened 2020-08-30
#1399 add varnish telegraf based monitor

xp-1000:
hello
this should address https://github.com/signalfx/signalfx-agent/issues/1399 and more.
tested on varnish 5 and 6.


xp-1000:
Hello @keitwb 

honnestly I don't know why intergrations tests failed here, I run it multiple times on my laptop and it works fine: 

```
[gw5] [ 94%] PASSED tests/monitors/telegraf_varnish/varnish_test.py::test_telegraf_varnish_metrics 
```

could you re run the CI ?

keitwb:
> Hello @keitwb
> 
> honnestly I don't know why intergrations tests failed here, I run it multiple times on my laptop and it works fine:
> 
> ```
> [gw5] [ 94%] PASSED tests/monitors/telegraf_varnish/varnish_test.py::test_telegraf_varnish_metrics 
> ```
> 
> could you re run the CI ?

I suspect you have `/usr/bin/varnishstat` installed on your laptop.  You made a function to run a varnish container but aren't ever running it.  Even if you did, it seems unreachable by the telegraf/varnish code since it assumes that varnish is running on the same host/namespace as the agent (I don't see any config for host or port, only the varnishstat binary path and instance name).  I'm open to this monitor but it seems to have sever restrictions and doesn't appear to possibly work in a containered environment at all.

xp-1000:
> > Hello @keitwb
> > honnestly I don't know why intergrations tests failed here, I run it multiple times on my laptop and it works fine:
> > ```
> > [gw5] [ 94%] PASSED tests/monitors/telegraf_varnish/varnish_test.py::test_telegraf_varnish_metrics 
> > ```
> > 
> > 
> > could you re run the CI ?
> 
> I suspect you have `/usr/bin/varnishstat` installed on your laptop. You made a function to run a varnish container but aren't ever running it. Even if you did, it seems unreachable by the telegraf/varnish code since it assumes that varnish is running on the same host/namespace as the agent (I don't see any config for host or port, only the varnishstat binary path and instance name). I'm open to this monitor but it seems to have sever restrictions and doesn't appear to possibly work in a containered environment at all.

you are right @keitwb, it was late yesterday and I did a mess ^^ 

I will need the varnish package installed where the tests are run (and the varnish should be started obviously).

xp-1000:
@keitwb I am not sure I can update the docker dev image myself and the only way I see to pass tests is to install varnish package in dev image. I updated the Dockerfile.dev for local dev usage but I see circle ci uses image from quay.io which should be updated no ?

keitwb:
You can install it in https://github.com/signalfx/signalfx-agent/blob/master/.circleci/scripts/setup-integration-tests.sh but you will still have the issue of how to get varnishstat to talk to a varnish instance in the container.  You could run varnish straight on the CI box outside a container but I would rather not if possible.  I think when I checked how it works before it does some stuff with a shared memory block or something, so it might be tough.

xp-1000:
> You can install it in https://github.com/signalfx/signalfx-agent/blob/master/.circleci/scripts/setup-integration-tests.sh but you will still have the issue of how to get varnishstat to talk to a varnish instance in the container. You could run varnish straight on the CI box outside a container but I would rather not if possible. I think when I checked how it works before it does some stuff with a shared memory block or something, so it might be tough.

yes, it seems little tricky, do you think can we omit tests exceptionally for this monitor @keitwb ?

keitwb:
Yes, that's fine.  We already know the telegraf shim and the telegraf plugin works from existing tests so as long as you are confident in the operation of it for your needs it should be fine.

xp-1000:
sorry @keitwb I see the following error on go-lint:

```
level=warning msg="[runner] Can't run linter unused: buildssa: analysis skipped: errors in package: [/__w/1/s/pkg/monitors/telegraf/monitors/varnish/varnish.go:53:64: Varnish not declared by package varnish]"
```

I confess it is not obvious for me may be it speak to you ? else I will check that tomorrow ;)

keitwb:
The linter failure was on Windows.  Apparently that plugin isn't supported on Windows in Telegraf, so it can't be in the agent either.  I disabled it building on Windows.

xp-1000:
hello @keitwb, me again ^^ could you just confirm the ci failed is not due to my changes ? or do i still have things to do?

keitwb:
Yes, CI failures are some random other thing.  Looks good.  Thanks for submitting this.

PR #1454, opened 2020-09-10
FEED-3393: Upgrade to openjdk-10

jchengsfx:


Dockerfile
```diff
@@ -155,7 +159,7 @@ ENV CXXFLAGS $CFLAGS
 
 # In the bundle, the java plugin so will live in /lib/collectd and the JVM
 # exists at /jre
-ENV JAVA_LDFLAGS "-Wl,-rpath -Wl,\$\$\ORIGIN/../../jre/lib/${TARGET_ARCH}/server"
+ENV JAVA_LDFLAGS "-Wl,-rpath -Wl,\$\$\ORIGIN/../../jre/lib/server"
```
keitwb, line 162: Make sure this is tested with the standalone bundle as well (if our automated tests don't already cover Java stuff for that).  This makes me wary changing this stuff.

jchengsfx: > Make sure this is tested with the standalone bundle as well (if our automated tests don't already cover Java stuff for that). This makes me wary changing this stuff.

This is tested in `bundle_tests`, and failed before this change due to the new path.

https://app.circleci.com/pipelines/github/signalfx/signalfx-agent/8430/workflows/2ea832c4-a696-4e05-bbcc-a8fe67dd853f/jobs/238429/steps

```b'time="2020-09-10T15:21:59Z" level=error msg="dlopen (\\"/opt/signalfx-agent/lib/collectd/java.so\\") failed: libjvm.so: cannot open shared object file: No such file or directory. The most common cause for this problem is missing dependencies. Use ldd(1) to check the dependencies of the plugin / shared object." collectdInstance=global'
b'time="2020-09-10T15:21:59Z" level=info msg="ERROR: dlopen (\\"/opt/signalfx-agent/lib/collectd/java.so\\") failed: libjvm.so: cannot open shared object file: No such file or directory. The most common cause for this problem is missing dependencies. Use ldd(1) to check the dependencies of the plugin / shared object." collectdInstance=global'
b'time="2020-09-10T15:21:59Z" level=error msg="Load plugin \\"java\\" failed with status 1." collectdInstance=global plugin=plugin_load'
b'time="2020-09-10T15:21:59Z" level=warning msg="Found a configuration for the `java\' plugin, but the plugin isn\'t loaded or didn\'t register a configuration callback." collectdInstance=global'
b'time="2020-09-10T15:21:59Z" level=warning msg="Found a configuration for the `java\' plugin, but the plugin isn\'t loaded or didn\'t register a configuration callback." collectdInstance=global'
b'time="2020-09-10T15:21:59Z" level=warning msg="There is a `Plugin\' block within the configuration for the java plugin. The plugin either only expects \\"simple\\" configuration statements or wasn\'t loaded using `LoadPlugin\'. Please check your configuration." collectdInstance=global'
b'time="2020-09-10T15:21:59Z" level=info msg="Error: Reading the config file failed!" collectdInstance=global'
b'time="2020-09-10T15:21:59Z" level=info msg="Read the logs for details." collectdInstance=global'
b'time="2020-09-10T15:21:59Z" level=error msg="Collectd died when it was supposed to be running, restarting..." collectdInstance=global'
```

keitwb (approved):
Did you confirm that 10 is backwards compatible with 8?

jchengsfx:
> Did you confirm that 10 is backwards compatible with 8?

According to https://docs.oracle.com/javase/10/migrate/:
```If an application runs in JDK 8, then it will run on JDK 9 and later releases as long as it uses APIs that are supported and intended for external use.```

But not sure specifically what we need to watch for.

Issue #1455, opened 2020-09-11
Have explicit user Agent for HTTP checks


BzSpi:
When the agent performs HTTP checks, the user agent used currently used is `Go-http-client/1.1`.

Is it possible to change it to something like `SignalFx SmartAgent/<agent-version>` ?

BzSpi:
Also, if the default value could be overridden from the configuration, this would be fine.

keitwb:
The related PR was merged and should be doable now.

PR #1456, opened 2020-09-11
HTTP Monitor/User-Agent Improvements

keitwb:
See individual commits.

This addresses #1455.


keitwb:
@jrcamp can you review since you did the HTTP client?

Issue #1458, opened 2020-09-14
HTTP monitor: configured URL should be sent as metadata along with the redirection URL


BzSpi:
Currently the `url` metadata sent by the agent is the final redirection URL if any.
This causes any detector configured on this metric to stay unresolved if it's triggered while the redirection changed.

One quick example: we check the status code on a URL monitored with redirection. In case the application shows a maintenance page with an undesired status code (4xx or 5xx), the alert will never recover when application will go back to normal.

Expected behavior:
* `url` metadata sent is the one in the agent configuration 
* a new `last_url` metadata is sent with the current value

PR #1466, opened 2020-09-21
http monitor: optional aggregation and last url option

xp-1000:
Hello @keitwb 

related to https://github.com/signalfx/signalfx-agent/pull/1459 and https://github.com/signalfx/signalfx-agent/issues/1458

thanks for taking care of that during my holidays. Nevertheless, I would like to change little the behavior.

- remove `original_url` and restore the original purpose to `url`.
- add option to add optional dimension for `last_url`, false by default to keep compat. this is sent only if different from original url.
- add url normalization to force consistency between client http config and deprecated url lists. this also allow to compare url and lastURL.
- log as debug a redirected url for troubleshooting purpose (even when `addLastURL: false`)

I understand the purpose of the previous PR and I think this one respects the logic and the ability to have a "dynamic" dimension which change when redirection changes even if it uses last url instead of original one but:

- I think it is a little less breaking now while `last_url` dimension is disabled by default and the existing `url` one has the same purpose than originally.
- avoiding dynamic change on dimension will help implement a "generic" detector for this. indeed, while `last_url` can change time to time this will generate false alert without explicit aggregation (on `url` to ignore `last_url`). without `original_url` or `last_url` we can simply not use aggregation and this will work in every cases (not matter the env, standard deployment or not, `disabledHostDimensions: true` ...)

Additional context: for now, recent changes are broke our http detectors https://github.com/claranet/terraform-signalfx-detectors/blob/master/network/http/ and I prefer to not have to use aggregation to keep compatibility everywhere.

thanks in advance 

here is a sample result:
image

with following conf: 

```yaml
  - type: http
    host: "splunk.com"
    #port: 80
    path: "/en_us/investor-relations/acquisitions/signalfx.html"
    addLastURL: true
  - type: http
    host: "signalfx.com"
    #port: 80
    addLastURL: true
  - type: http
    urls:
      - "https://www.google.com"
      - "https://google.com"
      - "http://monip.org/"
    addLastURL: true
```

### DIFF
--- a/docs/monitors/http.md
+++ b/docs/monitors/http.md
@@ -57,6 +57,7 @@ Configuration](../monitor-config.md#common-configuration).**
 | `urls` | no | `list of strings` | DEPRECATED: list of HTTP URLs to monitor. Use `host`/`port`/`useHTTPS`/`path` instead. |
 | `regex` | no | `string` | Optional Regex to match on URL(s) response(s). |
 | `desiredCode` | no | `integer` | Desired code to match for URL(s) response(s). (**default:** `200`) |
+| `addLastURL` | no | `bool` | Add `last_url` dimension which could differ from `url` when redirection is followed. (**default:** `false`) |
 
 
 ## Metrics
@@ -67,8 +68,10 @@ Metrics that are categorized as
 (*default*) are ***in bold and italics*** in the list below.
 
 
- - ***`http.cert_expiry`*** (*gauge*)<br>    Certificate expiry in seconds. This metric is reported only if HTTPS is available (from last followed URL).
+ - ***`http.cert_expiry`*** (*gauge*)<br>    Certificate expiry in seconds. This metric is reported only if HTTPS  is available (from last followed URL).
+
  - ***`http.cert_valid`*** (*gauge*)<br>    Value is 1 if certificate is valid (including expiration test) or 0 else. This metric is reported only if HTTPS is available (from last followed URL).
+
  - ***`http.code_matched`*** (*gauge*)<br>    Value is 1 if `status_code` value match `desiredCode` set in config. Always reported.
  - ***`http.content_length`*** (*gauge*)<br>    HTTP response body length. Always reported.
  - ***`http.regex_matched`*** (*gauge*)<br>    Value is 1 if pattern match in response body. Only reported if `regex` is configured.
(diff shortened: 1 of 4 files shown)


pkg/monitors/http/http.go
```diff
@@ -146,6 +158,29 @@ func (m *Monitor) Shutdown() {
 	}
 }
 
+func (m *Monitor) forgeURL(site string) (normalizedURL *url.URL, err error) {
```
keitwb, line 161: Call this `normalizeURL` -- "forge" is what you do to hot iron.

docs/monitors/http.md
```diff
@@ -92,10 +95,9 @@ dimensions may be specific to certain metrics.
 
 | Name | Description |
 | ---  | ---         |
+| `last_url` | Last URL retrieved (after redirects) from configured one. Only sent if `noRedirects: false` and `addLastURL: true` and if URL responds with a redirect different from the original `url`. |
```
keitwb, line 98: I would change everywhere it is called the "last" URL to be the "redirect" URL (e.g. `redirect_url`) instead.  This makes it immediately clear what it represents, at least clearer than "last" is.

keitwb:
@xp-1000 Seems good except for the two minor comments I had on naming.  Normally we wouldn't do this kind of breaking changes but I think we can consider this monitor in beta still and I'm pretty sure you all at Claranet are the only ones using it so far.

xp-1000:
> @xp-1000 Seems good except for the two minor comments I had on naming. Normally we wouldn't do this kind of breaking changes but I think we can consider this monitor in beta still and I'm pretty sure you all at Claranet are the only ones using it so far.

I totally understand this is a breaking change and I thank you for the exception. 
In fact the recent change from https://github.com/signalfx/signalfx-agent/pull/1459 is also considered as breaking change for us and this PR should allow to:

- respect the goal of the previous PR
- but resolve the breaking change inserted in v5.5.0

should be the best of both worlds I think :)

pkg/monitors/http/http.go
xp-1000: ^^ sorry, my english ...

docs/monitors/http.md
xp-1000: should be good now

keitwb (approved):

PR #1472, opened 2020-09-29
default rollup for cumulative counters is delta

xp-1000:
Hello,

I don't think the FAQ change is crucial but I did not find the repo where is stored the official doc to update https://docs.signalfx.com/en/latest/charts/resolution-rollups.html#rollups with right rollup (i.e. delta by default for cumulative counter)


keitwb (approved):
Thanks for fixing this -- we changed from rate/sec to delta as the default a few months ago.

PR #1477, opened 2020-09-29
Extract span correlation into separate library

jrcamp:
This extracts the span correlation (but leaves span annotation/enrichment) from signalfx-agent into a smaller `lib` module meant to be used by other agents (ie otel). The only new dependency is zap. It's a bit weird to now have 2 logging libraries in the same process. Zap will be configured to log in JSON if logrus is configured to.

Testing still in progress.


Dockerfile
```diff
@@ -18,12 +18,14 @@ ENV GOPATH=/go
...
 
 COPY go.mod go.sum ./
+COPY lib/go.mod lib/go.sum ./lib/
 RUN go mod download
 
 COPY cmd/ ./cmd/
 COPY scripts/collectd-template-to-go scripts/make-versions ./scripts/
 COPY Makefile .
 COPY pkg/ ./pkg/
+COPY lib/ ./lib/
```
keitwb, line 28: Convention is to put library code in pkg as described here: https://github.com/golang-standards/project-layout#pkg.  We put all the agent code in `pkg` because some people use it as a library.

Just put it in `pkg/core/correlation` or something.  You can still give it it's own go.mod if you really need it to be a separate module, just like we do in collector-contrib.

lib/correlations/client.go
```diff
@@ -59,6 +57,7 @@ type request struct {
 // Client is a client for making dimensional correlations
 type Client struct {
 	sync.RWMutex
+	log           *zap.Logger
```
keitwb, line 60: Not a big fan of pulling in another logging library just because collector happens to use it.  Maybe try and make the interface return errors as appropriate and let the users of the library print errors with whatever log library they want.  Not sure if that is possible given the data flows.

keitwb (changes requested):
Will finish reviewing in a bit.

lib/correlations/client.go
jrcamp: Yeah not really possible to return errors as there's a lot of informational messages as well. I think the only option is having a generic logging interface similar to slf4j in the Java world. I'll see if I can find one that's already out there we can use.

jrcamp: The logging library wrapper I found wasn't able to work with how we pass zap loggers around in OT so I just added a simple wrapper instead.

Dockerfile
jrcamp: I put it in pkg/apm (It already contains a package called correlation so easier to put it under something else).

keitwb (approved):

keitwb:
I think this broke the perf test.

PR #1478, opened 2020-09-30
Add writer block agent configuration to helm chart

edwinf:
This PR includes a new feature and a minor bug fix:

-  (feat) Adds the writer map to the helm chart for greater flexibility when customizing the writer configuration.
-  (fix) solves a coalesce warning when adding globalDimensions by initializing the value in the values file as an empty map.

The values file used to test both of these updates where

```
writer:
  traceExportFormat: sapm
  addGlobalDimensionsAsSpanTags: true
globalDimensions:
  environment: dev
```


keitwb (approved):

PR #1481, opened 2020-10-01
Add pod labels as datapoint dimensions

kerbyhughes:
We would like for resource labels, such as Pod labels, to show up as metadata on datapoints so they can be pulled in with other metadata with the `signalfx-go` client when queried via Signalflow.

Currently, only `kubernetes_namespace`, `kubernetes_pod_uid`, `kubernetes_pod_name`, and `kubernetes_node` are visible, this change would add label values by adding them to the dimensions map.

If there are issues with changing pod dimensions, it might be possible to change the function to use `dimensions.NewWithMeta` and place the labels in the Meta map, however I am not yet sure that would be picked up in the signalflow metadata.

### DIFF
--- a/pkg/monitors/kubernetes/cluster/metrics/pods.go
+++ b/pkg/monitors/kubernetes/cluster/metrics/pods.go
@@ -23,6 +23,9 @@ func datapointsForPod(pod *v1.Pod) []*datapoint.Datapoint {
 		"kubernetes_node":      pod.Spec.NodeName,
 	}
 
+	props := dimensionForPod(pod).Properties
+	dimensions = datapoint.AddMaps(dimensions, props)
+
 	dps := []*datapoint.Datapoint{
 		datapoint.New(
 			"kubernetes.pod_phase",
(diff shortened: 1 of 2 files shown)


keitwb:
> so they can be pulled in with other metadata with the `signalfx-go` client when queried via Signalflow.

This is achievable already because of the properties added to the `kubernetes_pod_uid` dimension by the `kubernetes-cluster` monitor.  In the MTSs that you query for in SignalFlow programs, both dimensions and dimension properties are available to use with filter, group by, etc.

The SignalFlow program itself is identical regardless of whether labels are dimension or dimension properties.

keitwb:
Also if you want properties to be "promoted" to be programatically useable in Go code using `signalfx-go` you can use the [promote method](https://developers.signalfx.com/signalflow_analytics/methods/promote_stream_method.html)

kerbyhughes:
@keitwb Thank you for the quick reply! I was unaware of the promote method - that did exactly what I needed after figuring out how it sanitized some characters in our custom labels.

PR #1483, opened 2020-10-06
Make default dimensions to correlate to configurable

jrcamp:
These dimensions are used to correlate services and environments to. In otel
these will need to correlate using otel semantic conventions if we are sending
metrics in that format (ie not using the SignalFx translations).


keitwb (approved):
Looks good.  As far as I can tell this is no difference as far as the smart agent is concerned.

PR #1484, opened 2020-10-07
Upgrade openjdk to 1.11

jchengsfx:


Dockerfile
```diff
@@ -255,13 +255,13 @@ RUN apt update &&\
 
 RUN add-apt-repository ppa:openjdk-r/ppa &&\
     apt update &&\
-    apt install -y openjdk-10-jdk maven
+    apt install -y openjdk-11-jdk maven
```
keitwb, line 258: Make sure none of the development tools are going into the bundle.  It should just be the JRE with the core binaries and std library classes.

jchengsfx: It'll currently has:
```drwxr-xr-x  1 root root 4096 Oct  6 21:14 bin
drwxr-xr-x  4 root root 4096 Oct  6 21:14 conf
drwxr-xr-x  3 root root 4096 Oct  6 21:14 docs
drwxr-xr-x  3 root root 4096 Oct  6 21:14 include
drwxr-xr-x  2 root root 4096 Oct  6 21:14 jmods
drwxr-xr-x 72 root root 4096 Oct  6 21:14 legal
drwxr-xr-x  6 root root 4096 Oct  6 21:14 lib
-rw-r--r--  1 root root 1161 Oct  6 21:14 release
```

Should everything except `bin` and `lib` be removed?

keitwb: Not sure, make sure the tar bundle works on a system without any JRE installed.

keitwb: Perferably one that isn't Ubuntu/Debian based either.

jchengsfx: Alpine and centos are included in our bundle integration tests.

keitwb (approved):

PR #1486, opened 2020-10-07
fix `kubernetes.container_cpu_request` and `kubernetes.container_cpu_limit`

atishpatel:
`kubernetes.container_cpu_request` and `kubernetes.container_cpu_limit` are always 0 or 1. 

I believe the issue is here:
/signalfx-agent/pkg/monitors/kubernetes/cluster/metrics/containers.go -> func datapointsForContainerSpec 

// CPU, in cores. (500m = .5 cores)
ResourceCPU ResourceName = "cpu"

CPU resources are set as xm which is equal to .00x. (500m -> .5 cpu or 5m -> .005 cpu)

I am not familiar with this system enough to run tests but this would be a simple fix. Feel free to modify this pull request as you see fit and test it. Thank you.

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -1,6 +1,6 @@
 FROM registry.access.redhat.com/rhel7-atomic
 
-ARG AGENT_VERSION="5.5.1"
+ARG AGENT_VERSION="5.5.2"
 LABEL name="SignalFx Smart Agent" \
 	  maintainer="SignalFx, Inc." \
 	  vendor="SignalFx, Inc." \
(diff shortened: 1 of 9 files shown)


pkg/monitors/kubernetes/cluster/metrics/containers.go
```diff
@@ -37,7 +37,7 @@ func datapointsForContainerSpec(c v1.Container, contDims map[string]string) []*d
 			datapoint.New(
 				"kubernetes.container_cpu_request",
 				contDims,
-				datapoint.NewIntValue(val.Value()),
```
keitwb, line 40: Really it should be

`datapoint.NewFloatValue(float64(val.MilliValue())/1000.0)`

But I'll test and update this PR once I've verified that that works.

Thanks for bringing this up.

atishpatel: I'm fine with either that `datapoint.NewFloatValue(float64(val.MilliValue())/1000.0)` or `datapoint.NewIntValue(val.MilliValue())`.

keitwb:
Ok, @atishpatel I continued this at #1487 since I messed up your fork somehow and github closed the PR and won't let me do anything with it.

PR #1487, opened 2020-10-08
kubernetes-cluster Monitor: Use floating point value for CPU limit/re…

keitwb:
…quest

Previously these metrics were truncated to the nearest integer reducing
their usefulness.

This is a replacement for #1486 which got messed up somehow when I pushed an update.


chien-splunk:
I'm finding that my pods that have a limit < 1000millicores is being rounded up to 1. I think it stems from the [val.MilliValue](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity.MilliValue) doing a ceil() behind the scenes. So we're effectively rounding all limits <1 to 1 CPU...

PR #1489, opened 2020-10-08
make correlation config settings more intuitive

jrcamp:
This doesn't change the config settings in the agent yaml, just internal names such that they align with names used in otel and that more accurately reflect their usage.


keitwb (approved):
LGTM

PR #1491, opened 2020-10-12
Update README.md

slernersplunk:
Add updated info for Openshift and the included default node selector in the daemonset file.
Both of these prevented RedHat from deploying these .yaml configurations until fixed.

### DIFF
--- a/deployments/k8s/README.md
+++ b/deployments/k8s/README.md
@@ -22,6 +22,13 @@ A few things to do before deploying these directly (when **not using Helm**):
 
 	`kubectl create secret generic --from-literal access-token=MY_ACCESS_TOKEN signalfx-agent`
 
+ 4. Doublecheck the `daemonset.yaml` included stanza and remove if not consistent with your environment.
+
+```    
+	nodeSelector:
+        kubernetes.io/os: linux
+	```
+
 Then to deploy run the following from the present directory:
 
 `cat *.yaml | kubectl apply -f -`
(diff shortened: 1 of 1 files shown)


deployments/k8s/README.md
```diff
@@ -22,6 +22,13 @@ A few things to do before deploying these directly (when **not using Helm**):
 
 	`kubectl create secret generic --from-literal access-token=MY_ACCESS_TOKEN signalfx-agent`
 
+ 4. Doublecheck the `daemonset.yaml` included stanza and remove if not consistent with your environment.
```
brianashby-sfx, line 25: do you have to just remove it if it's not the same as your environment, or make sure that it's consistent with your environment?

keitwb: @slernersplunk what other values were you seeing for this?  The problem is if you remove it and you have Windows nodes, it will make a bunch of errors trying to deploy the Linux agent to Windows.

deployments/k8s/README.md
```diff
@@ -85,6 +92,11 @@ The [serverless](./serverless) directory contains a set of sample YAML
 resources that you can start from to deploy the agent in a serverless K8s
 environment.
 
+## OpenShift
+
+`configmap.yaml` should have a value at first indent setting `internalHost: ${MY_NODE_IP}`  
+This lets the SignalFx SmartAgent run the `signalfx-agent status` command normally.
```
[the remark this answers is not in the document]
keitwb: It uses localhost for the status command normally.  Does this mean localhost doesn't work right on OpenShift containers?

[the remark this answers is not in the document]
slernersplunk: We saw this at RedHat running their own OpenShift cluster:

```
Could not get status: Get "http://localhost:8095/?section=": dial tcp: lookup localhost on 10.27.167.107:53: no such host
Are you sure the agent is currently running
```

Harnit figured out to do that and it fixed it- I'm not saying its a must, but we should put the suggestion in our docs so that people can find that answer quickly if they have the same problem.

deployments/k8s/README.md
slernersplunk: Another RedHat discovery- our default entry for node selector tanked the SmartAgent pods... removing it fixed the problem...  I'm not saying we SHOULD remove it- i'm saying IF that nodeselector is inconsistent with one's environment, remove it...

slernersplunk: Trying to get these things that block us for 3 hours but are only a single config change into our docs as suggestions so people can find them- btw this was also in OpenShift

keitwb (approved):

PR #1492, opened 2020-10-13
Make correlation source span attribute and destination dimension configurable

jrcamp:
* Allow using of different span attributes from destination dimension
   This allows the collector to map new attribute names to old names, new names, or both.
* Improve logging of correlation struct

### DIFF
--- a/pkg/apm/correlations/client.go
+++ b/pkg/apm/correlations/client.go
@@ -126,7 +126,7 @@ func (cc *Client) putRequestOnChan(r *request) error {
 		// and because this isn't being taken off on the request sender and subject to retries, this could
 		// potentially spam the logs
 		atomic.AddInt64(&cc.TotalInvalidDimensions, int64(1))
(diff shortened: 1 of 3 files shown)


pkg/apm/correlations/client.go
```diff
@@ -126,7 +126,7 @@ func (cc *Client) putRequestOnChan(r *request) error {
 		// and because this isn't being taken off on the request sender and subject to retries, this could
 		// potentially spam the logs
 		atomic.AddInt64(&cc.TotalInvalidDimensions, int64(1))
-		cc.log.WithFields(log.Fields{"method": r.operation, "correlation": r.Correlation}).Debug("No dimension key or value to correlate to")
+		cc.log.WithFields(log.Fields{"method": r.operation}).WithFields(r.Fields()).Debug("No dimension key or value to correlate to")
```
keitwb, line 129: Can the `r.Fields` method instead be `r.Logger` and return a new logger instead so that this could be something like:

`r.Logger(cc.log).Debug("No dimension...")`

That seems a lot clearer and easier to read.

keitwb (approved):

PR #1496, opened 2020-10-16
Extends cassandra metrics

xp-1000:
hello,

adding exceptions storage metrics and fix the type for existing ones.


keitwb (approved):
Looks good, thanks.

PR #1500, opened 2020-10-18
add basic tomcat collectd genericjmx based monitor

xp-1000:
hello,

here is a basic tomcat monitor based on collectd/genericjmx


keitwb (approved):
Seems pretty straight forward.  Thanks for submitting this.

PR #1501, opened 2020-10-18
nginx_ingress

xp-1000:
vts has been removed since nginx ingress 0.16, this new monitor will allow to easily monitors newer version.


keitwb (approved):
Thanks for adding this!

PR #1502, opened 2020-10-18
add velero prometheus based monitor

xp-1000:


pkg/monitors/prometheus/velero/metadata.yaml
```diff
@@ -0,0 +1,47 @@
...
+    This is recommended to use service discovery:
+
+    ```yaml
+    monitors:
+    - type: prometheus/velero
+      discoveryRule: container_image =~ "velero" && port == 8085
+      port: 8085
+    ```
+
+  metrics:
```
keitwb, line 27: Is this just a subset of the metrics exposed by Valero or is this all of them?

keitwb:
I'm not so sure about having first-class monitors for everything prometheus exporter-based.  The main reason we have some of these exporter 'wrapper' monitor in the first place is to have automatic metric filtering when some of the metrics are included and others are custom.  If everything is custom then it doesn't really help anything.

Can you give some good use case for why this is helpful to you?

xp-1000:
> I'm not so sure about having first-class monitors for everything prometheus exporter-based. The main reason we have some of these exporter 'wrapper' monitor in the first place is to have automatic metric filtering when some of the metrics are included and others are custom. If everything is custom then it doesn't really help anything.
> 
> Can you give some good use case for why this is helpful to you?

I understand and I don't have this goal (or the time) to cover everything but:

- as a MSP we have some recurrent needs for every of our customers. For example, every kubernetes clusters are saved with `velero` for every of our customers which is, de facto, a "standard" for us. So this is simply and selfishly a convenient way to ease the configuration of collection over multiple similar projects.
- this is also a safety feature for us (and I think for others) because prometheus exporters often provide high amount of metrics which could impact billing for lot of useless custom metrics. I provide sample configuration but I cannot enforce the filter or force people to read the doc (https://github.com/claranet/terraform-signalfx-detectors/tree/master/container/kubernetes/velero#notes).
- both of previous statements are opinionated and easy to manage for few orgs but difficult to scale for lot of orgs/customers/projects..
- I think it is a good thing for signalfx to propose first citizen and popular softwares "native" monitors (i.e. I often have to deal with "datadog fan boys" who often argue that signalfx propose less "integrations", some do not understand the extensibility of the tool thanks to standards protocols monitors like prometheus, statsd .. but only compare the monitors list in the doc).
- I am not sure how but I still want to propose dashboards to signalfx to make them "builtin". this should improve all previous statements (help signalfx and us) and having a native monitor seems mandatory for that.

To resume, it is an help to ease integration at scale of popular tools and avoid mistake adding thousand of undesired custom metrics. In my opinion, nginx ingress is a first citizen tool to have (https://github.com/signalfx/signalfx-agent/pull/1501 merged so cool!) but velero this is very questionable.
So I will let you decide for this one. if you prefer to not include it I will provide a terraform module for dashboard instead of builtin in signalfx, this is not a big deal.

mstumpfx:
Hi @xp-1000 

>I am not sure how but I still want to propose dashboards to signalfx to make them "builtin". this should improve all previous statements (help signalfx and us) and having a native monitor seems mandatory for that.

I help maintain a lot on our built-in content and would be happy to work with you on this. We don't have a formal process (yet) for customer proposed dashboards, but if you have a dashboard built already, or are building one, you can pass along a link to that dashboard to me for a review and we can start that conversation. You can find my email in my profile, I think.

xp-1000:
> Hi @xp-1000
> 
> > I am not sure how but I still want to propose dashboards to signalfx to make them "builtin". this should improve all previous statements (help signalfx and us) and having a native monitor seems mandatory for that.
> 
> I help maintain a lot on our built-in content and would be happy to work with you on this. We don't have a formal process (yet) for customer proposed dashboards, but if you have a dashboard built already, or are building one, you can pass along a link to that dashboard to me for a review and we can start that conversation. You can find my email in my profile, I think.

Hello @mstumpfx, thanks for your answer, this will be a real pleasure. I will contact you soon :)

pkg/monitors/prometheus/velero/metadata.yaml
xp-1000: yes, just a subset, velero expose lot of metrics (in addition to default golang metrics).

keitwb (approved):
Ok, if your primary reason for it is the filtering, then that is a good use case.

PR #1505, opened 2020-10-20
cassandra: add total, caswrite and casread latencies

xp-1000:
hello,

after https://github.com/signalfx/signalfx-agent/pull/1496 this one is focus on latencies.

especially it add CASWrite and CASRead latencies (fully similar as existing) which is very crucial to monitor (I already had real production use case where read/write are not enough to see problem)

also it add the total latency (in microseconds) which will allow to calculate the real time latency which seems to be a good info for alerting.

### DIFF
--- a/docs/monitors/collectd-cassandra.md
+++ b/docs/monitors/collectd-cassandra.md
@@ -91,6 +91,18 @@ Metrics that are categorized as
 This monitor will also emit by default any metrics that are not listed below.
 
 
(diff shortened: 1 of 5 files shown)


keitwb (approved):
Looks good, thanks for adding these.

xp-1000:
> Looks good, thanks for adding these.

thanks @keitwb do you have any idea of the next version of the agent including this ? 

for now I assumed it could be 5.5.5 : https://github.com/claranet/terraform-signalfx-detectors/pull/152/files#diff-bf1d8db707a8b65103f1a2bd4a3db9fb5a4deadf72a5b1e0592d22b05607c01dR3

PR #1506, opened 2020-10-22
Change apt_key for get_url

cvauvarin:
Hi,

`apt_key` do not support proxy environment settings or apt configuration (https://github.com/ansible/ansible/issues/31691), this makes this role unsable when we have to rely on proxies to setup our instances. `apt_key` is also not anymore the recommended to install keys (https://github.com/ansible/ansible/issues/55590).

`get_url` seems to be the easiest way to deal with this problem as it support proxy environment if we set it at task or playbook level (https://docs.ansible.com/ansible/latest/user_guide/playbooks_environment.html). 

Thanks,

### DIFF
--- a/deployments/ansible/roles/signalfx-agent/tasks/debian_repo.yml
+++ b/deployments/ansible/roles/signalfx-agent/tasks/debian_repo.yml
@@ -10,10 +10,10 @@
     state: absent
 
 - name: Add an Apt signing key for Signalfx Agent
-  apt_key:
+  get_url:
     url: "{{ sfx_repo_base_url }}/signalfx-agent-deb/splunk-B3CD4420.gpg"
-    keyring: /etc/apt/trusted.gpg.d/splunk.gpg
-    state: present
+    dest: /etc/apt/trusted.gpg.d/splunk.gpg
+    mode: 0644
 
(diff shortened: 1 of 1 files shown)


xp-1000 (approved):

jchengsfx (approved):
Thanks @cvauvarin.  We'll try to release these changes to ansible galaxy soon.

saudadeandbliss:
> Hi,
> 
> `apt_key` do not support proxy environment settings or apt configuration ([ansible/ansible#31691](https://github.com/ansible/ansible/issues/31691)), this makes this role unsable when we have to rely on proxies to setup our instances. `apt_key` is also not anymore the recommended to install keys ([ansible/ansible#55590](https://github.com/ansible/ansible/issues/55590)).
> 
> `get_url` seems to be the easiest way to deal with this problem as it support proxy environment if we set it at task or playbook level (https://docs.ansible.com/ansible/latest/user_guide/playbooks_environment.html).
> 
> Thanks,

How use *get_url* for add gpg keys?

Example:

`curl -fsSL https://download.docker.com/linux/debian/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg`

PR #1511, opened 2020-10-26
Update installer for RHEL6

jchengsfx:
Manually tested on RHEL 6.10 EC2 instance.

### DIFF
--- a/deployments/installer/install.sh
+++ b/deployments/installer/install.sh
@@ -166,24 +166,18 @@ repo_for_stage() {
 }
 
 get_distro() {
-  local distro="$(. /etc/os-release && echo $ID || true)"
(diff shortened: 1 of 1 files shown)


deployments/installer/install.sh
```diff
@@ -166,26 +166,20 @@ repo_for_stage() {
...
+      distro="centos"
+    elif [ -e /etc/redhat-release ]; then
+      distro="rhel"
+    fi
   fi
 
   echo "$distro"
 }
 
-get_distro_version() {
```
jchengsfx, line 179: Removed this function since it is not used.

dloucasfx (approved):
LGTM
CI is failures don't look related to this change.

PR #1513, opened 2020-11-01
add mongodb replicaset metrics

xp-1000:
hello,

default=true because used in official builtin dashboard.


keitwb (approved):
Thanks  for adding this.

Issue #1515, opened 2020-11-03
helm configmap.yaml template does not support eventEndpointUrl 


stamak:
There is no way to pass eventEndpointUrl for configmap.yaml

keitwb:
This should be fixed via #1516 which was now merged.

PR #1516, opened 2020-11-04
Helm Add support of eventEndpointUrl (#1515)

stamak:


keitwb (approved):
Thanks for adding this.

PR #1518, opened 2020-11-04
Add expr filter config option to limit inventory collection

pmcollins:
This change lets users supply an optional "expr" filter expression which gets
evaluated per cluster. If the expression evaluates to true, metrics are collected
for all objects in that cluster, otherwise, it is skipped.

### DIFF
--- a/pkg/monitors/vsphere/model/model.go
+++ b/pkg/monitors/vsphere/model/model.go
@@ -36,6 +36,16 @@ type Config struct {
 	// all inventory objects at a time.
 	PerfBatchSize int `yaml:"perfBatchSize" default:"10"`
 
+	// An 'expr' expression to limit the inventory traversed by the monitor. Leave blank or omit
+	// to traverse and get metrics for the entire vSphere inventory. Otherwise, this expression
+	// is evaluated per cluster. If the expression evaluates to true, metrics are collected for
+	// the objects in the cluster, otherwise it is skipped. Made available to the expr expression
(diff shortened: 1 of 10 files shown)


pkg/monitors/vsphere/model/model.go
```diff
@@ -36,6 +36,16 @@ type Config struct {
...
 
+	// An 'expr' expression to limit the inventory traversed by the monitor. Leave blank or omit
+	// to traverse and get metrics for the entire vSphere inventory. Otherwise, this expression
+	// is evaluated per cluster. If the expression evaluates to true, metrics are collected for
+	// the objects in the cluster, otherwise it is skipped. Made available to the expr expression
+	// environment are the variables: `Datacenter` and `Cluster`. For example:
+	// expr: "Datacenter == 'MyDatacenter' && Cluster == 'MyCluster'"
+	// The above expr value will cause metrics collection for only the given datacenter + cluster.
+	// See https://github.com/antonmedv/expr for more advanced syntax.
+	Expr string `yaml:"expr"`
```
keitwb, line 47: Please find a better name for this config option.

keitwb: Is there anything other than datacenter and cluster available?

pmcollins: I've changed the field name to `filter`. Regarding the second comment, only datacenter and cluster are available at this level of the inventory traversal.

keitwb (approved):

Issue #1519, opened 2020-11-05
[Helm] no possibility to specify envFrom confimap environment variables


stamak:
It would be nice if there is a possibility to use *envFrom* to define all of the ConfigMap's data as container environment variables.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables

stamak:
fixed and released
thx everybody

PR #1520, opened 2020-11-05
Helm add possibility to specify confimap env vars(#1519)

stamak:
For example
```
envFrom:
  - configMapRef:
      name: signalfx-agent.env
```


keitwb (approved):
Looks good.

But I'm curious what envvars you are wanting to configure? Most things in the agent can be configured with the main agent.yaml config file.

stamak:
@keitwb basically we are using this for setting **globalDimensions**
Could you please merge this?

PR #1526, opened 2020-11-10
Eks updates

j4ys0n:
Ran into a few issues when deploying with AWS EKS.

Error 1:
`time="2020-11-10T19:48:33Z" level=error msg="Could not get summary metrics" error="failed to get pods from Kubelet URL \"ip-192-168-0-32.us-east-2.compute.internal:10250/pods\": Get \"ip-192-168-0-32.us-east-2.compute.internal:10250/pods\": unsupported protocol scheme \"ip-192-168-0-32.us-east-2.compute.internal\"" monitorType=kubelet-metrics`
resolved by specifying the correct url and port for the kubeletAPI

Error 2:
`E1110 19:59:40.404096       1 leaderelection.go:367] Failed to update lock: configmaps "signalfx-agent-leader" is forbidden: User "system:serviceaccount:corda-net:signalfx-agent" cannot update resource "configmaps" in API group "" in the namespace "corda-net"`
resolved by allowing configmaps to be updated

### DIFF
--- a/deployments/k8s/clusterrole.yaml
+++ b/deployments/k8s/clusterrole.yaml
@@ -32,7 +32,7 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - clustermaps
+  - configmaps
   verbs:
   - get
   - list
--- a/deployments/k8s/clusterrole.yaml
+++ b/deployments/k8s/clusterrole.yaml
@@ -32,6 +32,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - clustermaps
+  verbs:
+  - get
+  - list
+  - update
+- apiGroups:
+  - ""
+  resources:
   - nodes/stats
   verbs:
   - get
--- a/deployments/k8s/configmap.yaml
+++ b/deployments/k8s/configmap.yaml
@@ -54,6 +54,7 @@ data:
     - type: kubelet-metrics
       kubeletAPI:
         authType: serviceAccount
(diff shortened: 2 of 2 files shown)


deployments/k8s/clusterrole.yaml
```diff
@@ -29,6 +29,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
keitwb, line 32: Config map permissions was split out to a separate role: https://github.com/signalfx/signalfx-agent/blob/master/deployments/k8s/configmap-role.yaml.  

Perhaps you were using an older Helm chart?  The latest has been released as of a few days ago.  

This should be unnecessary.  The agent should not access configmaps outside of its own namespace, which is why we split out that permission to a separate role and not a cluster role.

deployments/k8s/configmap.yaml
```diff
@@ -54,6 +54,7 @@ data:
     - type: kubelet-metrics
       kubeletAPI:
         authType: serviceAccount
+        url: https://${MY_NODE_NAME}:10250
```
keitwb, line 57: Kubernetes has changed this default so many times that we don't really want to specify it here because it is pretty much guaranteed to be wrong for a significant subset of users.

emaderer:
Hey @j4ys0n, what's the status of this PR? can we close it?

emaderer:
Closing this since it is really stale at this point.

Issue #1532, opened 2020-11-17
url for repo_base in puppet deployment doesn't work for windows OS


mud5150:
`splunk.jfrog.io/splunk` doesn't appear to host any windows packages. I found a working url `dl.signalfx.com` in the chef deployment. It looks like there is an official chocolatey package. It seems like this would be the best way to go long term. Submitting a PR to update to the working hostname for now.

PR #1533, opened 2020-11-17
fixes repo base url for windows puppet deployments

mud5150:
Fixes issue #1532


jchengsfx:
@mud5150 We explicitly set `repo_base => 'dl.signalfx.com'` for windows (see https://github.com/signalfx/signalfx-agent/blob/master/deployments/puppet/manifests/init.pp#L84).  Please let us know if you are running into specific issues related to this.

mud5150:
@jchengsfx Ah! I wasn't looking closely enough and used my jump to conclusions mat when I experienced download issues. Sorry about that. Since your eyes are on this, I just created a new issue #1534 Perhaps you can help me with this?

Issue #1534, opened 2020-11-17
Puppet module doesn't appear to be idempotent on windows


mud5150:
I don't see anything that would stop the service from restarting, and reporting changed resources, every time the puppet agent runs on windows.

jchengsfx:
@mud5150 Thanks for the feedback.  The changes from #1536 have been released to puppet forge in v1.1.5.  Please let us know if it addresses this issue.

PR #1536, opened 2020-11-19
puppet: prevent unnecessary restarts/reinstalls on windows

jchengsfx:
https://github.com/signalfx/signalfx-agent/issues/1534


tests/deployments/puppet/puppet_test.py
```diff
@@ -263,6 +263,13 @@ def test_puppet_on_windows(puppet_version):
             assert wait_for(
                 p(has_datapoint_with_metric_name, backend, "sfxagent.datapoints_sent")
             ), "Didn't get internal metric datapoints"
+
+            # re-apply without any changes
+            run_win_puppet_agent(backend, monitors, INITIAL_VERSION, STAGE)
+            backend.reset_datapoints()
+            assert wait_for(
```
rmfitzpatrick, line 270: Would this have failed without the updates to the manifest?

jchengsfx: It shouldn't have failed. I just wanted to add another scenario to the test to make sure that everything is still running after applying without changes.

rmfitzpatrick (approved):

PR #1545, opened 2020-12-03
Adding the memory reporting for Windows containers.

vishalrp:
Windows containers don't report the same metrics as Linux containers. 

Tested this code change to report CommitPeak and PrivateWorkingSet. 
See: https://github.com/docker/engine/blob/master/api/types/stats.go

image

### DIFF
--- a/pkg/monitors/docker/conversion.go
+++ b/pkg/monitors/docker/conversion.go
@@ -149,18 +149,13 @@ func convertMemoryStats(stats *dtypes.MemoryStats, enhancedMetrics bool) []*data
 	// If not present, default value will be 0.
 	bufferCacheUsage := stats.Stats["total_cache"]
 
+	out = append(out, sfxclient.Gauge("memory.usage.limit", nil, int64(stats.Limit)))
 	if stats.PrivateWorkingSet == 0 {
-		out = append(out, []*datapoint.Datapoint{
-			sfxclient.Gauge("memory.usage.limit", nil, int64(stats.Limit)),
-			// See discussion at https://github.com/signalfx/signalfx-agent/issues/1009
-			sfxclient.Gauge("memory.usage.total", nil, int64(stats.Usage-bufferCacheUsage)),
-		}...)
+		// See discussion at https://github.com/signalfx/signalfx-agent/issues/1009
+		out = append(out, sfxclient.Gauge("memory.usage.total", nil, int64(stats.Usage-bufferCacheUsage)))
 	} else {
 		// This is used for Windows containers
-		out = append(out, []*datapoint.Datapoint{
-			sfxclient.Gauge("memory.usage.limit", nil, int64(stats.CommitPeak)),
-			sfxclient.Gauge("memory.usage.total", nil, int64(stats.PrivateWorkingSet)),
-		}...)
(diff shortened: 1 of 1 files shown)


dloucasfx:
@vishalrp thanks for submitting the PR, please `gofmt -s` the change and update the PR

dloucasfx:
LGTM 👍 

@keitwb the `PrivateWorkingSet` is only set for windows https://github.com/moby/moby/blob/master/daemon/daemon_windows.go#L571

pkg/monitors/docker/conversion.go
```diff
@@ -149,11 +149,19 @@ func convertMemoryStats(stats *dtypes.MemoryStats, enhancedMetrics bool) []*data
...
+	if stats.PrivateWorkingSet == 0 {
+		out = append(out, []*datapoint.Datapoint{
+			sfxclient.Gauge("memory.usage.limit", nil, int64(stats.Limit)),
+			// See discussion at https://github.com/signalfx/signalfx-agent/issues/1009
+			sfxclient.Gauge("memory.usage.total", nil, int64(stats.Usage-bufferCacheUsage)),
+		}...)
+	} else {
+		// This is used for Windows containers
+		out = append(out, []*datapoint.Datapoint{
+			sfxclient.Gauge("memory.usage.limit", nil, int64(stats.CommitPeak)),
```
keitwb, line 161: Do you have any references or doc sources that verify these values are the memory limit and total usage?  I'm not familiar enough with Windows memory to know.

keitwb: I read the article at https://technet.microsoft.com/en-us/magazine/ff382715.aspx and the memory usage total metric makes sense but the Commit Peak == the usage limit isn't clear to me from that.

vishalrp: To be honest, I looked for documentation as well and there wasn't anything concrete. I made the determination that that is the correct value based on the output of `docker stats`. 

This maybe relevant: https://github.com/moby/moby/issues/36359

Without that value the charts on signalfx Docker Host dashboard don't work. If CommitPeak is a problem we could revert that and only report the private working set as total usage since we are using custom dashboards for Windows containers anyway.

dloucasfx:
the code where the windows "hcs shim" is setting this value is a closed source and with the lack of documentation, it's hard to tell what this peak value really means.
@keitwb I wonder if we should simply introduce new metrics with for the windows memory struct and update the content or introduce a new chart

pkg/monitors/docker/conversion.go
keitwb: Try changing the memory limit on the container when you start them and see if it gets reflected in CommitPeak.  I don't understand why they wouldn't just reuse the same struct field for memory limit since that is a Docker thing.

vishalrp: I did try that, and doesn’t look like the memory limit is being used it seems to stay at the 1gb default.

keitwb: Ok, I recommend taking out the changes in this PR for the limit and keeping the changes for the usage.  Then I will merge that and at least we can get part of it covered.

vishalrp: Updated the code.

keitwb (approved):

PR #1546, opened 2020-12-07
Add monitor level interval configuration for collectd jmx

bjsignalfx:


keitwb (commented):
What is the point of this?  They will automatically get the interval set on the top level of collectd automatically already right?

bjsignalfx:
> What is the point of this? They will automatically get the interval set on the top level of collectd automatically already right?

Yes, you are right. Please have look at the latest commit.

pkg/monitors/collectd/monitorcore.go
```diff
@@ -80,7 +80,7 @@ func (bm *MonitorCore) SetConfigurationAndRun(conf config.MonitorCustomConfig) e
 // SetConfiguration adds various fields from the config to the template context
 // but does not render the config.
 func (bm *MonitorCore) SetConfiguration(conf config.MonitorCustomConfig) error {
-	return bm.collectdInstance().ConfigureFromMonitor(bm.monitorID, bm.Output, bm.UsesGenericJMX)
+	return bm.collectdInstance().ConfigureFromMonitor(bm.monitorID, bm.Output, bm.UsesGenericJMX, conf.MonitorConfigCore().IntervalSeconds)
```
keitwb, line 83: This will cause probably the last monitor that uses generic jmx to have its interval configured.  That is a confusing behavior and it would be better to figure out a way to set  an interval for each monitor separately.

bjsignalfx: I don't know how we can resolve this short of having multiple collectd processes. So you will need a collectd process for every java plugins configured with a different interval.

emaderer:
Closing this since it is stale.

PR #1553, opened 2020-12-12
nagios monitor with event for context

xp-1000:
Hello,

this pr aims to bring ability to reuse old nagios scripts.

this is already possible thanks to `telegraf/exec` but as explained in this monitor doc it lacks of context (we only know if the check is OK or not but not why).

this is why we use event to bring more context like stdout or errors from the command execution.

image


docs/monitors/nagios.md
```diff
@@ -0,0 +1,94 @@
...
+It is very similar to [telegraf/exec](https://docs.signalfx.com/en/latest/integrations/agent/monitors/telegraf-exec.html) 
+monitor configured with `dataFormat: nagios` but:
+  - it does not retrieve perfdata metrics, only the state of the script for alerting purpose.
+  - it will override the state if exit code == 0 but output string starts with `warn`, `crit` or `unkn` 
+  (not case sensitive).
+
+Also the main advantage and purpose of this monitor is to add more context to this status check state 
+thougth SignalFx [events](https://docs.signalfx.com/en/latest/detect-alert/events-intro.html).
+Indeed, in addition to the state metric, it will send an event which includes the output and the error 
+catched from the command execution.
```
keitwb, line 29:
```suggestion
caught from the command execution.
```

docs/monitors/nagios.md
```diff
@@ -0,0 +1,94 @@
...
+
+This should make the troubleshooting more efficient and allow the user to remain in SignalFx withtout
```
keitwb, line 31:
```suggestion
This should make the troubleshooting more efficient and allow the user to remain in SignalFx without 
```

docs/monitors/nagios.md
```diff
@@ -0,0 +1,94 @@
...
+  - it will override the state if exit code == 0 but output string starts with `warn`, `crit` or `unkn` 
+  (not case sensitive).
+
+Also the main advantage and purpose of this monitor is to add more context to this status check state 
+thougth SignalFx [events](https://docs.signalfx.com/en/latest/detect-alert/events-intro.html).
+Indeed, in addition to the state metric, it will send an event which includes the output and the error 
+catched from the command execution.
+
+This should make the troubleshooting more efficient and allow the user to remain in SignalFx withtout 
+to have to connect to the machine in case of anormal state to understand what is happen.
```
keitwb, line 32:
```suggestion
to have to connect to the machine in case of anormal state to understand what is happening.
```

docs/monitors/nagios.md
```diff
@@ -0,0 +1,94 @@
...
+It will also give the ability to create a dashboard similar to what nagios user are used to have.
```
keitwb, line 33:
```suggestion
It will also give the ability to create a dashboard similar to what nagios user are accustomed to.
```

docs/monitors/nagios.md
```diff
@@ -0,0 +1,94 @@
...
+
+**For a list of monitor options that are common to all monitors, see [Common
+Configuration](../monitor-config.md#common-configuration).**
+
+
+| Config option | Required | Type | Description |
+| --- | --- | --- | --- |
+| `command` | **yes** | `string` | The command to exec with any arguments like: `"LC_ALL=\"en_US.utf8\" /usr/lib/nagios/plugins/check_ntp_time -H pool.ntp.typhon.net -w 0.5 -c 1"` |
+| `service` | **yes** | `string` | Corresponds to the nagios `service` column and allows to aggregate all instances of the same service (when calling the same check script with different arguments) |
+| `timeout` | no | `integer` | The max execution time allowed in seconds be fore to send SIGTERM. In any case, a SIGKILL will be send at `intervalSeconds` value. (**default:** `9`) |
```
keitwb, line 59:
```suggestion
| `timeout` | no | `integer` | The max execution time allowed in seconds before sending SIGTERM. In any case, a SIGKILL will be sent at `intervalSeconds` value. (**default:** `9`) |
```

pkg/monitors/nagios/nagios.go
```diff
@@ -0,0 +1,253 @@
...
+			datapoint.New(
+				nagiosState,
+				dimensions,
+				datapoint.NewIntValue(int64(state)),
+				datapoint.Gauge,
+				time.Time{}),
+		}...)
+
+		properties, diffProperties := makeProperties(state, err, stdout, stderr)
+		m.logger.Warn(properties)
```
keitwb, line 91: Remove these or make Debug levels.

pkg/monitors/nagios/nagios.go
```diff
@@ -0,0 +1,253 @@
...
+		m.logger.Warn(diffProperties)
+		// Compare with previous event if it exists
+		sendEvent := true
+		if x, found := c.Get(cacheKey); found {
+			lastProperties := x.(map[string]interface{})
+			if diff := deep.Equal(diffProperties, lastProperties); diff == nil {
+				m.logger.Debug("a same avent has already been send, do not send again")
```
keitwb, line 98:
```suggestion
				m.logger.Debug("the same event has already been sent, do not send again")
```

pkg/monitors/nagios/nagios.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// After user timeout, try to gracefully stop the process with sigterm
+	termTimer := time.AfterFunc(time.Duration(termTimeout)*time.Second, func() {
+		err = cmd.Process.Signal(syscall.SIGTERM)
+		if err != nil {
+			return
+		}
+		// Kill the process to respect the monitor intervalSeconds
+		killTimer = time.AfterFunc(time.Duration(killTimeout)*time.Second, func() {
+			err = cmd.Process.Signal(syscall.SIGTERM)
```
keitwb, line 157: Is this supposed to be SIGKILL here?

pkg/monitors/nagios/nagios.go
```diff
@@ -0,0 +1,253 @@
...
+			if err != nil {
+				return
+			}
+		})
+	})
+	err = cmd.Wait()
+	// Shutdown timers
+	if killTimer != nil {
```
keitwb, line 165: This doesn't seem thread-safe since killTimer is created in a  separate goroutine.

pkg/monitors/nagios/nagios.go
```diff
@@ -0,0 +1,253 @@
...
+	// executed to its end even if the exitcode still can be an error in
+	// the script (not the nagios state). In any case we return the exit
+	// code as state and the user could find more information in the event
+	return ws.ExitStatus(), nil
+}
+
+func makeProperties(state int, err error, stdout []byte, stderr []byte) (map[string]interface{}, map[string]interface{}) {
+	properties := make(map[string]interface{})
+	if len(stdout) > 0 {
+		properties["std_out"] = string(stdout[:propertiesLength])
```
keitwb, line 231:
```suggestion
		properties["stdout"] = string(stdout[:propertiesLength])
```

pkg/monitors/nagios/nagios.go
```diff
@@ -0,0 +1,253 @@
...
+		properties["exit_reason"] = err.Error()
+	} else {
+		properties["exit_code"] = state
+	}
+
+	// Some scripts could produce different output (and stderr) for
+	// each interval "normally", so we do not want to compare them
+	diffProperties := make(map[string]interface{})
+	for k, v := range properties {
+		if !strings.HasPrefix(k, "std_") {
```
keitwb, line 246:
```suggestion
		if !strings.HasPrefix(k, "std") {
```

pkg/monitors/nagios/nagios.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	// An error with the exit code so we can be sure the command has been
+	// executed to its end even if the exitcode still can be an error in
+	// the script (not the nagios state). In any case we return the exit
+	// code as state and the user could find more information in the event
+	return ws.ExitStatus(), nil
+}
+
+func makeProperties(state int, err error, stdout []byte, stderr []byte) (map[string]interface{}, map[string]interface{}) {
```
keitwb, line 228: I would just make this function return the map of all of the properties and then have another function that filters out what you are omitting from `diffProperties`.  That will make it a bit clearer to read this in the main monitor method that calls this.

pkg/monitors/nagios/nagios.go
xp-1000: sorry it was a debug test I forgot in the last commit ..

pkg/monitors/nagios/nagios.go
xp-1000: good catch!

pkg/monitors/nagios/nagios.go
xp-1000: you are right

pkg/monitors/nagios/nagios.go
xp-1000: is it better now ? else please could you advice ? thanks

keitwb (approved):
Yes looks much better with the `select`.  Thanks for submitting this.

PR #1555, opened 2020-12-14
Update k8s and other deps

rmfitzpatrick:
These changes attempt to add compatibility with shared Splunk OTel Collector dependencies for usage in that library.

1. Move `k8s.io/*` libs from 0.17.0 to 0.19.4
2. Move `hashicorp/consul` 1.4.0 to `hashicorp/consul/api` 1.7.0
3. Move `hashicorp/vault` from `v1.1.1-0.20190321125746-66ef59957aaf` to 1.6.0
4. Adopt `signalfx/ingest-protocols` v0.0.16

In order to comply with these updates a few signature changes needed to be adopted and some conflicting dependencies required some workarounds:

1. Per @keitwb's request I'm vendoring in an updated `soheilhy/cmux` version that adopts go modules to remove [a breaking test dependency](https://github.com/soheilhy/cmux/pull/81).
2. I've added a replace directive to enforce `google.golang.org/grpc v1.29.1` since etcd's test dependencies cap at this version before a [breaking package removal occurred](https://github.com/grpc/grpc-go/releases/tag/v1.30.0).  When we want to move past this version, upstream or a forked version of etcd will need updates to no longer rely on `grpc/naming`.

~~status: Investigating potential CI fallout.~~
status: I believe I've made all the necessary changes, but will investigate the apparent perf test issues.
  --  edit: I missed adding the `thirdparty` directory to the `fakek8s` test service so these should be ok.


go.mod
```diff
@@ -2,185 +2,110 @@ module github.com/signalfx/signalfx-agent
...
-
-replace github.com/dancannon/gorethink => gopkg.in/gorethink/gorethink.v4 v4.0.0
-
-replace github.com/signalfx/signalfx-agent/pkg/apm => ./pkg/apm
+replace (
+	code.cloudfoundry.org/go-loggregator => github.com/signalfx/go-loggregator v1.0.1-0.20200205155641-5ba5ca92118d
+	github.com/dancannon/gorethink => gopkg.in/gorethink/gorethink.v4 v4.0.0
+	github.com/influxdata/telegraf => github.com/signalfx/telegraf v0.10.2-0.20201211214327-200738592ced
+	github.com/signalfx/signalfx-agent/pkg/apm => ./pkg/apm
+	github.com/soheilhy/cmux => github.com/rmfitzpatrick/cmux v0.1.4 // required to drop google.golang.org/grpc/examples/helloworld/helloworld test dep
```
keitwb, line 10: Take this repo and put it into a new directory in this repo called `thirdparty/cmux` and make the replace reference that dir instead of your personal fork.  Would rather not rely on personal forks.

rmfitzpatrick: What about forking in the `signalfx` project to avoid taking on vendored indirect deps?

keitwb: That would be better than a personal fork but putting it here in the repo avoids cluttering our github org with random packages.  It looks like a fairly small number of files so wouldn't be a huge burden here.  Whichever you prefer.

rmfitzpatrick: Updated with your initial request since go modules will let other libraries' replace directives use the new path and a standalone repo isn't necessary.

keitwb (approved):
Looks good.  go.mod has a conflict now from a PR I just merged but otherwise looks like tests are all good now.

PR #1556, opened 2020-12-15
nagios: allow to filter stdout and stderr from event comparison

xp-1000:
Hello 

related to https://github.com/signalfx/signalfx-agent/pull/1553 

it seems users want this to be customizable because there are as much scripts with different and same output from nagios world.

thanks for the quick merge of the previous one it will be enough to unstuck my colleague.

### DIFF
--- a/pkg/monitors/nagios/nagios.go
+++ b/pkg/monitors/nagios/nagios.go
@@ -38,6 +38,12 @@ type Config struct {
 	Service string `yaml:"service" validate:"required"`
 	// The max execution time allowed in seconds before sending SIGKILL
 	Timeout int `yaml:"timeout" default:"9"`
+	// If `false` and change is detected on `stdout` compared to the last
+	// event it will send a new one
+	FilterStdOut bool `yaml:"filterStdOut" default:"false"`
+	// If `false` and change is detected on `stderr` compared to the last
+	// event it will send a new one
+	FilterStdErr bool `yaml:"filterStdErr" default:"false"`
 }
 
 // Monitor that collect metrics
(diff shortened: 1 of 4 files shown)


pkg/monitors/nagios/nagios.go
```diff
@@ -38,6 +38,12 @@ type Config struct {
 	Service string `yaml:"service" validate:"required"`
 	// The max execution time allowed in seconds before sending SIGKILL
 	Timeout int `yaml:"timeout" default:"9"`
+	// If `false` and change is detected on `stdout` compared to the last
+	// event it will send a new one
+	FilterStdOut bool `yaml:"filterStdOut" default:"false"`
```
keitwb, line 43: I would call this `IgnoreStdOut`, etc.  "Filter" is not very clear.

pkg/monitors/nagios/nagios.go
```diff
@@ -95,7 +101,10 @@ func (m *Monitor) Configure(conf *Config) error {
 		properties := makeProperties(state, err, stdout, stderr)
 		// Some scripts could produce different output (and stderr) for
 		// each interval "normally", so we do not want to compare them
-		diffProperties := filterProperties(properties)
+		m.logger.Warn(conf.FilterStdOut)
```
keitwb, line 104: Remove these two log lines.

xp-1000: again, sorry

pkg/monitors/nagios/nagios.go
xp-1000: done

keitwb (approved):

PR #1557, opened 2020-12-16
Update tracker.go to add pod name as a tag

harnitsignalfx:
It is valuable to the customers to track the pod name in the span tag. Did some testing and confirmed that this worked as the kubernetes_pod_name seems to be available in the map of dims that the agent checks against (dims: map[container_id,container_image,container_name,container_spec_name:,kubernetes_namespace, kubernetes_pod_name, kubernetes_pod_uid]
Though there may be other implications that I'm not aware of (the MTS with these dimensions also get updated with properties?)

### DIFF
--- a/pkg/apm/tracetracker/tracker.go
+++ b/pkg/apm/tracetracker/tracker.go
@@ -21,6 +21,7 @@ const spanCorrelationMetricName = "sf.int.service.heartbeat"
 var DefaultDimsToSyncSource = map[string]string{
 	"container_id":       "container_id",
 	"kubernetes_pod_uid": "kubernetes_pod_uid",
+	"kubernetes_pod_name": "kubernetes_pod_name",
 }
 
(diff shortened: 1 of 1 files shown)


keitwb:
It's the _other implications_ that I'm a bit wary of.  @jrcamp since you've dealt with this stuff recently can you think of the implications of this?

jrcamp (changes requested):
This field is currently used in 2 different ways. One is to enrich spans (what you're trying to do it seems) and the other is for correlation. We don't want the dimensions we correlate against to change so this will need some minor refactoring so correlation uses the existing list and a new list can be made for span enrichment.

harnitsignalfx:
Ack, I'll try to propose that change! Thank you both!

harnitsignalfx:
Closing in lieu of https://github.com/signalfx/signalfx-agent/pull/1559

PR #1558, opened 2020-12-17
Add support to URL Query

dloucasfx:
If path includes a query, then the `normalizeURL` unwillingly strips the query part
example:
with this config, the monitor ends up hitting the endpoint `/Tools-and-Home-Improvement/b/` instead the full path
```
  - type: http
    host: amazon.com
    useHTTPS: true
    path: '/Tools-and-Home-Improvement/b/?ie=UTF8&node=228013&ref_=nav_cs_hi'
    method: GET
```

### DIFF
--- a/pkg/monitors/http/http.go
+++ b/pkg/monitors/http/http.go
@@ -139,7 +139,11 @@ func (m *Monitor) Configure(conf *Config) (err error) {
 			}
 
 			if m.conf.AddRedirectURL && !m.conf.NoRedirects {
-				normalizedURL, _ := m.normalizeURL(fmt.Sprintf("%s://%s:%s%s", redirectURL.Scheme, redirectURL.Hostname(), redirectURL.Port(), redirectURL.Path))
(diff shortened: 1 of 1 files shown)


keitwb (changes requested):
Do the same thing at https://github.com/signalfx/signalfx-agent/blob/9a7160accab22b51c11756177259f5fedf75dc90/pkg/monitors/http/http.go#L142

keitwb:
Also @xp-1000,  heads up in case you have any  input on this.

dloucasfx:
> Do the same thing at
> 
> https://github.com/signalfx/signalfx-agent/blob/9a7160accab22b51c11756177259f5fedf75dc90/pkg/monitors/http/http.go#L142

done

keitwb (approved):

xp-1000:
> Also @xp-1000, heads up in case you have any input on this.

thanks for the ping and the improvement, it seems good for me.

PR #1559, opened 2020-12-18
added separate map for default dimensions to add onto spans

harnitsignalfx:
Am I on the right track? Sorry for the crazy number of PRs, I'm closing the previous one. Ideally I'd also like to add kubernetes_node_name onto the list, but thats a separate conversation and has 2 issues that I can think of -  
a) redundant with host so not sure if that is reasonable.
b) not available in the default dimensions on the endpoint, I'd have to pass in the hostIDDims map from the writer to do it

### DIFF
--- a/pkg/core/writer/tracetracker/host.go
+++ b/pkg/core/writer/tracetracker/host.go
@@ -9,7 +9,6 @@ import (
 	"github.com/signalfx/golib/v3/trace"
 	"github.com/sirupsen/logrus"
 
-	apmtracker "github.com/signalfx/signalfx-agent/pkg/apm/tracetracker"
 	"github.com/signalfx/signalfx-agent/pkg/core/common/constants"
 	"github.com/signalfx/signalfx-agent/pkg/core/services"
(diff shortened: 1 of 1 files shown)


pkg/apm/tracetracker/tracker.go
```diff
@@ -23,6 +23,14 @@ var DefaultDimsToSyncSource = map[string]string{
 	"kubernetes_pod_uid": "kubernetes_pod_uid",
 }
 
+// DefaultDimsToAddToSpans are the default dimensions to add as span tags for correlated environments
+var DefaultDimsToAddToSpans = map[string]string{
```
keitwb, line 27: Just put this in the `tracetracker` package since it isn't used at all for APM dimension update stuff.

keitwb (commented):
This seems like a better solution other than the location of the map.  Fix that and I'll approve it.

jrcamp:
lgtm, just what Ben said.

harnitsignalfx:
Ty! Fixed, please take a look.

jrcamp (approved):

PR #1573, opened 2021-01-12
Add docker support for windows

dloucasfx:
Wrote the conditions in the template to turn off unnecessary mounts when docker is disabled (in case of CRI-O).

 cc: @keitwb 


keitwb (approved):

keitwb:
I think you need to regenerate the non-helm resources with the `generate-from-helm` script.

PR #1574, opened 2021-01-13
Add a shutdown hook to processlist monitor

dloucasfx:
While debugging a windows performance issue related to the processlist usage of wmi, I noticed that if I just edit the config to remove `processlist` monitor, the monitor is removed from the internal list but the main function loaded in `RunOnInterval` periodically runs and hits the wmi interface.

The only workaround is to restart the agent.

this bug adds the shutdown hook to processlist monitor


dloucasfx:
cc: @keitwb @jrcamp

keitwb (approved):
Thanks,  good catch.

PR #1575, opened 2021-01-14
Add inline yaml struct tags for all monitor configs

rmfitzpatrick:
These changes allow each monitor config to be unmarshalled directly from yaml instead of requiring the current two pass system using the catch-all extra config.  They are effectively noops regarding current unmarshalling behavior and are being added for better compatibility for agent module* consumers.


keitwb (approved):
Seems reasonable.

PR #1579, opened 2021-01-21
aws instance-identity response; log the body for better debugging

dloucasfx:
This will help debugging errors like this:

```
time="2021-01-21T06:42:31Z" level=debug msg="Failed to unmarshal AWS instance-identity response" error="invalid character '<' looking for beginning of value"
```

### DIFF
--- a/pkg/core/hostid/aws.go
+++ b/pkg/core/hostid/aws.go
@@ -44,6 +44,7 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 	if err != nil {
 		log.WithFields(log.Fields{
 			"error": err,
+			"body":  string(body),
 		}).Debug("Failed to unmarshal AWS instance-identity response")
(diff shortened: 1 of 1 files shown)


pkg/core/hostid/aws.go
```diff
@@ -44,7 +44,7 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 	if err != nil {
 		log.WithFields(log.Fields{
 			"error": err,
-		}).Debug("Failed to unmarshal AWS instance-identity response")
+		}).Debugf("Failed to unmarshal AWS instance-identity response. Body: %s", body)
```
jrcamp, line 47: I would add a field "body" that has it instead of adding it to the message itself.

dloucasfx: done

jrcamp (approved):

PR #1580, opened 2021-01-22
Improve windows performance by fetching all processes name in one call

dloucasfx:
The host observer usage of the `gopsutil -> process` method `Name()`, for windows, is not efficient as the latter is a super expensive call.
The host observer gets the list of Processes IDs on the system, iterates through the list and calls `Name()` for every PID that is listening on a port.

[Host observer ](https://github.com/signalfx/signalfx-agent/blob/master/pkg/observers/host/host.go#L122-L133)

The [Name() ](https://github.com/shirou/gopsutil/blob/master/process/process_windows.go#L638) method creates a snapshot of all the running processes and iterates through the list to find a PID that matches the caller’s PID and returns the process name.
`CreateToolhelp32Snapshot` and `Process32Next` are the main resource intensive methods; 
the caller invokes this method for every PID on the system!

As a solution, our code does not need to make the call for each PID, instead, we should use a method that iterates through the windows processes snapshot, builds and returns a PID->ProcessName map of all the processes on the system.

Note: Telegraf/procstat monitor suffer from the same issue.
https://github.com/signalfx/telegraf/blob/signalfx-integration/plugins/inputs/procstat/native_finder_windows.go#L35-L46

cc: @jrcamp @keitwb 

### DIFF
--- a/pkg/observers/host/host_windows.go
+++ b/pkg/observers/host/host_windows.go
@@ -1,0 +1,44 @@
+// +build windows
+
+package host
+
+import (
+	"fmt"
+	"unsafe"
+
+	"github.com/shirou/gopsutil/process"
+	"golang.org/x/sys/windows"
+)
+
+func (p *processName) getName(proc *process.Process) (string, error) {
+	name, ok := p.pidNameMap[proc.Pid]
+	if !ok {
+		return "", fmt.Errorf("could not find name for PID %v", proc.Pid)
+	}
+	return name, nil
+}
+
+// setPidNameMap fills up the pidNameMap with all processes running on
+// the system by iterating through the windows process snapshot
+func (p *processName) setPidNameMap() error {
+	snapshot, err := windows.CreateToolhelp32Snapshot(windows.TH32CS_SNAPPROCESS, 0)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/observers/host/host_windows.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+	"unsafe"
+
+	"github.com/shirou/gopsutil/process"
+	"golang.org/x/sys/windows"
+)
+
+func (p processName) getName(proc *process.Process) (string, error) {
```
jrcamp, line 13: While not strictly necessary since the map in the struct is a pointer itself, maybe make these pointer receivers just for principle of least surprise for somebody down the road?

jrcamp (commented):
lgtm

pkg/observers/host/host_windows.go
dloucasfx: agree 👍

dloucasfx: done

jrcamp (approved):

keitwb (approved):

PR #1582, opened 2021-01-27
Allow k8s-api observer's discovery settings to be independently configured via Helm

KingJ:
When using the Helm chart, it is not currently possible to set the k8s-api observer's `discoverAllPods` or `discoverNodes` parameters except by setting the `isServerless` parameter to true. There are some situations where the observer needs to be configured to discover all pods or nodes in a non-serverless environment, thus this change will set the values to true if the new `discoverAllPods`/`discoverNodes` setting is true or if `isServerless` is set to true.

With the current default values, this does result in a minor change to the ConfigMap - removing `discoverAllPods` and `discoverNodes` if they are not set to true. This does not result in a behavioural change, [as the observer itself will treat these values as false if not explicitly defined](https://docs.signalfx.com/en/latest/integrations/agent/observers/k8s-api.html#configuration).

One such use case for this functionality on a non-serverless cluster is for configuring monitors that target services running on some of the cluster's nodes that are not discoverable via Kubernetes itself. For example, [a similar discovery rule as is used to discover the Kubelet on Fargate clusters](https://github.com/signalfx/signalfx-agent/blob/fc1d6cbdfe042855b1ad3e6f55416cd027f86955/deployments/k8s/helm/signalfx-agent/templates/configmap.yaml#L154) can be used to discover etcd on Rancher clusters. Both rely on the ability to discover the nodes themselves.

### DIFF
--- a/deployments/k8s/configmap.yaml
+++ b/deployments/k8s/configmap.yaml
@@ -35,8 +35,6 @@ data:
 
     observers:
     - type: k8s-api
-      discoverAllPods: false
-      discoverNodes: false
 
     monitors:
     - type: cpu
--- a/deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
@@ -96,8 +96,12 @@ data:
         skipVerify: true
       {{- end }}
       {{- if semverCompare ">=5.2.0" $agentSemver.Original }}
-      discoverAllPods: {{ toYaml .Values.isServerless }}
-      discoverNodes: {{ toYaml .Values.isServerless }}
+      {{- if or (.Values.isServerless) (.Values.discoverAllPods) }}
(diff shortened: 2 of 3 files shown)


keitwb (approved):
Seems reasonable.  Thanks for submitting this.

KingJ:
Thank you for the quick approval! I've had a quick look at the failed integration test, and I don't think it is related to my change - I also can't trigger a retry however. 

I have also realised that I have not updated the Helm chart version as part of my PR - should I update the version in this PR, or will the maintainers do so at a later date?

keitwb:
Yeah, integration test is unrelated.  One of us will upgrade the version when we do a release, you needn't worry about it. Thanks.

KingJ:
> Yeah, integration test is unrelated. One of us will upgrade the version when we do a release, you needn't worry about it. Thanks.

Awesome, thank you for clarifying!

PR #1585, opened 2021-01-28
Allow for additional container image exclusions via Helm values

KingJ:
The kubelet-metrics, kubelet-stats and docker-container-stats monitors exclude some or all metrics from two common 'pause' containers by default. However, some Kubernetes distributions will use different pause container images and thus the list of container images to exclude from these monitors needs to be extensible.

n.b. The indent levels for the included list are set to an odd number in order to match the same indent level as the existing items in the image list. This is also consistent with the excluded image list indentation in the [kubelet-stats documentation](https://github.com/signalfx/signalfx-agent/blob/master/docs/monitors/kubelet-stats.md#L43-L44).


keitwb (approved):

Issue #1588, opened 2021-01-29
Incorrect usage of package causes constant reporting


pearcec:
https://github.com/signalfx/signalfx-agent/blob/89ea2abe91b1ed64fcdd295e2fdfbeaafa6db2cd/deployments/puppet/manifests/init.pp#L36

The use of package in this context is not installing the package.  It is simply installing the service.  There IS a service resource type, or you could use exec.  Using the package in this way constant runs this install.  Why? Because you executable doesn't actually install as a windows package.  So when puppet looks for an installed package it sees it as absent and run this command over and over again.

Notice: /Stage[main]/Signalfx_agent/Package[signalfx-agent]/ensure: created

If you moved to an installed package this might solve a lot of pain. Including this issue

https://github.com/signalfx/signalfx-agent/issues/1584

jchengsfx:
@pearcec v1.2.0 has been released to puppet forge.  Please let us know if the changes address this issue.

atoulme:
Closing this issue as inactive. Please reopen if this issue is still occurring.

Issue #1584, opened 2021-01-28
Windows install still not idempotent


pearcec:
https://github.com/signalfx/signalfx-agent/blob/e2ff35fa932c376ce8ed215cf707ec7bda48d15c/deployments/puppet/manifests/win_repo.pp#L19

file and find_file execute on the server not the agent.

pearcec:
You need to write a fact.  Something like this.  Make sure you get the correct path and set a version and not a url in the version.txt.

```
Facter.add(:signalfx_agent_version) do
    confine :osfamily => "Windows"
    setcode 'type "C:\\Program Files\\SignalFx\\version.txt"'
  end
```

jchengsfx:
Thanks @pearcec.  Was thinking the same about adding the new fact.

jchengsfx:
@pearcec v1.2.0 has been released to puppet forge.  Please let us know if the changes address this issue.

atoulme:
Closing this issue as inactive. Please reopen if this issue is still occurring.

PR #1590, opened 2021-01-31
Add windows registry facts for puppet

jchengsfx:
For #1584 and #1588 
- Requires the `puppetlabs/registry` module to access/update the agent service registry key
- Remove `package` resource for windows; install agent service with `exec`

### DIFF
--- a/deployments/puppet/.fixtures.yml
+++ b/deployments/puppet/.fixtures.yml
@@ -12,4 +12,7 @@ fixtures:
     powershell:
       repo: "puppetlabs/powershell"
       ref: "2.3.0"
+    registry:
+      repo: "puppetlabs/registry"
+      ref: "3.2.0"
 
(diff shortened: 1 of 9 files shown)


keitwb (commented):
A few questions:

1) If somebody has the agent already installed, will this transition them to using the new registry values so that the second time they run it after updating the puppet module it will work properly?
2) Did you test this manually with different  scenarios?

jchengsfx:
> A few questions:
> 
> 1. If somebody has the agent already installed, will this transition them to using the new registry values so that the second time they run it after updating the puppet module it will work properly?
> 2. Did you test this manually with different  scenarios?

1. If the registry values do not exist, they will be created.  This applies for both previous and new installs.
2. Manually tested with upgrade from latest released module.

keitwb (approved):

PR #1591, opened 2021-01-31
cpu: if cpu.utilzation > 100, round down to 100

mstumpfx:
I'm not sure if this is the best solution, but this seems slightly more favorable in cases of rounding errors where CPU is something like 100.000003 but we don't send a datapoint. 
In the case of some ridiculous value, 100 would still be enough to trigger detectors and show the CPU pegged, with the logs showing the true value.

### DIFF
--- a/pkg/monitors/cpu/cpu.go
+++ b/pkg/monitors/cpu/cpu.go
@@ -188,7 +188,7 @@ func getUtilization(prev *totalUsed, current *totalUsed) (utilization float64, e
 			err = fmt.Errorf("percent %v < 0 total: %v used: %v", utilization, totalDiff, usedDiff)
 		}
 		if utilization > 100.0 {
-			err = fmt.Errorf("percent %v > 100 total: %v used: %v ", utilization, totalDiff, usedDiff)
+			utilization = 100
 		}
 	}
 


pkg/monitors/cpu/cpu.go
```diff
@@ -188,7 +188,8 @@ func getUtilization(prev *totalUsed, current *totalUsed) (utilization float64, e
 			err = fmt.Errorf("percent %v < 0 total: %v used: %v", utilization, usedDiff, totalDiff)
 		}
 		if utilization > 100.0 {
-			err = fmt.Errorf("percent %v > 100 total: %v used: %v ", utilization, usedDiff, totalDiff)
+			utilization = 100
+			m.logger.Warningf("actual percent %v > 100, total: %v, used: %v, sent value: 100", utilization, usedDiff, totalDiff)
```
jrcamp, line 192: Hmm I'm not sure if we want to log if this can happen just due to rounding errors and spams the logs. 🤔 idk

mstumpfx: Yeah, fair enough. I made a quick change and didn't realize this was a function instead of a method, so didn't have access to the logger anyway.

keitwb (approved):

PR #1595, opened 2021-02-04
memory Monitor: Add Swap Memory Metrics

### LINKED ISSUE #1594: What is the metric for monitor Virtual Memory in Windows  and Swap memory in Unix . We need utilization in %.
What is the metric for monitor Virtual Memory in Windows  and Swap memory in Unix . We need utilization in %.
Currently we are using "vmem" but it is reflecting data in bytes instead of %.

keitwb:
Adds swap total, used, and free for Windows, Linux, and Mac.

They are all non-default metrics.

Fixes #1594

### DIFF
--- a/pkg/monitors/memory/metadata.yaml
+++ b/pkg/monitors/memory/metadata.yaml
@@ -29,6 +29,18 @@ monitors:
       description: (Linux Only) Bytes of memory available for use.
       default: true
       type: gauge
+    memory.swap_free:
+      description: Bytes of swap memory available for use.
+      default: false
+      type: gauge
+    memory.swap_total:
+      description: Total bytes of swap memory on the system
+      default: false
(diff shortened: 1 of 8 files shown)


pkg/monitors/memory/metadata.yaml
```diff
@@ -29,6 +29,18 @@ monitors:
       description: (Linux Only) Bytes of memory available for use.
       default: true
       type: gauge
+    memory.swap_free:
+      description: Bytes of swap memory available for use.
+      default: false
+      type: gauge
+    memory.swap_total:
+      description: Total bytes of swap memory on the system
```
pmcollins, line 37:
```suggestion
      description: Total bytes of swap memory on the system.
```

pmcollins (approved):

PR #1598, opened 2021-02-04
vSphere: set `host` dimension to `vm_ip` value when available

pmcollins:
Previously, metrics for VMs in a vSphere cluster all had the same `host` dimension, the hostname on which the smart agent was running. This change explicitly sets the host dimension to the `vm_ip` dimension's value, if available.


keitwb (approved):
The helm issue was resolved by @jchengsfx a few days ago, you just needed to rebase with master and it fixed it.

PR #1600, opened 2021-02-09
Add config override to observers in K8S Helm chart

chien-splunk:
I have a use case where I want to run a monitor only agent in serverless mode with a single k8s pod.

This PR allows ability to override observers config block completely instead of assuming k8s-api discovery is always enabled.

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
@@ -90,6 +90,7 @@ data:
       {{- end }}
 
     observers:
+    {{- if .Values.configureStandardObservers }}
     - type: k8s-api
       {{- if .Values.apiServerSkipVerify }}
       kubernetesAPI:
@@ -109,6 +110,10 @@ data:
       - {{ toYaml . | indent 8 | trim }}
       {{- end }}
       {{- end }}
+    {{- end }}
+    {{- with .Values.observers }}
+    {{ toYaml . }}
+    {{- end }}
 
     monitors:
 {{- if .Values.configureStandardMonitors }}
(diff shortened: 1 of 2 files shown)


deployments/k8s/helm/signalfx-agent/templates/configmap.yaml
```diff
@@ -109,6 +116,7 @@ data:
       - {{ toYaml . | indent 8 | trim }}
       {{- end }}
       {{- end }}
+{{- end }}
 
     monitors:
 {{- if .Values.configureStandardMonitors }}
```
keitwb, line 122: I would prefer if this followed the same pattern as we do for monitors: there would be a config option `configureStandardObservers` that defaults to true that you could turn off  to disable the normal observers, and then provide the `observers` config that would configure those observers, possibly in addition to the standard ones.  Does that make sense?

keitwb: This allows for the possibility that you want the standard observers but also want to add another observer without having to figure out manually what the standard observer config is.

chien-splunk: I can do that but one of the core issues with leaving the standard observers list is that `discoverAllPods` and `discoverNodes` are hardcoded to `true` if running in serverless mode.

Should I also change that to obey the value of those settings regardless of serverless mode?

keitwb: Yes, in your use case, you should just set `configureStandardObservers: false` and provide your own observer list.

keitwb (approved):
Thanks for adding this!

PR #1604, opened 2021-02-12
(during build) Upgrade pip before installing python packages

cetanu:
The `cryptography` crate contains a Rust dependency;
If pip is not up to date it will expect the rust toolchain to be installed on the machine.
Otherwise, if pip is up to date, it will install a binary wheel.

I discovered this while building a DEB package for ARM64

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 ARG GO_VERSION=1.14.2
+ARG PIP_VERSION=21.0.1
 
 ###### Agent Build Image ########
 FROM ubuntu:16.04 as agent-builder
@@ -255,7 +256,9 @@ RUN patchelf --add-needed libm-2.23.so /opt/deps/libvarnishapi.so.1.0.4
 ###### Python Plugin Image ######
 FROM collectd as python-plugins
 
-RUN python3 -m pip install yq &&\
+ARG PIP_VERSION
+
+RUN python3 -m pip install --upgrade pip==$PIP_VERSION && python3 -m pip install yq &&\
     wget -O /usr/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 &&\
     chmod +x /usr/bin/jq
 


Dockerfile
```diff
@@ -255,7 +255,7 @@ RUN patchelf --add-needed libm-2.23.so /opt/deps/libvarnishapi.so.1.0.4
 ###### Python Plugin Image ######
 FROM collectd as python-plugins
 
-RUN python3 -m pip install yq &&\
+RUN python3 -m pip install --upgrade pip && python3 -m pip install yq &&\
```
jchengsfx, line 258: It may be best to pin the pip version, e.g. `pip install --upgrade pip==21.0.1`, to avoid any issues that could be introduced with future pip versions.

cetanu: Good idea. I added the suggested version as a Dockerfile ARG, alongside a similar one for golang at the top of the file.

cetanu:
As a side question: where do these ARM binaries normally get published? I was planning on compiling it myself

keitwb:
@jchengsfx looks like the CI tests aren't working with the new version, do you know why?

keitwb:
@cetanu we don't publish ARM mainly because we don't really have any way of testing on ARM with our existing CI frameworks so we are a bit wary of making it an official distribution.

Dockerfile
```diff
@@ -255,7 +256,7 @@ RUN patchelf --add-needed libm-2.23.so /opt/deps/libvarnishapi.so.1.0.4
 ###### Python Plugin Image ######
 FROM collectd as python-plugins
 
-RUN python3 -m pip install yq &&\
+RUN python3 -m pip install --upgrade pip==$PIP_VERSION && python3 -m pip install yq &&\
```
jchengsfx, line 259: Need to also add `ARG PIP_VERSION` before this line for multi-stage builds.

jchengsfx (changes requested):

Dockerfile
cetanu: Thanks for spotting that 👍 didn't know it worked that way

jchengsfx (approved):

PR #1607, opened 2021-02-16
Support postgresql 13

rmfitzpatrick:
These changes add support for Postgres 13, whose pg_stat_statements extension provided a breaking column name change of `total_time` -> `total_exec_time` to distinguish it from the added, optional `total_plan_time`.  This is done by querying the view's columns for the configured master db and formatting* the queries before the underlying SQL monitor Configure().

### DIFF
--- a/pkg/monitors/postgresql/monitor.go
+++ b/pkg/monitors/postgresql/monitor.go
@@ -136,7 +136,6 @@ func (m *Monitor) Configure(conf *Config) error {
 
 	m.serverMonitor, err = m.monitorServer()
 	if err != nil {
-		m.database.Close()
 		return fmt.Errorf("could not monitor postgresql server: %v", err)
 	}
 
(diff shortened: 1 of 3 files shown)


keitwb (changes requested):
The procedure that determines the column name needs to be resilient against the DB being down when the agent first starts up.

rmfitzpatrick:
> The procedure that determines the column name needs to be resilient against the DB being down when the agent first starts up.

@keitwb I've shifted the sql monitor creation to the main postgres monitor loop to allow retries.  This better?

keitwb (approved):
Yes, make sure the integration test is actually checking for the metric in question and it is good to go.

rmfitzpatrick:
> Yes, make sure the integration test is actually checking for the metric in question and it is good to go.

I've verified the added 13-alpine run fails without these changes.

PR #1608, opened 2021-02-18
Fix pip failures in ansible tests

jchengsfx:
Upgrade pip version in ansible tests to resolve test failures due to rust/cryptography dependencies.


keitwb (approved):
I trust that you have a good reason, but I'm curious why there are two different pip versions used?

jchengsfx:
> I trust that you have a good reason, but I'm curious why there are two different pip versions used?

Those are just the latest available versions for the python on each distro image.  I will update to use the same 20.3.4 version for consistency.

PR #1610, opened 2021-02-18
Add refresh cache logic to keep docker and sfx in sync

dloucasfx:
Add refresh cache logic to keep docker and sfx in sync when the event stream fails

cc: @keitwb @jrcamp 

### DIFF
--- a/pkg/core/common/docker/containerlist.go
+++ b/pkg/core/common/docker/containerlist.go
@@ -8,7 +8,6 @@ import (
 	dtypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
 	docker "github.com/docker/docker/client"
-	"github.com/signalfx/signalfx-agent/pkg/utils"
 	"github.com/signalfx/signalfx-agent/pkg/utils/filter"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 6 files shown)


pkg/core/common/docker/containerlist.go
```diff
@@ -117,33 +172,9 @@ func ListAndWatchContainers(ctx context.Context, client *docker.Client, changeHa
 
 	<-watchStarted
 
-	f := filters.NewArgs()
-	f.Add("status", "running")
-	options := dtypes.ContainerListOptions{
-		Filters: f,
-	}
-	containerList, err := client.ContainerList(ctx, options)
+	err := syncContainerList()
```
keitwb, line 175: Move this to immediately before the for loop on line 122 and then always return nil err from this method.  We don't want the monitor to crash and not work if docker happens to be down when the agent starts up.

keitwb: Then you can get rid of the watchStarted stuff as well.

dloucasfx: done. 
also did more cleanup for unused

keitwb (approved):
I can't see any edge cases that aren't covered.  This should be quite a bit better.  Thanks for adding this.

PR #1612, opened 2021-02-20
opcache monitor based on curl-json collectd

xp-1000:
add monitor for opcache


pkg/monitors/nodetool/monitor.go
```diff
@@ -0,0 +1,28 @@
+package nodetool
```
keitwb, line 1: I think this is extraneous?

xp-1000: my bad, fixed

keitwb (approved):
Looks good, I think you have got the hang of this.  This is a highly specific thing to monitor but it seems legitimate.

keitwb:
Don't worry about the check_links CI failure, that always happens when adding a new monitor.

PR #1617, opened 2021-03-02
K8s helm add pdb

chien-splunk:
We have an internal helm standard of requiring poddisruptionbudgets for all non-daemonset installs.


keitwb (approved):
Ok, seems like a good option to have for the chart.  Thanks for adding it.

keitwb:
@chien-splunk can you run the `deployments/k8s/generate-from-helm` script and commit the changes?

PR #1620, opened 2021-03-04
Make cloud provider metadata log more robust

dloucasfx:
I set the first error to info and the rest to Error. 
If we made it through the curl, this means that we are on AWS, so the anything error later on is an accurate error

cc: @jrcamp @keitwb 

### DIFF
--- a/pkg/core/hostid/aws.go
+++ b/pkg/core/hostid/aws.go
@@ -20,8 +20,8 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 	resp, err := c.Get("http://169.254.169.254/2014-11-05/dynamic/instance-identity/document")
 	if err != nil {
 		log.WithFields(log.Fields{
-			"error": err,
-		}).Debug("Failed to get AWS instance-identity")
+			"detail": err,
+		}).Info("No AWS metadata server detected, assuming not on EC2")
 		return ""
 	}
 	defer resp.Body.Close()
@@ -30,7 +30,7 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 	if err != nil {
 		log.WithFields(log.Fields{
 			"error": err,
-		}).Debug("Failed to read AWS instance-identity response")
+		}).Info("Failed to read AWS instance-identity response")
 		return ""
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/core/hostid/gcp.go
```diff
@@ -41,7 +41,7 @@ func getMetadata(path string, cloudMetadataTimeout timeutil.Duration) string {
 
 	resp, err := c.Do(req)
 	if err != nil {
-		logrus.WithError(err).Debugf("Failed to query GCP Metadata endpoint at %s", url)
+		logrus.WithError(err).Errorf("Failed to query GCP Metadata endpoint at %s - It's possible the agent is not running on GCP", url)
```
jrcamp, line 44: Should this be Info since it may trigger an error if running on AWS?

jrcamp (commented):
Should we change Azure as well for consistency?

pkg/core/hostid/gcp.go
dloucasfx: oh yeah, that's GCP, how did I set it to Error? 🤦

pkg/core/hostid/aws.go
```diff
@@ -21,7 +21,7 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 	if err != nil {
 		log.WithFields(log.Fields{
 			"error": err,
-		}).Debug("Failed to get AWS instance-identity")
+		}).Info("Failed to get AWS instance-identity - It's possible the agent is not running on AWS")
```
keitwb, line 24:
```suggestion
		}).Info("No AWS metadata server detected, assuming not on EC2.")
```

That is a bit more explicit and has less potential for confusion.  It's overwhelming likely that they aren't on AWS, not just "possible".

dloucasfx: done

jrcamp (approved):

dloucasfx:
updated to remove `error` from the message and make the integration tests happy

keitwb:
That didn't seem to make them happy yet.

dloucasfx:
@keitwb 
this is strange, the assertion looks for `level=error` https://github.com/signalfx/signalfx-agent/blob/3d1c91ba28c7921fcec1dc264b39464bb69f5d37/tests/helpers/assertions.py#L186

And the only error log I have changed is after the initial connection succeeding https://github.com/signalfx/signalfx-agent/pull/1620/files#diff-5c32148c814c31d5bd8804f4de667e16e11040ec1e9190a18373e301a1ab772aR33
so looks like for whatever reasons, the aws metedata server connection is not throwing an error.
Is the CI on AWS?

Issue #1625, opened 2021-03-10
On Azure, agent thinks he is on AWS


Poil:
Hi

Running the agent 5.9.0-1 on Ubuntu 20.04 LTS on Azure

```
Failed to unmarshal AWS instance-identity response" body="<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\" \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">\r\n<html xmlns=\"http://www.w3.org/1999/xhtml\">\r\n<head>\r\n<meta http-equiv=\"Content-Type\" content=\"text/html; charset=iso-8859-1\"/>\r\n<title>404 - File or directory not found.</title>\r\n<style type=\"text/css\">\r\n<!--\r\nbody{margin:0;font-size:.7em;font-family:Verdana, Arial, Helvetica, sans-serif;background:#EEEEEE;}\r\nfieldset{padding:0 15px 10px 15px;} \r\nh1{font-size:2.4em;margin:0;color:#FFF;}\r\nh2{font-size:1.7em;margin:0;color:#CC0000;} \r\nh3{font-size:1.2em;margin:10px 0 0 0;color:#000000;} \r\n#header{width:96%;margin:0 0 0 0;padding:6px 2% 6px 2%;font-family:\"trebuchet MS\", Verdana, sans-serif;color:#FFF;\r\nbackground-color:#555555;}\r\n#content{margin:0 0 0 2%;position:relative;}\r\n.content-container{background:#FFF;width:96%;margin-top:8px;padding:10px;position:relative;}\r\n-->\r\n</style>\r\n</head>\r\n<body>\r\n<div id=\"header\"><h1>Server Error</h1></div>\r\n<div id=\"content\">\r\n <div class=\"content-container\"><fieldset>\r\n  <h2>404 - File or directory not found.</h2>\r\n  <h3>The resource you are looking for might have been removed, had its name changed, or is temporarily unavailable.</h3>\r\n </fieldset></div>\r\n</div>\r\n</body>\r\n</html>\r\n" error="invalid character '<' looking for beginning of value"
Mar 10 13:22:36 dpl-prd-vm-1003 systemd-resolved[563]: Server returned error NXDOMAIN, mitigating potential DNS violation DVE-2018-0001, retrying transaction with reduced feature level UDP
```

[...]

PR #1626, opened 2021-03-10
short circuiting the code and avoid log noise

dloucasfx:
address https://github.com/signalfx/signalfx-agent/issues/1625

### DIFF
--- a/pkg/core/hostid/aws.go
+++ b/pkg/core/hostid/aws.go
@@ -25,6 +25,10 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 		return ""
 	}
 	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		log.Debug("HTTP status 404, assuming not on EC2")
+		return ""
+	}
 
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {


keitwb (approved):

keitwb (changes requested):
Make the log message at debug level.

pkg/core/hostid/aws.go
```diff
@@ -25,6 +25,10 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 		return ""
 	}
 	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		log.Info("HTTP status 404, assuming not on EC2")
```
keitwb, line 29: Debug level. This will cause confusion at info level.

dloucasfx: I don't see it much different from the previous message.
If we move to debug then we potentially undid the previous fix of exposing errors without debug

keitwb: I'm not really concerned with a real AWS metadata service returning a 404, that would be a very strange scenario.

dloucasfx: Fair enough! it's just those corner cases that always come back to bite us.
I will push the change

keitwb (approved):

keitwb:
This is superseded by #1633

Issue #1628, opened 2021-03-11
[RFE] Add possibility to skip adding the yum repo in ansible installation task


savec80:
Hello team,
Can you please look at the possibility to add simple when condition to task:
https://github.com/signalfx/signalfx-agent/blob/master/deployments/ansible/roles/signalfx-agent/tasks/yum_repo.yml#L7
to skip this step if some variable (skip_repo_install for example) is passed?
We would like to utilize your role, though we can't rely on external repo and would like to keep rpm locally.
Thanks.

jchengsfx:
@savec80 The changes have been merged and released to ansible galaxy.  Please try it out and open a new issue if necessary.

PR #1630, opened 2021-03-12
Add sfx_skip_repo option to ansible

jchengsfx:
For #1628

### DIFF
--- a/deployments/ansible/README.md
+++ b/deployments/ansible/README.md
@@ -80,6 +80,10 @@ This role sources the following variables:
  - `sfx_win_dl_base_url`: The base URL for Windows agent downloading (**default:** 
    'https://dl.signalfx.com')
 
+ - `sfx_skip_repo` (Linux only): If installing the agent from a custom or self-hosted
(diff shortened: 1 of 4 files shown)


keitwb (changes requested):
Is it possible to just detect if the default `sfx_repo_base_url` has been overridden and skip doing the repo in that case?  If it isn't possible this is fine but I'd rather go the other route if possible.

savec80:
Hey,
We use RH Satellite and just overwrite sfx_repo_base_url is not easily achievable if at all because you have hardcoded: "signalfx-agent-rpm" in url as well as name of gpgkey.
Thanks.

keitwb:
Ok, so @savec80 you basically just want it to completely skip out on any repo management whatsoever and just try to install the package `signalfx-agent` in whatever you have preconfigured on the system?

savec80:
yes, that's correct. Adding external repo on isolated host results in broken yum.

keitwb (approved):

PR #1633, opened 2021-03-18
Use SDK to derive AWSUniqueId

asuresh4:
Use SDK to derive AWSUniqueId, this will give us support for both imdsv2 vs imdsv1 endpoints as opposed to just the latter.

### DIFF
--- a/pkg/core/hostid/aws.go
+++ b/pkg/core/hostid/aws.go
@@ -1,51 +1,47 @@
 package hostid
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 
-	"github.com/signalfx/signalfx-agent/pkg/utils/timeutil"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
 	log "github.com/sirupsen/logrus"
+
+	"github.com/signalfx/signalfx-agent/pkg/utils/timeutil"
 )
 
 // AWSUniqueID constructs the unique EC2 instance of the underlying host.  If
 // not running on EC2, returns the empty string.
 func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
-	c := http.Client{
+	// Pass in the HTTP client so cloudMetadataTimeout from the config
+	// is respected.
+	c := &http.Client{
 		Timeout: cloudMetadataTimeout.AsDuration(),
 	}
 
-	resp, err := c.Get("http://169.254.169.254/2014-11-05/dynamic/instance-identity/document")
+	sess, err := session.NewSession(aws.NewConfig().WithHTTPClient(c))
 	if err != nil {
 		log.WithFields(log.Fields{
 			"detail": err,
-		}).Info("No AWS metadata server detected, assuming not on EC2")
-		return ""
-	}
-	defer resp.Body.Close()
-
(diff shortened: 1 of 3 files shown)


pkg/core/hostid/aws.go
```diff
@@ -1,51 +1,40 @@
...
-
-	err = json.Unmarshal(body, &doc)
+	client := ec2metadata.New(sess)
+	doc, err := client.GetInstanceIdentityDocument()
 	if err != nil {
 		log.WithFields(log.Fields{
 			"error": err,
-			"body":  string(body),
-		}).Info("Failed to unmarshal AWS instance-identity response")
+		}).Errorf("Failed to get AWS instance-identity")
```
keitwb, line 37: For agents that aren't on EC2, there should only be a debug message saying that they don't appear to be on EC2.  There shouldn't be an error log for people running on non-EC2 platforms, it causes a lot of confusion.

keitwb (changes requested):
See comment.

pkg/core/hostid/aws.go
asuresh4: I've updated to retain original log levels ptal.

pkg/core/hostid/aws.go
```diff
@@ -1,51 +1,40 @@
...
-
-	err = json.Unmarshal(body, &doc)
+	client := ec2metadata.New(sess)
+	doc, err := client.GetInstanceIdentityDocument()
 	if err != nil {
 		log.WithFields(log.Fields{
 			"error": err,
-			"body":  string(body),
-		}).Info("Failed to unmarshal AWS instance-identity response")
+		}).Info("Failed to get AWS instance-identity, assuming not on AWS EC2")
```
keitwb, line 37: Don't have a field called `error`, it also causes confusion. Call the field `detail` like it was done before in the code you removed.

asuresh4: Done

keitwb (approved):
@dloucasfx please take a look as well.

pkg/core/hostid/aws.go
```diff
@@ -1,51 +1,40 @@
...
 			"detail": err,
-		}).Info("No AWS metadata server detected, assuming not on EC2")
+		}).Info("Failed to create new session for AWS metadata collection")
 		return ""
 	}
-	defer resp.Body.Close()
 
-	body, err := ioutil.ReadAll(resp.Body)
+	client := ec2metadata.New(sess)
+	doc, err := client.GetInstanceIdentityDocument()
```
dloucasfx, line 33: The SDK has this method https://docs.aws.amazon.com/sdk-for-go/api/aws/ec2metadata/#EC2Metadata.Available (and with context version) that we can use to avoid polluting the logs if the agent is not on EC2.
Shall we take advantage of it?

dloucasfx (changes requested):
aside from the recommendation to use `AvailableWithContext` or `Available`, looks good for me

pkg/core/hostid/aws.go
dloucasfx: Also, does this throw an error `session.NewSession` if you are not on EC2? if yes, then I don't think there is a point to execute `Available`

asuresh4: That's a good point, will update to use the `Available` method. Right now, `session.NewSession` does not throw an error, if it's not an EC2 instance.

asuresh4: Updated.

dloucasfx (approved):
LGTM 👍

PR #1634, opened 2021-03-19
telegraf metrics are cumulative

xp-1000:
hello,

some wrong types on varnish but it is not easy to know the right one (I did not find an accurate varnishstat documentation to rely on ..) and the telegraf does not provide this info sadly.


keitwb (approved):
Seems like they are all cumulatives.  Note that this won't change the reported metric type for the MTSs as for that we are dependent on what telegraf uses.  But good to have it documented better.

PR #1647, opened 2021-03-24
Update procfs dep

rmfitzpatrick:
There's a dependency conflict when sourcing the Smart Agent from the splunk-otel-collector that requires bumping the prometheus/procfs dep to 0.6.0.  These changes address its pointer usage for meminfo: https://github.com/prometheus/procfs/pull/331


keitwb (approved):
Looks good.  If MemTotal for some reason is 0, go will make the result +Inf which at least won't be a runtime error, so it should be good.

PR #1690, opened 2021-03-29
Adjust pylint to ignore duplicate-code

flands:


jchengsfx:
@flands This is already added in https://github.com/signalfx/signalfx-agent/pull/1685 by @keitwb

PR #1731, opened 2021-04-21
Use static 404 for unsupported signalfx-forwarder paths

rmfitzpatrick:
Instead of returning 200 with dynamic message echoing potentially undesired/unsanitized content, the signalfx-forwarder monitor should return generic 404.

### DIFF
--- a/pkg/monitors/forwarder/forwarder.go
+++ b/pkg/monitors/forwarder/forwarder.go
@@ -83,8 +83,14 @@ func setupHandler(ctx context.Context, router *mux.Router, chainType string, sin
 }
 
 func notFoundHandler(w http.ResponseWriter, r *http.Request) {
-	errMsg := fmt.Sprintf("Datapoint or span request received on invalid path '%s'. "+
-		"You should send to the same path that you would on the Smart Gateway.", r.URL.Path)
-	logger.ThrottledError(errMsg)
-	_, _ = w.Write([]byte(errMsg + "\n"))
+	errMsg := "Datapoint or span request received on invalid path"
+	logger.ThrottledError(fmt.Sprintf("%s: %s", errMsg, r.URL.Path))
+
+	errMsg = fmt.Sprintf(
+		"%s. Supported paths: /v2/datapoint, %s, %s, and %s.\n", errMsg,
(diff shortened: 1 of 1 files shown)


pkg/monitors/forwarder/forwarder.go
```diff
@@ -83,8 +83,9 @@ func setupHandler(ctx context.Context, router *mux.Router, chainType string, sin
 }
 
 func notFoundHandler(w http.ResponseWriter, r *http.Request) {
-	errMsg := fmt.Sprintf("Datapoint or span request received on invalid path '%s'. "+
-		"You should send to the same path that you would on the Smart Gateway.", r.URL.Path)
+	errMsg := fmt.Sprintf("Datapoint or span request received on invalid path. " +
+		"You should send to the same path that you would on the Smart Gateway.")
 	logger.ThrottledError(errMsg)
```
rmfitzpatrick, line 88: I wonder if it's still better to log the requested path.

rmfitzpatrick: updated.

pkg/monitors/forwarder/forwarder.go
```diff
@@ -83,8 +83,9 @@ func setupHandler(ctx context.Context, router *mux.Router, chainType string, sin
 }
 
 func notFoundHandler(w http.ResponseWriter, r *http.Request) {
-	errMsg := fmt.Sprintf("Datapoint or span request received on invalid path '%s'. "+
-		"You should send to the same path that you would on the Smart Gateway.", r.URL.Path)
-	logger.ThrottledError(errMsg)
-	_, _ = w.Write([]byte(errMsg + "\n"))
+	errMsg := "Datapoint or span request received on invalid path"
+	logger.ThrottledError(fmt.Sprintf("%s: %s", errMsg, r.URL.Path))
+	errMsg = fmt.Sprintf("%s. You should send to the same path that you would on the Smart Gateway.\n", errMsg)
```
keitwb, line 88: Actually this is an obsolete message since we don't really use Gateway anymore.  This ideally would just be a list of the supported paths for maximum clarity.

rmfitzpatrick: Updated w/ supported paths.

keitwb (approved):

PR #1750, opened 2021-04-30
docs: fix typo in filtering example for prometheus-exporter

artem-nefedov:


keitwb (approved):
Thanks for fixing this @artem-nefedov

PR #1763, opened 2021-05-13
Snowflake support

markfink-splunk:
Add support for querying Snowflake databases, including standard performance and usage metrics.


keitwb (approved):

keitwb:

The docs still need to be regenerated with `make docs` on a Linux dev env.

PR #1817, opened 2021-07-01
Expose globalDimensions in Cloud Foundry tile

pellared:
## Why

Add possibility to add `globalDimensions` e.g. to identify a Cloud Foundry foundation.

## What

Add `global_dimensions` option to the Cloud Foundry's Ops Manager tile which appends its value to `globalDimensions` Smart Agent config option.

## Testing

The new input in the tile config:

image

Metrics with additional dimensions:

https://app.signalfx.com/#/chart/v2/new?template=default&filters=sf_metric:rep.memory

image

### DIFF
--- a/deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
+++ b/deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
@@ -14,7 +14,7 @@ globalDimensions:
   bosh_id: {"#from": "/var/vcap/instance/id", optional: true}
   deployment: {"#from": "/var/vcap/instance/deployment", optional: true}
 <% end %>
-  {"#from": "/var/vcap/jobs/signalfx-agent/config/global-dimensions.yaml", optional: true}
+  _: {"#from": "/var/vcap/jobs/signalfx-agent/config/global-dimensions.yaml", optional: true, flatten: true}
 
 <% if_p("signalfx.interval_seconds") do |interval_seconds| %>
 intervalSeconds: <%= interval_seconds %>
--- a/deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
+++ b/deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
@@ -14,7 +14,7 @@ globalDimensions:
   bosh_id: {"#from": "/var/vcap/instance/id", optional: true}
   deployment: {"#from": "/var/vcap/instance/deployment", optional: true}
 <% end %>
-  _: {"#from": "/var/vcap/jobs/signalfx-agent/config/global-dimensions.yaml", optional: true, flatten: true}
+  {"#from": "/var/vcap/jobs/signalfx-agent/config/global-dimensions.yaml", optional: true}
 
 <% if_p("signalfx.interval_seconds") do |interval_seconds| %>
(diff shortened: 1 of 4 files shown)


deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
```diff
@@ -9,11 +9,12 @@ ingestUrl: {"#from": "/var/vcap/jobs/signalfx-agent/config/ingest_url"}
 apiUrl: {"#from": "/var/vcap/jobs/signalfx-agent/config/api_url"}
 cluster: <%= p("signalfx.cluster", "") %>
 
-<% if not p("nozzle.enabled") %>
 globalDimensions:
+<% if not p("nozzle.enabled") %>
   bosh_id: {"#from": "/var/vcap/instance/id", optional: true}
   deployment: {"#from": "/var/vcap/instance/deployment", optional: true}
 <% end %>
+  {"#from": "/var/vcap/jobs/signalfx-agent/config/global-dimensions.yaml", optional: true}
```
rmfitzpatrick, line 17: Should there be `flatten: true` ?

pellared: Works without it. Any reason to put it?

rmfitzpatrick: Oh, for maps the key must be `_` for it to be supported.  I'm not sure I understand its use cases so I guess not.

rmfitzpatrick (approved):
deferring to @agoallikmaa

deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
pellared: For some reason it was not working 08b588a

pellared:
@keitwb @agoallikmaa PTAL

agoallikmaa:
Looks good to me.

deployments/cloudfoundry/bosh/jobs/signalfx-agent/templates/agent.yaml.erb
keitwb: Yes if you want to merge a map into another map, you should do

`_: {"#from": "/file.yaml", optional: true, flatten: true}`

keitwb: Otherwise I would just make the globalDimensions value a YAML map from CF and then loop through it in the ruby template code and render each key/value separately.

keitwb: If the if conditional above this is true then this won't work because it won't be valid yaml anymore.  Test that and see.

pellared: @keitwb You are correct. Fixed: https://github.com/signalfx/signalfx-agent/pull/1817/commits/d9c7fff37e9d8e2fd534bd8b48a5fd660087642b. I probably did some configuration error when I was testing  https://github.com/signalfx/signalfx-agent/commit/08b588aaeac6651b08ccca4fa19be49d05ea20c5

keitwb (approved):

PR #1823, opened 2021-07-02
Remove biased language from redis metric descriptors

mstumpfx:


docs/monitors/collectd-redis.md
```diff
@@ -137,15 +137,15 @@ Metrics that are categorized as
  - `gauge.changes_since_last_save` (*gauge*)<br>    Number of changes since the last dump
  - `gauge.client_biggest_input_buf` (*gauge*)<br>    Biggest input buffer among current client connections
  - `gauge.client_longest_output_list` (*gauge*)<br>    Longest output list among current client connections
- - ***`gauge.connected_clients`*** (*gauge*)<br>    Number of client connections (excluding connections from slaves)
- - `gauge.connected_slaves` (*gauge*)<br>    Number of connected slaves
+ - ***`gauge.connected_clients`*** (*gauge*)<br>    Number of client connections (excluding connections from followers)
+ - `gauge.connected_slaves` (*gauge*)<br>    Number of connected followers
```
rmfitzpatrick, line 141: Can we rename and use mapping service for backward compatibility?  Doesn't have to be in these changes though.

rmfitzpatrick (approved):

docs/monitors/collectd-redis.md
mstumpfx: I think that would work @rmfitzpatrick that's a good idea.

PR #1830, opened 2021-07-13
daemonize fixture threads and bump ci resource class

rmfitzpatrick:
Integration tests are hanging and I was unable to quickly find a solution when rolling back sanic, pytest-xdist, and pytest-rerunfailures.  Making sure that test helper threads are in daemon mode appears to resolve.

Also increasing our build and test machine resource class to large to help speed up runs and corrects stale url for postgres tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -360,6 +360,7 @@ jobs:
       - run_pytest:
           with_sudo: true
           split: true
+          pytest_options: --reruns 2
 
   k8s_integration_tests:
     executor: machine_image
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -358,6 +360,7 @@ jobs:
       - run_pytest:
           with_sudo: true
           split: true
+          pytest_options: --reruns 3
```
keitwb, line 363: Would maybe set this to at most 2 so that failing tests don't make it waste too much time to fail.  If a test fails more than twice in a row it should be fixed to be more robust.

keitwb (approved):
Seems good overall.

PR #1837, opened 2021-07-16
Retry datapoint submissions for temporary errors

rmfitzpatrick:
EOF and similar connection errors are common but generally don't prevent successful retried submissions.  These changes determine if httpsink's `AddDatapoints()` calls fail for retryable reasons and try once more before dropping points.


keitwb (approved):

PR #1841, opened 2021-07-21
Document Cloud Foundry metric dimensions

agoallikmaa:
Document all common Cloud Foundry metric dimensions. This includes application metadata dimensions added in TAS 2.8.0, for which we needed to document which version they were added in, to avoid confusion as to why they are present for some installations but not others.

### DIFF
--- a/pkg/monitors/cloudfoundry/metadata.yaml
+++ b/pkg/monitors/cloudfoundry/metadata.yaml
@@ -1,9 +1,37 @@
 monitors:
 - dimensions:
+    origin:
+      description: Origin name of the metric. Equal to the prefix of the metric name.
     source_id:
-      description: The source of the metric
+      description: For application container metrics, this is the GUID of the application (equal to `app_id`), for system metrics, this is the origin name (equal to `origin`).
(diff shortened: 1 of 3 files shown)


pkg/monitors/cloudfoundry/metadata.yaml
```diff
@@ -1,9 +1,37 @@
 monitors:
 - dimensions:
+    origin:
```
pellared, line 3: I am not sure what is the main purpose of `metadata.yaml`, however I see that `genmetadata.go` probably contains code generated out of it.

I guess that we should probably either revert changes in `metadata.yaml` or regenerate `genmetadata.go`.

agoallikmaa: Only dimensions were changed in here, and `genmetadata.go` does not use dimensions, so nothing needed to change in that.

pellared (approved):

jrcamp (approved):

keitwb (approved):
LGTM

PR #1848, opened 2021-07-23
use latest FQDN function

dloucasfx:
`fqdn.Get()` has been deprecated for a while and the new function `fqdn.FqdnHostname` is supposedly more accurate https://github.com/Showmax/go-fqdn/blob/master/fqdn.go#L220-L238 and should address many reported issues with wrong hostname.

The only concern I have is that this change might impact a large number of users, especially the ones with invalid hostname dimension but not aware or don't pay much attention to this dimension. It can also result of a large amount of meta updates and a temporary invalid content.
If the concern is valid, I am happy to introduce a new config option to give the user the option to keep using the legacy function `fqdn.Get()` (by default) or switch to the new one `fqdn.FqdnHostname`

cc: @keitwb @jrcamp 

**Update**
@keitwb shares the same concerns described above and suggested to manually set the hostname when we can.
If this is not an option, we'll make the new function configurable

### DIFF
--- a/docs/config-schema.md
+++ b/docs/config-schema.md
@@ -45,6 +45,7 @@ if not set.
 | `signalFxRealm` | no | string | The SignalFx Realm that the organization you want to send to is a part of.  This defaults to the original realm (`us0`) but if you are setting up the agent for the first time, you quite likely need to change this. (**default:** `"us0"`) |
 | `hostname` | no | string | The hostname that will be reported as the `host` dimension. If blank, this will be auto-determined by the agent based on a reverse lookup of the machine's IP address. |
 | `useFullyQualifiedHost` | no | bool | If true (the default), and the `hostname` option is not set, the hostname will be determined by doing a reverse DNS query on the IP address that is returned by querying for the bare hostname.  This is useful in cases where the hostname reported by the kernel is a short name. (**default**: `true`) |
+| `useNewFQDNFunction` | no | bool | If true, the agent will use the newly improved FQDN function, otherwise, use the legacy function. (**default**: `false`) (**default:** `false`) |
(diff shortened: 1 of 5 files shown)


emaderer:
Do you think that it's safer to introduce that new config option you described as part of this review?

dloucasfx:
> Do you think that it's safer to introduce that new config option you described as part of this review?

@emaderer done

emaderer:
> > Do you think that it's safer to introduce that new config option you described as part of this review?
> 
> @emaderer done

@dloucasfx  I just noticed this comment https://github.com/Showmax/go-fqdn/blob/869ae776d5881666b8e89ff9f536176ef1736658/legacy.go#L14 that says that the old fqdn.Get() function will be removed next version so maybe it's better to just move to the new function without the config option? otherwise we won't be able to use newer versions of that library. (sorry for only referencing this after you've already added the config option..)

dloucasfx:
> > > Do you think that it's safer to introduce that new config option you described as part of this review?
> > 
> > 
> > @emaderer done
> 
> @dloucasfx I just noticed this comment https://github.com/Showmax/go-fqdn/blob/869ae776d5881666b8e89ff9f536176ef1736658/legacy.go#L14 that says that the old fqdn.Get() function will be removed next version so maybe it's better to just move to the new function without the config option? otherwise we won't be able to use newer versions of that library. (sorry for only referencing this after you've already added the config option..)

@emaderer right, but I don't think there is any plan to upgrade the showmax library in the smart agent, the agent is currently in maintenance mode and approaching EOL, beside the go-fqdn repo has not been updated for almost 2 years.
Also, based on the internal slack discussion, using the new function as the default will introduce a possible side effect to existing customer.

keitwb (approved):
Looks good.

atoulme:
Closed as inactive. Feel free to reopen if this PR is still being worked on.

PR #1870, opened 2021-08-16
Add partition stats for folder mounted volumes in Windows to monitor filesystems

bjsignalfx:
Made changes to collect partitions stats for partitions in volumes mounted to drives and folders in Windows.

- Tested in Windows Server 2016
- Compared against gopsutil v3.10.20

### DIFF
--- a/pkg/monitors/filesystems/partitions_windows.go
+++ b/pkg/monitors/filesystems/partitions_windows.go
@@ -1,0 +1,197 @@
+// +build windows
+
+package filesystems
+
+import (
+	"strings"
+	"syscall"
+	"unicode/utf16"
+
+	"github.com/pkg/errors"
+	gopsutil "github.com/shirou/gopsutil/disk"
+	"golang.org/x/sys/windows"
+)
+
+const volumeNameBufferLength = uint32(windows.MAX_PATH + 1)
+const volumePathBufferLength = volumeNameBufferLength
+
+func getPartitions(all bool) ([]gopsutil.PartitionStat, error) {
+	return getPartitionsWin(getDriveType, findFirstVolume, findNextVolume, findVolumeClose, getVolumePaths, getFsNameAndFlags)
+}
+
+// getPartitions returns partition stats.
+// Similar to https://github.com/shirou/gopsutil/blob/7e4dab436b94d671021647dc5dc12c94f490e46e/disk/disk_windows.go#L71
+func getPartitionsWin(
+	getDriveType func(rootPath string) (driveType uint32),
+	findFirstVolume func(volName *uint16) (findVol windows.Handle, err error),
+	findNextVolume func(findVol windows.Handle, volName *uint16) (err error),
+	findVolumeClose func(findVol windows.Handle) (err error),
+	getVolumePaths func(volNameBuf []uint16) ([]string, error),
+	getFsNameAndFlags func(rootPath string, fsNameBuf []uint16, fsFlags *uint32) (err error),
+) ([]gopsutil.PartitionStat, error) {
+
+	stats := make([]gopsutil.PartitionStat, 0)
+	volNameBuf := make([]uint16, volumeNameBufferLength)
+
+	handle, err := findFirstVolume(&volNameBuf[0])
+	if err != nil {
+		return stats, errors.WithMessagef(err, "cannot find first-volume")
+	}
+	defer findVolumeClose(handle)
+
+	var volPaths []string
+	if volPaths, err = getVolumePaths(volNameBuf); err != nil {
+		return stats, errors.WithMessagef(err, "find paths error for first-volume %s", windows.UTF16ToString(volNameBuf))
+	}
+
+	var partitionStats []gopsutil.PartitionStat
+
+	if len(volPaths) > 0 {
+		partitionStats, err = getPartitionStats(getDriveType(volPaths[0]), volPaths, getFsNameAndFlags)
+		if err != nil {
+			return stats, errors.WithMessagef(err, "cannot find partition-stats for first-volume %s", windows.UTF16ToString(volNameBuf))
+		}
+		stats = append(stats, partitionStats...)
+	}
+
+	var lastError error
+	for {
+		volNameBuf = make([]uint16, volumeNameBufferLength)
+		if err = findNextVolume(handle, &volNameBuf[0]); err != nil {
+			if errno, ok := err.(syscall.Errno); ok && errno == windows.ERROR_NO_MORE_FILES {
+				break
+			}
+			lastError = errors.WithMessagef(err, "find next-volume last-error")
+			continue
+		}
+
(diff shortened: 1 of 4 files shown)


pkg/monitors/filesystems/partitions_windows.go
```diff
@@ -0,0 +1,123 @@
+package filesystems
```
rmfitzpatrick, line 1: Is there a reason not to use the windows build tag?

pkg/monitors/filesystems/partitions_windows.go
```diff
@@ -0,0 +1,123 @@
...
+				found = true
+				break
+			}
+		}
+		if found {
+			found = false
+			continue
+		}
+		folderMountPoints = append(folderMountPoints, mountPoint)
+	}
```
rmfitzpatrick, line 33:
```suggestion
	for _, mountPoint := range allMountPoints {
		found := false
		for _, drivePartitionStat := range drivePartitionStats {
			if mountPoint == drivePartitionStat.Mountpoint {
				found = true
				break
			}
		}
		if !found {
			folderMountPoints = append(folderMountPoints, mountPoint)
		}
	}
```

imo this is clearer, but I'm not sure I understand the intended flow overall.  Is `allMountPoints` a superset of `drivePartitionStats(true)` or is it missing some of what's reported there?  Have you reimplemented or fixed a ~bug in `gopustil.Partitions(true)`?

bjsignalfx: Yes, `allMountPoints` is a superset of the set of drivePartitionStat.Mountpoint. No, to the second question.

pkg/monitors/filesystems/partitions_windows.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"strings"
+	"syscall"
+
+	gopsutil "github.com/shirou/gopsutil/disk"
+	"golang.org/x/sys/windows"
+)
+
+func (m *Monitor) getPartitions(all bool) ([]gopsutil.PartitionStat, error) {
```
rmfitzpatrick, line 13: It'd be helpful to explain what this does and attribute https://github.com/shirou/gopsutil/pull/781 with the 3 bsd  clauses.

bjsignalfx: Wait he opened that PR in 2019?! Why hasn't his PR merged? I wanted to submit a PR to break up PartitionsWithContext into getting mount points and getting partition stats using the list of mount points.

bjsignalfx: This implementation is inspired more by kostix's answer in https://stackoverflow.com/questions/62731962/get-list-of-fixed-drives

bjsignalfx: Added comments and attributions.

bjsignalfx:
@rmfitzpatrick writing the integration test seems to be more involved than I thought. So I instead ask @jchengsfx for access to his EC2 windows instance (ec2-35-163-38-92.us-west-2.compute.amazonaws.com) where I created a 100 MiB VHD and mounted it to **C:\mnt\testHD**. I'm running the partitions monitor there and you can see the data here https://app.signalfx.com/#/chart/v2/new?template=default&filters=metric_source:filesystems-bremer&filters=sf_metric:df_complex.free

rmfitzpatrick:
@bjsignalfx, it's pretty standard to hardcode a historic view or return value while writing unit tests to ensure that they result in the expected structures/values for things that are nontrivial to mock or that require cumbersome integration test environments.  Having a "works for me" example at PR time doesn't help detect regressions or provide helpful scaffolding for bug fixes and enhancements in the future.

bjsignalfx:
@rmfitzpatrick I added tests to compare the functionality against gopsutils. Crucially there is no test for getting partition stats of folder mounts which is what this fix is for.

pkg/monitors/filesystems/partitions_windows.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+			stats = append(stats, partitionStats...)
+		}
+	}
+
+	return stats, lastError
+}
+
+func getDriveType(rootPath string) (driveType uint32) {
+	rootPathPtr, _ := windows.UTF16PtrFromString(rootPath)
```
rmfitzpatrick, line 88: Are we swallowing a potential error here?

pkg/monitors/filesystems/partitions_windows.go
```diff
@@ -0,0 +1,195 @@
...
+	volPathsBuf := make([]uint16, volumePathBufferLength)
+	returnLen := uint32(0)
+	if err := windows.GetVolumePathNamesForVolumeName(&volNameBuf[0], &volPathsBuf[0], volumePathBufferLength, &returnLen); err != nil {
+		return nil, err
+	}
+
+	return findStrings(volPathsBuf, int(returnLen)), nil
+}
+
+// findStrings returns strings from subslices in uint16s delimited by 0.
```
rmfitzpatrick, line 115: an example in the comment could be helpful since it's not clear what this is actually doing

pkg/monitors/filesystems/partitions_windows.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	return ss
+}
+
+// getFsNameAndFlags sets inputs fsNameBuf and fsFlags with fetched filesystem name (NTFS, FAT32 etc) and flags.
+func getFsNameAndFlags(rootPath string, fsNameBuf []uint16, fsFlags *uint32) error {
+	volNameBuf := make([]uint16, 256)
+	volSerialNum := uint32(0)
+	maxComponentLen := uint32(0)
+	rootPathPtr, _ := windows.UTF16PtrFromString(rootPath)
```
rmfitzpatrick, line 139: same hidden error concern

pkg/monitors/filesystems/partitions_windows.go
bjsignalfx: An error occurs only when the input string has null bytes according to https://pkg.go.dev/golang.org/x/sys@v0.0.0-20210823070655-63515b42dcdf/windows#UTF16PtrFromString. We are passing "good" input strings in all getDriveType() calls.

pkg/monitors/filesystems/partitions_windows.go
bjsignalfx: rootPath never contains null bytes, so error not possible

pkg/monitors/filesystems/partitions_windows.go
bjsignalfx: This function is similar to strings.Split() with the separator being the null character. I'm not sure how I can show in a meaningful way the elements of a []uint16 representing some string. I renamed this function and added comments.

rmfitzpatrick (approved):

PR #1883, opened 2021-08-25
Relax and correct doc code owners

rmfitzpatrick:
~readme editing PRs require docs approval, blocking unrestricted merges for metadata content changes.  These changes undo that by adding the gdi approval team as co-owners.

Also removes gdi-specification team as code owners since it* has different responsibilities.


mstumpfx (approved):

theletterf:
@rmfitzpatrick The change was introduced as part of the repo standards. We need a mechanism to be in the loop whenever docs (MD files) are updated. Could we perhaps restrict writers to certain folders in this repo? Can you picture an alternative way of keeping us in the loop without having to remind us manually?

PR #1902, opened 2021-09-06
Parametrized missing options on templates'

cohandv:


deployments/k8s/helm/signalfx-agent/values.yaml
```diff
@@ -67,6 +67,9 @@ minReadySeconds:
 # Namespace to deploy agent in (Optional: Will default to release namespace)
 namespace:
 
+# For Troubleshooting if you do not want to bind to host change
```
keitwb, line 70:
```suggestion
# Set the DNS policy for the agent daemonset/deployment pods
```

deployments/k8s/helm/signalfx-agent/templates/deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       annotations:
         {{- toYaml .Values.podAnnotations | trim | nindent 8 }}
     spec:
-      dnsPolicy: ClusterFirstWithHostNet
+      dnsPolicy: {{ .Values.dnsPolicy }}
```
keitwb, line 40: Do this same thing for the DaemonSet too so it is consistent.

emaderer:
Hey @cohandv , what's the status of this PR?

deployments/k8s/helm/signalfx-agent/templates/deployment.yaml
cohandv: sorry the delay, I changed it to use the value

emaderer: Thanks! @keitwb  mind taking another look?

keitwb: I don't  see the change in the daemonset.

emaderer: @cohandv  ^^

PR #1903, opened 2021-09-08
monitors/http: support SNI

tchernomax:

PR #1909, opened 2021-09-14
monitors/http: support SNI - fix tls verification

tchernomax:
Else we get :
```
level=info msg="failed verify certificate" error="x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs"
```


tchernomax:
Note: I didn't test this PR as [I can't build the dev image](https://github.com/signalfx/signalfx-agent/pull/1903#issuecomment-915630781)
sorry

keitwb (approved):

pkg/monitors/http/http.go
```diff
@@ -264,6 +264,7 @@ func (m *Monitor) getTLSStats(site *url.URL, logger *logrus.Entry) (dps []*datap
 
 	dimensions := map[string]string{
 		"server_name": host,
+		"sni_server_name": serverName,
```
keitwb, line 267: The linter is saying some issue on this line.  See if you can figure out what -- it isn't obvious to me, maybe tabs vs spaces issue.

tchernomax: @keitwb 
```
max@mde-claranet % gofmt -s pkg/monitors/http/http.go | sponge pkg/monitors/http/http.go
max@mde-claranet % git diff
diff --git a/pkg/monitors/http/http.go b/pkg/monitors/http/http.go
index dad8b665..fd8e9335 100644
--- a/pkg/monitors/http/http.go
+++ b/pkg/monitors/http/http.go
@@ -263,7 +263,7 @@ func (m *Monitor) getTLSStats(site *url.URL, logger *logrus.Entry) (dps []*datap
        }

        dimensions := map[string]string{
-               "server_name": host,
+               "server_name":     host,
                "sni_server_name": serverName,
        }

max@mde-claranet % git push -f perso http_sni
```
I fixed and force-push.

keitwb (changes requested):
It looks like an integration test is broken due to this: https://app.circleci.com/pipelines/github/signalfx/signalfx-agent/12189/workflows/094add9c-a3b6-46a8-8835-0f872e535468/jobs/274730/tests#failed-test-0

Can you see if you can figure out the issue?

keitwb:
Actually this test is failing in other branches too so perhaps not due to your changes.

TLS is notoriously tricky to test with due to the cert generation problem.  I assume you tested this at least with a local build?

tchernomax:
> TLS is notoriously tricky to test with due to the cert generation problem. I assume you tested this at least with a local build?

No I did not, Is it possible to build the agent without the dev image ?

keitwb:
> > TLS is notoriously tricky to test with due to the cert generation problem. I assume you tested this at least with a local build?
> 
> No I did not, Is it possible to build the agent without the dev image ?

Yes, you can compile it on mac or linux with `make signalfx-agent` assuming you have Go and make installed.  You won't get any of the collectd stuff but as long as you just configure the `http` monitor you won't need that.

tchernomax:
@keitwb I build the agent and test it locally. It works.

I also added 2 other commits:
* one that add the dimention "req_host" to all http stats metrics ; you can have to monitor with url http://127.0.0.1/ and an different host header, so we need to be able to differentiate there metrics.
* on that add a documentation example using sniServerName (I had a hard time figuring out I had to also add the Host header, so I think it should be in the examples)

keitwb:
@xp-1000 I know you all were using the http monitor quite a bit, what do you think of this change?

xp-1000:
> > TLS is notoriously tricky to test with due to the cert generation problem. I assume you tested this at least with a local build?
> 
> No I did not, Is it possible to build the agent without the dev image ?

hi @tchernomax, feel free to join me on our claranet slack I can help you with that if needed

xp-1000:
> @xp-1000 I know you all were using the http monitor quite a bit, what do you think of this change?

thanks for the ping, from the code point of view it looks good to me.

no problem for the sni change but I would like to warn about the dimension adding:
in some cases, this can lead to undersired false alert especially for heartbeat detector.

@tchernomax for example our related detectors module does not use aggregation by default: https://github.com/claranet/terraform-signalfx-detectors/blob/master/modules/smart-agent_http/variables-gen.tf#L9. if you don't set it manually (e.g. `mean(by='host','req_host','method')`) so you should get alert each time you change (or remove) an existing host header in your agent configuration.

That said, this warning is out of scope of the agent and I agree we need to be able to differentiate it ideally.

Notice even without the new dimension, we will still get alert for any error even for "concurrent datapoints on the same MTS" because I forced the right `rollup` function to our detectors (e.g. https://github.com/claranet/terraform-signalfx-detectors/blob/master/modules/smart-agent_http/detectors-gen.tf#L37) but adding a new dimension will allow to keep an understandable alert message and help to troubleshoot and know from which exact monitor comes the alert.

may be the right way would be to generate a hash from every relevant configurations values and push it as dimension to ensure each "monitor instance" has its own mts.

xp-1000 (approved):
LGTM

theletterf (approved):

PR #1922, opened 2021-09-22
writer: new option to log rejected spans

jtmal-signalfx:


pkg/core/writer/signalfx/spans.go
```diff
@@ -21,9 +21,21 @@ func (sw *Writer) sendSpans(ctx context.Context, spans []*trace.Span) error {
 		// This sends synchonously
 		err := sw.client.AddSpans(context.Background(), spans)
 		if err != nil {
-			log.WithFields(log.Fields{
-				"error": err,
-			}).Error("Error shipping spans to SignalFx")
+			var meta log.Fields
+			if sw.conf.LogTraceSpansFailedToShip {
+				jsonEncodedSpans, _ := json.Marshal(spans)
```
owais, line 26: err should be checked or this can panic. Also if we failed to ship them it is possible json.Marshal could fail as well.

pkg/core/writer/signalfx/spans.go
```diff
@@ -21,9 +21,21 @@ func (sw *Writer) sendSpans(ctx context.Context, spans []*trace.Span) error {
...
+				meta = log.Fields{
+					"error":   err,
+					"payload": jsonEncodedSpans,
```
owais, line 29: I assume this case will be enabled for debugging purposes only so it is OK to log entire batches?

pkg/core/writer/signalfx/spans.go
keitwb: I think `jsonEncodedSpans` would be nil if there were an err so payload below would just show nil.  A json marshal error is pretty hard to imagine though since we aren't using generic data structures in the spans, just a well-defined span format that we know serializes fine.

docs/config-schema.md
```diff
@@ -154,6 +154,7 @@ The **nested** `writer` config object has the following fields:
 | `logDatapoints` | no | bool | If the log level is set to `debug` and this is true, all datapoints generated by the agent will be logged. (**default:** `false`) |
 | `logEvents` | no | bool | The analogue of `logDatapoints` for events. (**default:** `false`) |
 | `logTraceSpans` | no | bool | The analogue of `logDatapoints` for trace spans. (**default:** `false`) |
+| `LogTraceSpansFailedToShip` | no | bool | If `true`, traces and spans which weren't successfully received by the backend, will be logged as json. (**default:** `false`) |
```
keitwb, line 157: You have to run `make docs` in a linux env to properly generate this, not manually edited.

pkg/core/writer/signalfx/spans.go
jtmal-signalfx: yes, this option is only to be enabled when investigating `tooManySpansInTrace` as the backend gets rid of these

keitwb (approved):

keitwb:
@jtmal-signalfx see if you can figure out what is messing with the spaces in the docs that is causing the docs_test to fail.  It might be that your editor is doing formatting on it unintentionally.

keitwb (approved):

PR #1937, opened 2021-09-28
Fix telegraf/exec timeout setting

dloucasfx:


dloucasfx:
Thanks @keitwb 
I updated the PR to set the default value to 5 seconds (and make the test happy), which is the default set by telegraf https://github.com/signalfx/telegraf/blob/signalfx-agent/plugins/inputs/exec/exec.go#L59

rmfitzpatrick (approved):

PR #1963, opened 2021-10-21
Add SAP Hana monitor

pmcollins:
This monitor was written by the CK team, then modified and tested against an SAP Hana cloud instance.


mstumpfx (approved):
LGTM

rmfitzpatrick (approved):

PR #1970, opened 2021-10-27
Set log message to debug

dloucasfx:
With this config 
```
monitors:
  - type: http
    host: localhost
    port: 4443
    useHTTPS: true
    skipVerify: true
```
and if the server is using invalid cert, we'll log this info message which can pollute the log file.
```
failed verify certificate       {"kind": "receiver", "name": "smartagent/http", "monitorType": "http", "url": "https://localhost:4443/", "error": "x509: certificate is not valid for any names, but wanted to match localhost"}
```
instead, move it to debug

### DIFF
--- a/pkg/monitors/http/http.go
+++ b/pkg/monitors/http/http.go
@@ -257,7 +257,7 @@ func (m *Monitor) getTLSStats(site *url.URL, logger *logrus.Entry) (dps []*datap
 		}
 		_, err := cert.Verify(opts)
 		if err != nil {
-			logger.WithError(err).Info("failed verify certificate")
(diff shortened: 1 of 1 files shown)


AchimGrolimund:
I think that by setting the tag `skipVerify: true`
this message should no longer come because you skip it, so no more testing should take place.
Am i wrong?

 Greetings Achim

dloucasfx:
@AchimGrolimund 
`skipVerify` is used by the http client when it checks the server status `getHTTPStats`.
The error addressed here is part of the monitor cert validation `getTLSStats` , which is independent of `skipVerify`

PR #1984, opened 2021-11-03
sap hana: update metric metadata

pmcollins:
After doing some testing, found that some metrics were missing from metadata and some were mistakenly marked `default: false`.

### DIFF
--- a/docs/monitors/hana.md
+++ b/docs/monitors/hana.md
@@ -56,7 +56,7 @@ Metrics that are categorized as
 
  - ***`sap.hana.connection.message.received.count`*** (*gauge*)<br>    Total count of messages received
 
- - `sap.hana.connection.message.received.size` (*gauge*)<br>    Total size of messages received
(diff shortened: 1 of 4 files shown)


rmfitzpatrick:
Should all of these be default?  It looks like only 4 aren't:

* sap.hana.statement.active.execution.memory.max
* sap.hana.statement.active.execution.memory.sum
* sap.hana.statement.active.execution.sum
* sap.hana.statement.active.execution.time.max

If there's likely only one hana service being monitored at a time then I guess it may not matter though.

rmfitzpatrick (approved):

mstumpfx (approved):

pmcollins:
Looks like there are just two metrics that aren't in the dashboard json:

sap.hana.statement.active.execution.memory.sum
sap.hana.statement.active.execution.sum

(I wrote a script that updates the meta yaml based on the dashboard json, and it works, but I appear to have messed up getting those results into this PR somehow 🤦 . Anyway I've marked them as default: false and pushed that change.)

PR #1991, opened 2021-11-04
Better logging for aws metadata

dloucasfx:
Part of the logging in this [change](https://github.com/signalfx/signalfx-agent/commit/8b4d34d8efebe975955390b7b5ce108b3ba317b9#diff-5c32148c814c31d5bd8804f4de667e16e11040ec1e9190a18373e301a1ab772a)  was undone when we converted to use SDK

In this PR, I opted not to use `Available()` as this func does not return the actual error , which is needed for debugging. 

### DIFF
--- a/pkg/core/hostid/aws.go
+++ b/pkg/core/hostid/aws.go
@@ -31,17 +31,11 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
 
 	client := ec2metadata.New(sess)
 
-	// Checks whether it's an EC2 instance and Metadata service is available.
-	if !client.Available() {
-		log.Debug("No EC2 metadata server detected, assuming not on AWS EC2")
-		return ""
-	}
-
 	doc, err := client.GetInstanceIdentityDocument()
 	if err != nil {
 		log.WithFields(log.Fields{
-			"error": err,
-		}).Error("Failed to get AWS instance-identity")
+			"detail": err,
+		}).Info("No AWS metadata server detected, assuming not on EC2")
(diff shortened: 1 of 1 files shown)


pkg/core/hostid/aws.go
```diff
@@ -31,17 +31,11 @@ func AWSUniqueID(cloudMetadataTimeout timeutil.Duration) string {
...
-		return ""
-	}
-
 	doc, err := client.GetInstanceIdentityDocument()
 	if err != nil {
 		log.WithFields(log.Fields{
-			"error": err,
-		}).Error("Failed to get AWS instance-identity")
+			"detail": err,
+		}).Info("No AWS metadata server detected, assuming not on EC2")
```
asuresh4, line 38: @dloucasfx I maybe missing something but wouldn't this mean even legitimate errors get logged at Info level?

dloucasfx: correct; the reason we made this change in https://github.com/signalfx/signalfx-agent/pull/1620 because it was a nightmare to debug missing AWSUniqueID unless we set the logs to debug and wait for the issue to be reproduce, on the other hand setting the log to error can raise unwanted alerts, so we settled to info and made the message more descriptive (this applies to all cloud metadata Azure, GCP and AWS)

In addition, in the new SDK code, if I kept `if !client.Available() {` and set that error to info, then we'll be missing the error message

asuresh4: Got it, sounds good 👍

asuresh4 (approved):

PR #1992, opened 2021-11-04
Set encoding for ansible/salt test images

jcheng-splunk:
pip fails to install packages for tests unless the encoding is set to utf-8

### DIFF
--- a/deployments/ansible/Dockerfile
+++ b/deployments/ansible/Dockerfile
@@ -1,6 +1,8 @@
 FROM ubuntu:18.04
 
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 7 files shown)


rmfitzpatrick:
I guess we started getting hit with https://github.com/pypa/pip/issues/10219#issuecomment-887950345?

rmfitzpatrick (approved):

jcheng-splunk:
> I guess we started getting hit with [pypa/pip#10219 (comment)](https://github.com/pypa/pip/issues/10219#issuecomment-887950345)?

I think so since these distro images only have python 3.6.

PR #1993, opened 2021-11-09
helm: upgrade to 1.9.4

chien-splunk:
* add config for hostNetwork in podSpec
* use correct secret for splunk token if specified
* allow custom annotations on serviceaccount

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
+++ b/deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
@@ -55,6 +55,13 @@ Create the name of the secret holding the token
 {{- define "signalfx-agent.secretName" -}}
     {{ default (include "signalfx-agent.fullname" .) .Values.signalFxAccessTokenSecretName }}
 {{- end -}}
+{{/*
+Name of secret holding splunk hec token
+*/}}
+{{- define "signalfx-agent.secretNameSplunk" -}}
+    {{ default (include "signalfx-agent.secretName" .) .Values.splunkTokenSecretName }}
+{{- end -}}
+
 
 {{/*
 Create the configmap name. It will have -v<major version> appended to it from v5 and onward.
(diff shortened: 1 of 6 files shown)


deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
```diff
@@ -55,6 +55,13 @@ Create the name of the secret holding the token
 {{- define "signalfx-agent.secretName" -}}
     {{ default (include "signalfx-agent.fullname" .) .Values.signalFxAccessTokenSecretName }}
 {{- end -}}
+{{/*
+Name of secret holding splunk hec token
+*/}}
+{{- define "signalfx-agent.secretSplunkName" -}}
+    {{ default (include "signalfx-agent.secretName") .Values.splunkTokenSecretName }}
```
keitwb, line 62: Don't you need to supply the `.` to the `include` as is done on line 56 above?

deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
```diff
@@ -55,6 +55,13 @@ Create the name of the secret holding the token
 {{- define "signalfx-agent.secretName" -}}
     {{ default (include "signalfx-agent.fullname" .) .Values.signalFxAccessTokenSecretName }}
 {{- end -}}
+{{/*
+Name of secret holding splunk hec token
+*/}}
+{{- define "signalfx-agent.secretSplunkName" -}}
```
keitwb, line 61: Call this `splunkSecretName` or `secretNameSplunk` instead of sticking `splunk` in the middle.

keitwb (changes requested):
See comments.

deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
chien-splunk: Fixed

deployments/k8s/helm/signalfx-agent/templates/_helpers.tpl
chien-splunk: Modified var name

keitwb (approved):

PR #2018, opened 2021-12-02
Circleci to GitHub gotests docs test modules and generate

nitaliya:
Migrate gotests, docs_test, and modules_and_generate jobs
Ticket: [OTL-1259](https://signalfuse.atlassian.net/browse/OTL-1326).

Adding GitHub Action jobs of gotests, docs_test, and modules_and_generate
Removing CircleCi jobs of gotests, docs_test, and modules_and_generate.


.github/workflows/modules_and_generate.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          [[ -d /go/pkg/mod/cache ]] && exit 0
+          # retry up to 3 times in case of network issues
+          for i in $(seq 1 3); do
+              go mod download && exit 0
+              sleep 10
+          done
+          exit 1
+
+      - name: Docs test
```
jcheng-splunk, line 48:
```suggestion
      - name: go mod tidy
```

scripts/code-gen-helpers.sh
```diff
@@ -21,7 +21,7 @@ EOH
 )
 
 generate_selfdescribe_json() {
-  $AGENT_BIN selfdescribe > $selfdescribe_json
+  ./$AGENT_BIN selfdescribe > $selfdescribe_json
```
jcheng-splunk, line 24: Please revert this and set the `AGENT_BIN` env var in the workflows instead.

.github/workflows/docs_test.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+            ~/go/bin
+          key: ${{ runner.os }}-v1-go-mod-${{ hashFiles('**/go.mod') }}-${{ hashFiles('**/go.sum') }}
+
+      - name: go mod download
+        run: |
+          [[ -d /go/pkg/mod/cache ]] && exit 0
```
jcheng-splunk, line 40:
```suggestion
          [[ -d ~/go/pkg/mod ]] && exit 0
```

scripts/code-gen-helpers.sh
nitaliya: As suggested by you I have set `AGENT_BIN` in the workflow itself. 
From this now my understanding is the same function is used by `scripts/docs/add-integrations-doc-notice` so let's not affect it.

.github/workflows/docs_test.yml
nitaliya: I have made the above changes as per your suggestion in gotests, docs_test, and modules_and_generate jobs.
From this now my understanding is for ubuntu instead of checking the cash directory (`/go/pkg/mod/cache`) we need to check the go module path (`~/go/pkg/mod`). Correct me if I am wrong.

jcheng-splunk: We just need to check for the default path set by `actions/setup-go`.

.github/workflows/docs_test.yml
```diff
@@ -0,0 +1,57 @@
...
+          [[ -d ~/go/pkg/mod ]] && exit 0
+          # retry up to 3 times in case of network issues
+          for i in $(seq 1 3); do
+              go mod download && exit 0
+              sleep 10
+          done
+          exit 1
+
+      - name: Install gomplate
+        run: .github/scripts/install-gomplate.sh
```
jcheng-splunk, line 49: Let's just move the installation commands to this step and delete the `install-gomplate.sh` script.

nitaliya: As per suggestion, I have moved installation commands to the workflow step itself. Because `install-gomplate.sh` is not used by any other workflows.

jcheng-splunk (approved):

PR #2020, opened 2021-12-03
Fix cert validation

dloucasfx:
The current code fails to validate intermediate cert that has an expired root cert that is part of the trust store, example: `DST root ca` https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/
example, run the monitor against `letsencrypt.org` and the cert validation will fail, whereas it passes with openssl and chrome.

The problem with the current logic is that it verifies every cert in the chain "independently", so when it reaches the last CA in the chain , in our example: "ISRG Root CA" , the verification func fails to build a chain > 0 with a cert from the trust store as the CA for ISRG is DST which is expired .

The other problem, the current logic does not count for certs with multiple intermediate chains.

I discussed this issue with @keitwb and we decided that it made sense to simply use the client to do the validation, as it covers all those corner cases.

I changed the code to count on `Handshake` to set the TLS stats 

@xp-1000 please take a look

### DIFF
--- a/pkg/monitors/http/http.go
+++ b/pkg/monitors/http/http.go
@@ -3,7 +3,6 @@ package http
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -131,7 +130,7 @@ func (m *Monitor) Configure(conf *Config) (err error) {
 					}
 				}
 			} else {
-				logger.WithError(err).Error("Failed gathering HTTP stats, ignore other stats")
+				logger.WithError(err).Error("Failed gathering all HTTP stats, ignore TLS stats and push what we've successfully collected")
 			}
 
 			for i := range dps {
@@ -214,15 +213,20 @@ func (m *Monitor) getTLSStats(site *url.URL, logger *logrus.Entry) (dps []*datap
 		serverName = host
 	}
 
+	dimensions := map[string]string{
+		"server_name":     host,
+		"sni_server_name": serverName,
+	}
+
 	ipConn, err := net.Dial("tcp", host+":"+port)
 	if err != nil {
+		logger.WithError(err).Error("connection failed to host during TLS stat collection")
 		return
 	}
 	defer ipConn.Close()
 
 	tlsCfg := &tls.Config{
-		InsecureSkipVerify: m.conf.SkipVerify,
-		ServerName:         serverName,
+		ServerName: serverName,
 	}
 
 	if _, err := auth.TLSConfig(tlsCfg, m.conf.CACertPath, m.conf.ClientCertPath, m.conf.ClientKeyPath); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/monitors/http/http.go
```diff
@@ -214,15 +213,20 @@ func (m *Monitor) getTLSStats(site *url.URL, logger *logrus.Entry) (dps []*datap
...
+
 	ipConn, err := net.Dial("tcp", host+":"+port)
 	if err != nil {
+		logger.WithError(err).Error("connection failed to host during TLS stat collection")
 		return
 	}
 	defer ipConn.Close()
 
 	tlsCfg := &tls.Config{
-		InsecureSkipVerify: m.conf.SkipVerify,
```
keitwb, line 224: I think the point of this originally is so that you can validate the time left on a cert even if it is issued by a private CA that the agent doesn't have in its cert store.

So I would just leave this.

dloucasfx: If I leave it as is, then I won't be able to use the client to do the cert verification as  `m.conf.SkipVerify` is getting set to true.
If the cert is issued by a private CA, isn't the expectation that user will set the CA certs `auth.TLSConfig(tlsCfg, m.conf.CACertPath, m.conf.ClientCertPath, m.conf.ClientKeyPath)`  which will make the handshake to pass and the time left on cert successfully  calculated?

keitwb: ok, yeah I guess if it is marked as invalid in the monitor it is less important that the time until expiry get set properly.

keitwb (approved):

dloucasfx:
@xp-1000 do you have any concerns about the change?

xp-1000 (changes requested):
Hello @dloucasfx, sorry for the delay here.

I right understood the problem thanks to the detailed description of the PR.
What you propose seems good from the functional point of view and this drastically simplify the code on TLS part.
However, even if my code was clumsy the goal was to always send tls metrics including the case where certificate is invalid.

Here, if understand right, when certificate is invalid you will not send the other tls metrics anymore (like expiry time).
I personally think that metrics form agent should always be consistent even if it means sending a "zero value".
Having an erratic reporting behavior could lead to undesired display on chart or worst on false alert for detectors (especially for heartbeat ones).

That said, I think in this particular case this could be acceptable because:
- this is few probable someone built complex chart with tls metrics
- if someone created an heartbeat detectors for TLS related metrics he still can use `http.cert_valid` instead of `http.cert_expiry`
- in our self centered case (Claranet, the coroporate I work for) we only use `http.status_code` for heatbeat so it should not be problematic: https://github.com/claranet/terraform-signalfx-detectors/blob/master/modules/smart-agent_http/conf/00-heartbeat.yaml#L10
- back to the creation of this monitor, it was to fill the gap for simple webcheck monitoring. now with Rigor and Splunk Observability this is less necessary even if it is still useful for private checks.

So I think this change is acceptable but please could you add a note on readme to explain that tls related metrics can be erratic depending on the result of the certificate validation warn the users.

dloucasfx:
@xp-1000 
Thanks for looking!

> Here, if understand right, when certificate is invalid you will not send the other tls metrics anymore (like expiry time).
I personally think that metrics form agent should always be consistent even if it means sending a "zero value".

The current code change in this PR will always send all TLS metrics, https://github.com/signalfx/signalfx-agent/pull/2020/files#diff-09c5869e1f3558b54e8f6e5bdda3a471477975363b94d5aed0cc0978e2ab6122R251-R253
the non-initialized of `secondsLeft` will be 0 when cert validation fails

I will go ahead and merge if you there is no other concerns.

PR #2038, opened 2021-12-15
Migrate deb and rpm package tests, and bundle test jobs from CircleCI to GitHub

nitaliya:
Migrate `deb` and `rpm` package tests, and `bundle` test jobs
Ticket: [OTL-1328](https://signalfuse.atlassian.net/browse/OTL-1328).

Changes:
- Adding GitHub Action jobs of `deb` and `rpm` package tests, and `bundle` test.
- Removing CircleCi `package_tests`, and `bundle_tests` jobs.

Note:
- Instead of using [on.<push|pull_request>.paths](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths) in workflow, pushed changes uses `git diff`  to find the relevant changes related to the deployment type. Because by using [on.<push|pull_request>.paths](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths) we can not run specific jobs when some files change within a workflow.


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,102 @@
...
+          make bundle
+          mkdir -p /tmp/workspace
+          docker save -o /tmp/workspace/image.tar quay.io/signalfx/signalfx-agent-dev:latest
+          mv signalfx-agent-latest.tar.gz /tmp/workspace/
+
+      - name: Uploading artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: bundle
+          path: /tmp/workspace/*
```
jcheng-splunk, line 41: Let's split the artifacts into separate steps, one named `image` for `image.tar` and the other named `bundle` for `signalfx-agent-latest.tar.gz` so that dependent jobs won't need to unnecessarily download both.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Check out the codebase.
+        uses: actions/checkout@v2
+
+      - name: Setup bundle
+        env:
+          PULL_CACHE: yes
+          AGENT_VERSION: latest
+        run: |
+          make bundle
+          mkdir -p /tmp/workspace
```
jcheng-splunk, line 33: Let's rename `/tmp/workspace` to `./dist`, and update the paths below.  We only used `/tmp/workspace` in circleci because we didn't want to cache the entire project's workspace.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,102 @@
...
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v2
+
+      - name: Setup bundle
```
jcheng-splunk, line 27:
```suggestion
      - name: Build image and bundle
```

.github/workflows/package_tests.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Check out the codebase.
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Setup package tests
+        env:
+          PULL_CACHE: yes
+        run: |
+          make ${{ matrix.SYS_PACKAGE }}-test-package
```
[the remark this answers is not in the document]
nitaliya: I have tried this approach and it looks good. so I have added one more job `package_build`, and added a dependency to the job for `package_tests`.

.github/workflows/build_and_test.yml
nitaliya: As per your suggestion, I have changed the workflow to split the artifacts into separate steps.

.github/workflows/build_and_test.yml
nitaliya: Make sense. Changes made.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,107 @@
...
+    name: bundle_tests
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Verify Changed files
```
jcheng-splunk, line 58: Let's run the bundle tests for all pushes to `main` and pull requests since it only takes a few minutes.

For the other test jobs that have not yet been migrated to this workflow, maybe we can try https://github.com/dorny/paths-filter instead of the `changes-include-dir` script?

nitaliya: As per your suggestion, I have removed `Verify Changed files` step for `bundle_tests` job in GitHub action.

Yes, I can try with [paths-filter](https://github.com/dorny/paths-filter) to check relevant changes in PR or commit.

jcheng-splunk (approved):

PR #2051, opened 2021-12-28
Migration of installer tests, Check link, and Integration test from CircleCI to GitHub

nitaliya:
Migrate `installer_test`, `check_link`, and `integration_test` jobs
Ticket: [OTL-1327](https://signalfuse.atlassian.net/browse/OTL-1327).

Changes:
- Adding GitHub Action jobs of `installer_test`, `check_link`, and `integration_test`
- Removing CircleCi `installer_test`, `check_link`, and `Integration_test` jobs.

Note:
- **Check link job:**
  - The job is currently failing because dead links of Kubernetes doc (v1.18) are used in docs which is not accessible right now. if required, separate PR for updating Kubernetes doc (v1.19) links can be created later.
- **Integration test job:**
  - To reduce the execution time of CI I have used 4 parallel execution and `pytest-split` for the splitting feature.
  - Test is running on Ubuntu 18.04 Virtual environment, the reason is the Collectd python plugin required version `GLIBC_2.29` which is not available on `ubuntu-latest` (20.04) runner. 
  - By using `fail-fast: false` we are preventing canceling of the job if one of the jobs in the matrix expansion fails.
  - `cgroups_test`: Added args `cgroup_parent="/docker"` because I found that in GutHub runner by default nginx container is created with `cgroup_parent=/actions_job`
  - `collectd/vmem_test`: While running the integration test in GitHub runner I found one more datapoint `file_pmdmapped` is getting collected which is OS-specific. So this change includes a comparison between received datapoints and all expected datapoints using a subset, instead of tightly compare.
  - `docker_container_stats_test`: I found that default datapoint `blkio.io_service_bytes_recursive.read`, and `blkio.io_service_bytes_recursive.write` datapoint is not getting collected in CircleCi, but in GitHub runner, these are getting collected so I have made changes accordingly in integration test


.github/workflows/build_and_test.yml
```diff
@@ -91,3 +89,106 @@ jobs:
...
+    needs: [build]
+    strategy:
+      matrix:
+        group: [1, 2, 3, 4]
+      fail-fast: false
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v2
+
+      - name: Downloading image
```
jcheng-splunk, line 106: Can you try testing with the `bundle` artifact from https://github.com/signalfx/signalfx-agent/blob/main/.github/workflows/build_and_test.yml#L42 instead of the image?  We can try to save some time by not having to download and extract the bigger image.

.github/workflows/build_and_test.yml
```diff
@@ -91,3 +89,106 @@ jobs:
...
+          
+          mkdir -p "$BUNDLE_DIR"
+          cid=$(docker create quay.io/signalfx/signalfx-agent-dev:latest true)
+          docker export $cid | tar -C "$BUNDLE_DIR" -xf -
+          $BUNDLE_DIR/bin/patch-interpreter $BUNDLE_DIR
+          docker rm -fv $cid
+          [ -f "$AGENT_BIN" ] || (echo "$AGENT_BIN not found!" && exit 1)
+
+          MARKERS="integration";
+          if [ ${GITHUB_EVENT_NAME} != 'push' ] || [ ${GITHUB_REF##*/} != 'main' ]; then
```
jcheng-splunk, line 158: Were you able to confirm if these tests pass in github?  If not, please temporarily disable this section so that we can confirm that all tests pass before we merge.

.github/workflows/build_and_test.yml
nitaliya: Yes, I have tried with `bundle` artifact, we can use it instead of `image` to save some time.
I have made changes to that.

.github/workflows/build_and_test.yml
nitaliya: @jcheng-splunk yes I have tested this and it is passing test cases.
For testing I am triggering workflow for test branch by considering test branch as main. So I have made changes in workflow file accordingly  (`[ ${GITHUB_REF##*/} != 'main' ]` -> `[ ${GITHUB_REF##*/} != 'test' ]`)

Test Cases:
1. Push on _main_ branch -> all integration test should run : [Passed](https://github.com/nitaliya/signalfx-agent/runs/4786881473?check_suite_focus=true#step:9:22)
2. Pull request on _main_ or any branch -> specific integration test should run : [Passed](https://github.com/nitaliya/signalfx-agent/runs/4786957937?check_suite_focus=true#step:9:22)

jcheng-splunk:
@nitaliya Please rebase and resolve conflicts.

nitaliya:
> @nitaliya Please rebase and resolve conflicts.

@jcheng-splunk I have rebased the branch. could you please review it?

jcheng-splunk (approved):

PR #2086, opened 2022-01-26
add storage metrics for varnish

xp-1000:
Hello,

Adding 2 important metrics which deserve to be default in my opinion but I can set to false if you prefer.
The PR remains useful because it, at least, add them to metadata so it allow us to add them as extreMetrics explicitly (without to have to set `extraMetrics: ["*"]` and use filtering after.


atoulme:
Closed as inactive. Feel free to reopen if this PR is still being worked on.

Issue #2098, opened 2022-02-03
StatsD parser doesn't handle tag values that contain colons


gregbuehler:
Code: https://github.com/signalfx/signalfx-agent/blob/main/pkg/monitors/statsd/parser.go#L39-L46

The StatsD parser does not properly handle tag values that contain colons. An example tag that may contain a colon `src:12.34.56.78:1234`.

It might be reasonable to use the first element of part as the dimension key and concatenate the remainder of the elements as the value.

PR #2099, opened 2022-02-03
Parse StatsD tags that contain colons

gregbuehler:
This fixes https://github.com/signalfx/signalfx-agent/issues/2098.


pkg/monitors/statsd/parser.go
```diff
@@ -37,12 +37,15 @@ func parseDogstatsdTags(s string) (string, map[string]string) {
 		s = s[:tagsIdx]
 		dims = make(map[string]string)
 		for _, t := range strings.Split(tagsRaw, ",") {
-			parts := strings.Split(t, ":")
+			parts := strings.SplitAfterN(t, ":", 2)
```
keitwb, line 40: Why not just `strings.SplitN(t, ":", 2)`?  It seems like that would eliminate the need to do the trim below?

keitwb (approved):

PR #2101, opened 2022-02-08
Added and updated coredns metrics for metric name changes in coredns-1.7.0-release

jvoravong:
Description:
Currently when installing with the helm chart, we are planning to use signalfx receivers for monitoring Kubernetes control plane components. The coredns monitor in the signalfx agent is a little out of date, coredns updated several metric names in coredns-1.7.0-release. 4 of the metrics the signalfx coredns monitor had name changes. 3 of the metrics with new names are used in our k8s services dashboard. 

Added new metrics and did not delete older metrics for backwards compatibility. I did update the descriptions of the deprecated metrics.

The project docs did get regenerated.

Questions:
Do I have to update other projects or docs for newly added metrics?

### DIFF
--- a/docs/monitors/coredns.md
+++ b/docs/monitors/coredns.md
@@ -67,10 +67,11 @@ Metrics that are categorized as
 
 
  - `coredns_build_info` (*gauge*)<br>    A metric with a constant '1' value labeled by version, revision, and goversion from which CoreDNS was built.
+ - ***`coredns_cache_entries`*** (*cumulative*)<br>    Size of DNS cache.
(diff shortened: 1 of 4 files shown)


mstumpfx (approved):
>  Do I have to update other projects or docs for newly added metrics?

It looks like the following deprecated metrics are used in our built-in content:
 - `coredns_dns_response_rcode_count_total`
 - `coredns_dns_request_count_total`
 - `coredns_cache_size`
We should update that content to include the newer metrics.

PR #2126, opened 2022-03-02
Fix AzureUniqueID when VMSS is used

dloucasfx:
The `AzureUniqueID` has been using the wrong strings function `TrimLeft` to find out the instanceID, instead we should stick to `TrimPrefix`

````
fmt.Println("using TrimLeft: ", strings.TrimLeft("PDAPI290_0", "PDAPI290_"))
fmt.Println("using TrimPrefix: ", strings.TrimPrefix("PDAPI290_0", "PDAPI290_"))

Result:

using TrimLeft:  
using TrimPrefix:  0
````

Little bit more description: `TrimLeft` removes the leading characters of a string that matches the cutset.
In other words, it iterates through the string one character at a time, from left to right and checks if the character exists in the cutset (regardless of order), once it fails to find a char in the cutset it returns the index which is used to slice the string.
Problem in the example I showed, when all character of the string exists in the cutset.


mstumpfx (approved):
👁️ 👄 👁️ 
Good find.

PR #2135, opened 2022-03-04
Move required attributes together for Kafka docs

billg-splunk:
We missed a key configuration because it was at the bottom of a table of optional settings.


dloucasfx:
You can't directly edit the docs, those are generated from  metadata and the monitor config.
If you want to move this option to the top, switch it with `genericjmx.Config` https://github.com/signalfx/signalfx-agent/blob/main/pkg/monitors/collectd/kafka/kafka.go#L21-L23 and run `make docs`

dloucasfx (changes requested):
You can't directly edit the docs, those are generated from metadata and the monitor config.
If you want to move this option to the top, switch it with genericjmx.Config https://github.com/signalfx/signalfx-agent/blob/main/pkg/monitors/collectd/kafka/kafka.go#L21-L23 and run make docs

billg-splunk:
Thanks Dani for the explanation. I will close this one and give it another go.

PR #2216, opened 2022-05-12
Bump k8s.io/client-go to v0.24.0

jeffreyc-splunk:


rmfitzpatrick (approved):

rmfitzpatrick:
This lgtm w/ breaking change note. I don't believe this requires a service account update in the collector helm chart given we don't provide the aws container insights receiver and users are expected to use the otel k8s components instead of the agent that don't use a resource lock/coordination tmk. (cc @dmitryax, @jvoravong)

PR #2327, opened 2022-07-15
update mod tidy for 1.16

rmfitzpatrick:
Should help with intermittent dependabot failures where 1.16 and 1.17 deps differ: https://github.com/signalfx/signalfx-agent/runs/7358951044

### DIFF
--- a/.github/workflows/modules_and_generate.yml
+++ b/.github/workflows/modules_and_generate.yml
@@ -48,5 +48,5 @@ jobs:
       - name: go mod tidy
         run: |
           go generate ./...
-          go mod tidy
+          make tidy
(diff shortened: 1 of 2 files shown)


.github/workflows/modules_and_generate.yml
```diff
@@ -48,5 +48,5 @@ jobs:
       - name: go mod tidy
         run: |
           go generate ./...
-          go mod tidy
+          go mod tidy -go=1.16 && go mod tidy -go=1.17
```
jeffreyc-splunk, line 51: Maybe we can just use
```suggestion
          make tidy
```

PR #2328, opened 2022-07-18
Move comment to be in template

FalconerTC:
I am trying install the `signalfx-agent` via Gitops (by running `helm template` and loading raw manifests) and it's failing because the `deployment` file that gets generated is not empty

```
---
# Source: signalfx-agent/templates/deployment.yaml
# This is only used for the serverless deployment model where we can't use
# DaemonSets.
```

An easy solution to fix this and to prevent the empty `deployment` file from being created is to move this comment into the conditional

### DIFF
--- a/deployments/k8s/helm/signalfx-agent/templates/deployment.yaml
+++ b/deployments/k8s/helm/signalfx-agent/templates/deployment.yaml
@@ -1,6 +1,6 @@
+{{ if .Values.isServerless -}}
 # This is only used for the serverless deployment model where we can't use
 # DaemonSets.
-{{ if .Values.isServerless -}}
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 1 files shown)


rmfitzpatrick:
Thanks for the update. Would you mind providing a bit more context about what is failing with the empty manifest?

Something like `helm template signalfx/signalfx-agent -f values.yaml | kubectl apply -f -` ?

rmfitzpatrick (approved):

FalconerTC:
I think it's Anthos Config Manager specifically that doesn't like seeing empty manifests. I don't have the specific error on hand.

rmfitzpatrick:
@FalconerTC appreciate the info. The helm repo has been updated w/ your fix and the latest agent version.

PR #2334, opened 2022-07-19
All monitors own their loggers and set monitorID field

rmfitzpatrick:
These changes move from the global logrus standard logger to per-monitor instance loggers that also set their `"monitorID"` field to be able to better differentiate between monitor instances. This is required for future Splunk collector distribution functionality and to fix a limitation in the Smart Agent receiver where all instances of a particular monitor type are reported to use the same collector component logger despite collecting data from different entities (makes debugging difficult).

I believe that all subprocess monitors already set a `monitorID` field so this continues the pattern w/ native ones.

edit: there are also a number of import ordering changes that were automatically applied from goland.**

edit: Appears to be some deadlock/race condition scenarios in unrelated test that I've tried to address.

### DIFF
--- a/pkg/monitors/load/load.go
+++ b/pkg/monitors/load/load.go
@@ -7,12 +7,12 @@ import (
 
 	"github.com/shirou/gopsutil/load"
 	"github.com/signalfx/golib/v3/datapoint"
+	"github.com/sirupsen/logrus"
+
 	"github.com/signalfx/signalfx-agent/pkg/core/config"
 	"github.com/signalfx/signalfx-agent/pkg/monitors"
 	"github.com/signalfx/signalfx-agent/pkg/monitors/types"
 	"github.com/signalfx/signalfx-agent/pkg/utils"
-	"github.com/sirupsen/logrus"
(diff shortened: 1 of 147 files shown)


rmfitzpatrick:
I think the integration test failure is unrelated and should be resolved by https://github.com/signalfx/signalfx-agent/pull/2335

atoulme (approved):
LGTM. I have reviewed the changes, they look ok to me.

pkg/monitors/load/load.go
```diff
@@ -7,12 +7,13 @@ import (
 
 	"github.com/shirou/gopsutil/load"
 	"github.com/signalfx/golib/v3/datapoint"
+	"github.com/sirupsen/logrus"
```
jvoravong, line 10: Is this an extra import?

jvoravong (approved):
Just  one question/nit. LGTM. This will be useful.

pkg/monitors/load/load.go
rmfitzpatrick: thanks, missed this one.

PR #2385, opened 2022-08-05
BITS-Updated-Postgresql-Queries-For-Content

Dignesh-Patel:
Updated postgresql queries for our content which is using custom queries

### DIFF
--- a/pkg/monitors/postgresql/queries.go
+++ b/pkg/monitors/postgresql/queries.go
@@ -11,6 +11,359 @@ import (
 func defaultServerQueries(totalTimeColumn string) []sql.Query {
 	return []sql.Query{
 		{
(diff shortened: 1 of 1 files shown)


atoulme:
Thank you so much for this contribution. We are no longer extending functionality in this agent and are moving to work on the OpenTelemetry Collector instead. Please read more about this change with this [notice](https://github.com/signalfx/signalfx-agent/blob/main/docs/smartagent-deprecation-notice.md).

We look forward to helping you within the OpenTelemetry Collector Contrib repository - this [receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/postgresqlreceiver) is likely to be a great place for you to contribute.

PR #2413, opened 2022-08-23
Update deb/rpm signing

jeffreyc-splunk:
Refactor the release pipeline to use the new signing template for deb and rpm.

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -325,7 +325,7 @@ build_and_push_package() {
   local staging_username=$7
   local staging_token=$8
 
-  PUSH_TO_REPO=$push \
+  PUSH_TO_REPO="no" \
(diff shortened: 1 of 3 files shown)


scripts/release
```diff
@@ -325,7 +325,7 @@ build_and_push_package() {
   local staging_username=$7
   local staging_token=$8
 
-  PUSH_TO_REPO=$push \
+  PUSH_TO_REPO="no" \
```
jvoravong, line 328: Double checking. Do we want "no" here?

jeffreyc-splunk: Yeah, this is intentional.  We want to just build the deb/rpm packages here since they'll now be signed/pushed separately.

jvoravong (approved):

PR #2449, opened 2022-09-06
Update deployment versions for v5.23.0

crobert-1:
Ran script to update deployment versions as release process specifies.

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -1,6 +1,6 @@
 FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
 
-ARG AGENT_VERSION="5.22.0"
+ARG AGENT_VERSION="5.23.0"
 LABEL name="SignalFx Smart Agent" \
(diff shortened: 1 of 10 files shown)


dloucasfx (approved):

crobert-1:
```build_and_test / helm_check (pull_request)``` is failing because it expects a tag in the repository with the given release version. Since this commit is to prepare this repo for the next tagged release version, the tag for v5.23.0 does not exist yet.

crobert-1:
```check_links / check_links (pull_request)``` is failing on a link, but I can get there in a browser, so I believe it's a CI error.
[Here's the link.](https://www.brendangregg.com/blog/2017-08-08/linux-load-averages.html)

rmfitzpatrick (approved):
Sorry, I was under the impression the "correct" way would have incorporated the step from the release script (https://github.com/signalfx/signalfx-agent/blob/main/scripts/release#L262) but it's not documented. As long as we don't have a merge commit and use an annotated tag it should hopefully be ok.

PR #2529, opened 2022-10-21
upgrade protobuf dependency to >=3.0.0,<4.21.0

atoulme:


collectd-plugins.yaml
```diff
@@ -67,7 +67,7 @@
   version: 4d0e13f103b2e1b75ed4c7085784ef0279d6ab62
   repo: signalfx/signalfx-collectd-plugin
   pip_packages:
-    - protobuf==3.20.1
+    - protobuf==3.20.2
```
rmfitzpatrick, line 70: Any reason not to go to https://pypi.org/project/protobuf/3.20.3/ or share the proposed `protobuf>=3.0.0,<4.21.0` range from signalfx-python?

atoulme: I was conservative, but that works for me. Changing to use that.

jvoravong (approved):

PR #2661, opened 2023-01-03
update jwt-go and crypto libraries to point at the latest available versions

atoulme:


rmfitzpatrick (approved):

crobert-1 (commented):
I'm not familiar with the `go-test-race` job that's failing, but I can look into it in a bit if no one else gets a chance. At first glance the updated dependencies don't seem to be related to the test's functionality, but I'd need to spend more time looking into it to be sure.

atoulme:
Let’s see if it fails again - please do take a look.

crobert-1:
Looks like it [passed on the main branch](https://github.com/signalfx/signalfx-agent/actions/runs/3832665988), so it may have just been a random failure.

PR #2668, opened 2023-01-05
Update broken SQL link

### LINKED ISSUE #2663: check_links action should ignore SQL link
The `check_links` action is currently failing due to what it detects to be a broken link. The link is valid and working though, and has been in the document for a couple of years at this point, so it's a test failure.

Failing action: https://github.com/signalfx/signalfx-agent/actions/runs/3833349091/jobs/6525260979
Failing line: https://github.com/signalfx/signalfx-agent/blame/main/docs/monitors/sql.md#L82
Failling link: https://dev.mysql.com/doc/refman/8.0/en/show-slave-status.html

There may be a way to get the same link to work successfully, otherwise we should just have `check_links` ignore it.

crobert-1:
The link checker CI/CD action was failing due to a SQL link. The link has now been updated to the proper value.

Fixes #2663


atoulme (approved):

crobert-1:
I decided to leave in the longer check link timeout (default was 10s, updated to 30s) because while testing my fix there ended up being other failures due to ETIMEDOUT errors. This won't impact how long the the check takes because practically all checks finish before 10s. This simply allows some to take a little longer if necessary, rather than having random failures.

atoulme (approved):

PR #2679, opened 2023-01-12
Refresh k8s token

hughesjj:
We're seeing some `401 unauthorized` errors and I'm betting this will fix the issue.

Relevant documentation:
* https://docs.aws.amazon.com/eks/latest/userguide/service-accounts.html
* https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken

Confirmed we're using `1.21.0`.  Likely need to backport this to `4.x` as well.

### DIFF
--- a/pkg/core/common/kubelet/client.go
+++ b/pkg/core/common/kubelet/client.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	k8sTransport "k8s.io/client-go/transport"
 
 	"github.com/signalfx/signalfx-agent/pkg/core/common/auth"
 )
@@ -91,8 +92,8 @@ func NewClient(kubeletAPI *APIConfig, logger log.FieldLogger) (*Client, error) {
 		if err != nil {
 			return nil, err
 		}
-
-		token, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+		tokenPath := "/var/run/secrets/kubernetes.io/serviceaccount/token"
+		token, err := ioutil.ReadFile(tokenPath)
 		if err != nil {
 			return nil, errors.Wrap(err, "Could not read service account token at default location, are "+
 				"you sure service account tokens are mounted into your containers by default?")
@@ -106,12 +107,10 @@ func NewClient(kubeletAPI *APIConfig, logger log.FieldLogger) (*Client, error) {
 		tlsConfig.RootCAs = certs
 		t := transport.(*http.Transport)
 		t.TLSClientConfig = tlsConfig
-
-		transport = &auth.TransportWithToken{
-			RoundTripper: t,
-			Token:        string(token),
+		transport, err = k8sTransport.NewBearerAuthWithRefreshRoundTripper(string(token), tokenPath, t)
+		if err != nil {
+			return nil, errors.Wrap(err, "Could not set up refreshable context for kubernetes AuthTypeService")
 		}
-
 		logger.Debug("Using service account authentication for Kubelet")
 	default:
 		transport.(*http.Transport).TLSClientConfig = tlsConfig


pkg/core/common/auth/token.go
```diff
@@ -11,9 +13,31 @@ type TransportWithToken struct {
...
 	req.Header.Set("Authorization", fmt.Sprintf("bearer %s", t.Token))
-	return t.RoundTripper.RoundTrip(req)
+	attempt := 0
+	resp, err := t.RoundTripper.RoundTrip(req)
+	// Check for 401 unauthorized and reload service account if so
+	for attempt < 3 && err != nil && resp.StatusCode == 401 {
+		// reload auth
+		refreshedToken, authError := RefreshServiceAuth()
+		if authError != nil {
+			t.Token = refreshedToken
```
gsapkal, line 37: Do we need to set this token to Header ?

gsapkal:
> Weird, no idea why neither `gofmt` nor `make gofmt` seems to have any effect on `pkg/core/common/auth/token.go` even though the check fails
> 
> ```
> ❯ gofmt pkg/core/common/auth/token.go | diff pkg/core/common/auth/token.go -
> ❯ git stash
> Saved working directory and index state WIP on refresh_token: b94465ef Try to refresh token if 401 up to 3 times
> ❯ gofmt pkg/core/common/auth/token.go | diff pkg/core/common/auth/token.go -
> ❯ git status
> On branch refresh_token
> Your branch is up to date with 'origin/refresh_token'.
> 
> nothing to commit, working tree clean
> ```

Might be issue with whitespaces Did you try ```gofmt -w token.go .``` ?

pkg/core/common/kubelet/client.go
```diff
@@ -92,24 +92,24 @@ func NewClient(kubeletAPI *APIConfig, logger log.FieldLogger) (*Client, error) {
...
-		}
-
 		rootCAFile := "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 		if err := auth.AugmentCertPoolFromCAFile(certs, rootCAFile); err != nil {
 			return nil, errors.WithMessage(err, fmt.Sprintf("Could not load root CA config from %s", rootCAFile))
 		}
 
+		token, err := auth.RefreshServiceAuth()
+		if err != nil {
+			return nil, errors.Wrap(err, "Could not read service account token at default location, are "+
```
rmfitzpatrick, line 102: This error is already returned by the helper.

pkg/core/common/auth/token.go
```diff
@@ -11,9 +13,31 @@ type TransportWithToken struct {
 	Token string
 }
 
+func RefreshServiceAuth() (string, error) {
```
rmfitzpatrick, line 16: There are issues with this implementation given https://pkg.go.dev/net/http#RoundTripper:

```
A RoundTripper must be safe for concurrent use by multiple goroutines.
```

```
	// RoundTrip should not attempt to interpret the response. In
	// particular, RoundTrip must return err == nil if it obtained
	// a response, regardless of the response's HTTP status code.
	// A non-nil err should be reserved for failure to obtain a
	// response. Similarly, RoundTrip should not attempt to
	// handle higher-level protocol details such as redirects,
	// authentication, or cookies.
```

For k8s, it's not guaranteed that these clients will be using service account auth so I think the logic should be moved to the actual client.

This token helper is also used for the prometheus exporter client so would be problematic in this context.

rmfitzpatrick: 1. Off the cuff I'd assume `DoRequestAndSetValue()` is the place to update the transport's token (w/ RWMutex?).
2. The [file config source](https://github.com/signalfx/signalfx-agent/blob/main/pkg/core/config/sources/file/file.go) has logic for watching for changes but its interface is pretty tied to config content so I don't think would be directly usable as is. The diff detection is by comparing checksums and we could set up a similar loop goroutine in client creation.

pkg/core/common/kubelet/client.go
```diff
@@ -75,7 +76,7 @@ func NewClient(kubeletAPI *APIConfig, logger log.FieldLogger) (*Client, error) {
 		tlsConfig.InsecureSkipVerify = *kubeletAPI.SkipVerify
 	}
 
-	var transport http.RoundTripper = &(*http.DefaultTransport.(*http.Transport))
+	var clientTransport http.RoundTripper = &(*http.DefaultTransport.(*http.Transport))
```
rmfitzpatrick, line 79: nit, this is in `client.go` and I'd keep changes to a minimum.

atoulme:
Build fails here: https://github.com/signalfx/signalfx-agent/actions/runs/3904763219/jobs/6670958219 
```
pkg/core/common/kubelet/client.go:9: File is not `goimports`-ed (goimports)
	"k8s.io/client-go/transport"
make: *** [Makefile:77: lint] Error 1
```

pkg/core/common/kubelet/client.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"k8s.io/client-go/transport"
```
rmfitzpatrick, line 9:
```suggestion
	k8sTransport "k8s.io/client-go/transport"
```

rmfitzpatrick (approved):

PR #2680, opened 2023-01-12
Apply gofmt

hughesjj:
I believe this is currently causing build failures in https://github.com/signalfx/signalfx-agent/pull/2679


crobert-1 (changes requested):
Looks like this file is generated so the change needs to be made where it's generated from. I'll approve once the actions are passing.

atoulme:
I think this PR is not needed, the build fails on a goimports issue at this time, not a gofmt issue: 
https://github.com/signalfx/signalfx-agent/pull/2679#issuecomment-1380832444

PR #2688, opened 2023-01-17
AGENT_BIN env var is needed to run integration tests

hughesjj:


docs/development.md
```diff
@@ -8,7 +8,7 @@ development. To use it first do the following:
 $ make dev-image
 $ # The build will take a while...
 $ mkdir local-etc && cp packaging/etc/agent.yaml local-etc/  # Get a basic operating config file in place
-$ make run-dev-image
+$ AGENT_BIN=/bundle/bin/signalfx-agent make run-dev-image
```
jeffreyc-splunk, line 11: `AGENT_BIN` is not required when starting the dev image.  It is only required when running the integration tests in the dev image to override the default (https://github.com/signalfx/signalfx-agent/blob/main/Dockerfile.dev#L30).

jeffreyc-splunk: It may be better to update https://github.com/signalfx/signalfx-agent/blob/main/tests/README.md?plain=1#L34-L35 instead.

jeffreyc-splunk (approved):

docs/development.md
```diff
@@ -213,6 +214,14 @@ These are all written using Python's pytest and are located in the [tests
 directory](https://github.com/signalfx/signalfx-agent/tree/main/tests).  See
 there for more information.
 
+To run integration tests, run
+
+```bash
+make run-dev-image
+# once inside the dev-image, run
+AGENT_BIN=/bundle/bin/signalfx-agent make run-integration-tests
```
rmfitzpatrick, line 222: I think the paths test helper defaults to this so if you aren't running `make signalfx-agent` before running the tests that's what's expected? `unset AGENT_BIN`: https://github.com/signalfx/signalfx-agent/blob/main/tests/paths.py#L14

jeffreyc-splunk: Basically, if you run `make dev-image` it'll build and save the agent binary to `/bundle/bin/signalfx-agent` in the dev image.  Afterwards, If you make any changes to the agent code, running `make signalfx-agent` within the dev image will rebuild only the agent and and save the binary to `/usr/src/signalfx-agent/signalfx-agent`.  So `AGENT_BIN` will just need to be set to the desired path in order to run the tests.

PR #2743, opened 2023-03-09
Update README.md

crobert-1:
Add `cf` command to target the `system` organization. Without targeting an organization, all of the rest of the "cf" commands will fail. Updating README so the following commands are successful.


atoulme (approved):

atoulme:
Closing as this is no longer needed.

PR #2778, opened 2023-04-07
Fixing systemd metadata so sendLoadState, sendSubState and sendActiveState work again

dloucasfx:
cc: @rmfitzpatrick 

Adding an integration is more involved than first anticipated as it's tricky to get the unit tests containers to work with systemd.
Validated this manually.

### DIFF
--- a/selfdescribe.json
+++ b/selfdescribe.json
@@ -21724,112 +21724,115 @@
         "": {
           "description": "",
           "metrics": [
-            "gauge.active_state.activating",
-            "gauge.active_state.active",
-            "gauge.active_state.deactivating",
-            "gauge.active_state.failed",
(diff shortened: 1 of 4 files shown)


selfdescribe.json
```diff
@@ -21723,113 +21723,116 @@
       "groups": {
         "": {
           "description": "",
+          "metrics": [
+            "gauge.substate.running"
```
crobert-1, line 21727: Why is this metric in its own group instead of being in the `SubState` group?

jeffreyc-splunk:
@dloucasfx Please rebase to pick up https://github.com/signalfx/signalfx-agent/pull/2777.

atoulme:
Closing as incomplete.

